PR #37, opened 2021-07-15
[#223] Enabled file upload fields

Bartvaderkin:


src/formio/components/FileField.js
```diff
@@ -0,0 +1,16 @@
...
+
+
+class FileField extends Formio.Components.components.file {
+  get inputInfo() {
+    const info = super.inputInfo;
+    // change the default CSS classes
+    info.attr.class = applyPrefix('file');
+    return info;
+  }
+}
```
sergei-maertens, line 13: `inputInfo` is not used by the file parent component, since it doesn't inherit from `Input/MultiValue`. So this entire subclass doesn't really have a reason to exist and the default widget should do instead.

src/FormStep.js
```diff
@@ -124,6 +124,7 @@ const FormStep = ({ form, submission, onStepSubmitted }) => {
           <form onSubmit={onReactSubmit}>
             <FormIOWrapper
               ref={formRef}
+              url="http://localhost:3000/bla"
```
sergei-maertens, line 127: `url` is meant to point to a hosted form configuration on that particular and this conflicts with the `form` prop.

sergei-maertens (changes requested):
I'll pickup the feedback and styling now.

sergei-maertens (approved):

sergei-maertens:
Merging, any feedback can be tackled later.

PR #38, opened 2021-07-15
Add kvk validation in frontend

shea-maykinmedia:
Fixes https://github.com/open-formulieren/open-forms/issues/463

There are two things I noticed that could be improved but I'm not sure if it's outside the scope of the ticket:
1. It would be nice to put what is in the constructors of the TextField and Number to a base component that all other components use so that any future plugins don't require configuration of our components.  There didn't seem to be a good way to do this and I'm not sure the likelihood of this happening in the future so I didn't do it now but I can if you think it's a good idea.
2. I noticed with multiple errors the text overlaps because of the `transform: translateY(-31px);`.  I don't think there is a high likelihood of having multiple validations for one component but wanted to mention it.

**Screenshot**

Screenshot 2021-07-16 at 14 17 31

Error with multiple validators

Screenshot 2021-07-16 at 14 34 19

### DIFF
--- a/src/formio/validators/plugins.js
+++ b/src/formio/validators/plugins.js
@@ -3,69 +3,43 @@ import {post} from "../../api";
 const {REACT_APP_BASE_API_URL} = process.env;
 
 
-const KvkNumberValidator = {
-  key: "validate.kvk-kvkNumber",
-  message(component) {
-    return component.t(component.errorMessage('Invalid Kvk Number'), {
-      field: component.errorLabel,
-      data: component.data
-    });
-  },
-  check(component, setting, value) {
-    if (!value) {
-      return true;
+const VALIDATE_ENDPOINT = `${REACT_APP_BASE_API_URL}validation/plugins/`;
+
+const errorMessageMap = {
+  'kvk-kvkNumber': 'Invalid Kvk Number',
+  'kvk-rsin': 'Invalid RSIN',
+  'kvk-branchNumber': 'Invalid Branch Number',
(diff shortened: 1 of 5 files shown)


sergei-maertens:
> It would be nice to put what is in the constructors of the TextField and Number to a base component that all other components use so that any future plugins don't require configuration of our components. There didn't seem to be a good way to do this and I'm not sure the likelihood of this happening in the future so I didn't do it now but I can if you think it's a good idea.

Not possible, since we need to inherit from the FormIO base components and Javascript doesn't have multiple inheritance like Python has.

> I noticed with multiple errors the text overlaps because of the transform: translateY(-31px);. I don't think there is a high likelihood of having multiple validations for one component but wanted to mention it.

Good to be aware of it, for now a `wontfix` indeed.

src/formio/validators/plugins.js
```diff
@@ -0,0 +1,71 @@
+import {post} from "../../api";
+
+const {REACT_APP_BASE_API_URL} = process.env;
```
sergei-maertens, line 3: This is not available runtime, only at build time. It should be taken from the `sdk.OpenForm.baseUrl` configuration. It's included in a React context, so we'll have to pass it down as a custom option _somehow_ to the Form/components.

sergei-maertens (changes requested):
There's quite a bit of repetition of code that can be abstracted into helper functions instead.

sergei-maertens (approved):

PR #41, opened 2021-07-21
Autofill street name and city

shea-maykinmedia:
SDK PR for https://github.com/open-formulieren/open-forms/pull/499

Tested using both the Postcode component and a Textfield component for the postcode and the Textfield Component and Number component for the house number and both worked.  I tested with various formats of postcode (1015 CJ, 1015CJ, 1015cj, 1015 cj) and these all worked.  If an address is not found then the fields are just left blank.

Note: This does **not** work if the house number has a house letter appended (eg. 117a) since the BAG api returns a 400 if this is entered as a house number.  When 117 is filled in a request would be made to the backend and the fields would be autofilled but one would not be made after the 'a' was entered so if the user types 117a it's autofilled but not if they copy-and-pasted, for example. We _could_ strip off any letters when making a request but I'm not sure if we want to do this, if we do let me know and I'll add this. :) 

**Screenshot**

Screenshot 2021-07-22 at 14 30 54

### DIFF
--- a/src/formio/components/TextField.js
+++ b/src/formio/components/TextField.js
@@ -1,7 +1,7 @@
 import { Formio } from 'react-formio';
 
 import { applyPrefix } from '../utils';
-import {get} from "../../api";
+import { get } from '../../api';
 import enableValidationPlugins from "../validators/plugins";
 
 
--- a/src/formio/components/TextField.js
+++ b/src/formio/components/TextField.js
@@ -9,7 +9,7 @@ import {get} from "../../api";
 
 const {REACT_APP_BASE_API_URL} = process.env;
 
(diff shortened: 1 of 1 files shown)


src/formio/components/TextField.js
```diff
@@ -1,10 +1,14 @@
 import { Formio } from 'react-formio';
 
 import { applyPrefix } from '../utils';
-
+import {get} from "../../api";
```
sergei-maertens, line 4: Please keep the code style consistent with the existing code: single quotes for strings, space between curly braces. I'll be setting up `prettier` to enforce this in the near future.

```suggestion
import { get } from '../../api';
```

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #42, opened 2021-07-23
[WIP] Session timeouts

shea-maykinmedia:
Backend PR (which needs to be merged first): https://github.com/open-formulieren/open-forms/pull/514

### DIFF
--- a/src/sdk.js
+++ b/src/sdk.js
@@ -42,7 +42,10 @@ class OpenForm {
   }
 
   handleOnAction () {
-    console.log('In handleOnAction');
+    if (this.idleTimer.getRemainingTime() === 0) {
+      // Redirect user back to initial screen
+      window.location.href = this.basePath;
+    }
   }
 
   async init() {
--- a/src/sdk.js
+++ b/src/sdk.js
@@ -7,6 +7,8 @@ import { Formio, Templates } from 'react-formio';
 import OpenFormsModule from './formio/module';
 import OFLibrary from './formio/templates';
 
+import IdleTimer from 'react-idle-timer';
+
 import './styles.scss';
 
 import { get } from './api';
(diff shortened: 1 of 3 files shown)


src/sdk.js
```diff
@@ -33,6 +35,17 @@ class OpenForm {
...
+
+    // Set up idle timer
+    this.idleTimer = null;
+    this.handleOnAction = this.handleOnAction.bind(this);
+  }
+
+  handleOnAction () {
+    if (this.idleTimer.getRemainingTime() === 0) {
+      // Redirect user back to initial screen
+      window.location.href = this.basePath;
```
shea-maykinmedia, line 47: I think this will need to be updated when https://github.com/open-formulieren/open-forms/issues/447 is completed but for now this seemed like the most appropriate way to "log the user out".

sergei-maertens (commented):
I'm not sure I agree with the direction taken here, let's discuss this next week.

IMO the backend should handle all this - once the session times out, the session cookie will no longer be valid and the API endpoints will start returning HTTP 403 errors. I think the SDK should rather capture those 403s and deal with them. But maybe I'm missing some context that @joeribekker  has, so let's discuss later.

joeribekker (changes requested):
*Oops, I missed Sergei's comment*

Discussed with Sergei. This is not really the way we want this because it comes with 2 problems:

1) Timers don't work well with inactive browser windows/tabs
2) This is a client side approach. Sessions expire on the server as well, as they should.

Furthermore, you're not invalidating any sessions, just a redirect?

Please rework to what is described below:

If the server invalidated the session, the API returns a 403. This should trigger the SDK to show a message that the session most likely expired (most likely because something else could have happened).

**Later**, we'll make this more nice by showing a timeout warning BEFORE the actual timeout.

shea-maykinmedia:
Since we're taking a different approach to this I'm not going to close this PR and delete the branch since I don't think we'll need any of this code.

PR #44, opened 2021-07-28
Session Timeout

shea-maykinmedia:
Backend PR: https://github.com/open-formulieren/open-forms/pull/514

**Screenshot**

Screenshot 2021-07-29 at 11 17 58

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -2,9 +2,15 @@ const fetchDefaults = {
   credentials: 'include', // required for Firefox 60, which is used in werkplekken
 };
 
-const apiCall = (url, opts) => {
+const apiCall = async (url, opts) => {
   const options = { ...fetchDefaults, ...opts };
-  return window.fetch(url, options);
+  const response = await window.fetch(url, options);
+  if (response.status === 403) {
+    const data = await response.json();
+    alert(data.detail);
+    window.location.href = '/';
+  }
+  return response;
 };
 
 const get = async (url, params = {}, multiParams = []) => {
--- a/src/api.js
+++ b/src/api.js
@@ -8,7 +8,6 @@ const apiCall = async (url, opts) => {
   if (response.status === 403) {
     const data = await response.json();
     alert(data.detail);
-    window.location.href = '/';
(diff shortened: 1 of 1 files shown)


shea-maykinmedia:
I _think_ this is what we want for a solution.  I'm not sure if we want to handle more generic errors and not just 403s but I didn't want to over-catch errors so I just made it catch 403s for now.

We probably will also have to update this when https://github.com/open-formulieren/open-forms/issues/447 is implemented since we'll probably need to clear the state/storage but I guess we can just leave a comment in that ticket to mention doing this.

src/api.js
```diff
@@ -2,9 +2,15 @@ const fetchDefaults = {
...
 
-const apiCall = (url, opts) => {
+const apiCall = async (url, opts) => {
   const options = { ...fetchDefaults, ...opts };
-  return window.fetch(url, options);
+  const response = await window.fetch(url, options);
+  if (response.status === 403) {
+    const data = await response.json();
+    alert(data.detail);
+    window.location.href = '/';
```
sergei-maertens, line 11: I don't think this is useful nor correct - it could be that the form start point is hosted on `/subpath/` and then this is wrong.

The basePath is passed as part of the `sdk.js` `OpenForms` constructor.

The better approach would be to still raise/throw the exception and handle it in the ErrorBoundary.

Example of ErrorBoundary usage: https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/Form.js#L162
Definition of ErrorBoundary: https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/ErrorBoundary.js
Documentation on error boundaries: https://reactjs.org/docs/error-boundaries.html

sergei-maertens (changes requested):

src/api.js
shea-maykinmedia: Ah cool,  I didn't know React had this ErrorBoundary thing.  Definitely useful for this situation!

sergei-maertens (approved):

Issue #23, opened 2021-06-11
Set up storybook for component documentation


sergei-maertens:
See https://storybook.js.org/docs/react/get-started/install

- [ ] Install storybook
- [ ] Configure storybook
- [ ] Document usage

sergei-maertens:
Gave this an initial try, but it doesn't seem to be able to handle the null-coalesce operator in an external library, which we'll need to figure out a workaround for.

PR #46, opened 2021-07-29
Project structure update

sergei-maertens:
* Absolute imports are now supported instead of relative imports. Absolute imports are tested against the `src/` directory.
* (React) components that have tests are now in a directory - the compent itself lives in `index.js` and the tests in `tests.spec.js`
* All the React components are now grouped in `src/components` to de-clutter.


SilviaAmAm (commented):
Looks good! I have a question though: why did you choose to have the structure `src > components > ` and then a folder per component (with tests) while the tests for `src > formio > components` are in `jstests`? Why not have, for example, a folder `jstest` inside `src > components ` and `src > formio > components` with all the tests?

shea-maykinmedia (approved):

sergei-maertens:
@SilviaAmAm I am following a recommendation from create-react app to keep the tests as close to the code they are testing as possible. So, in effect, this means for React components that a component lives in its own folder, containing the implementation + tests.

It seemed a bit silly to just create folders already for components without tests, so I didn't do that.

The Formio components are very different from React components, so that's why I left the set up as it is, but probably at some point we should also apply the same philosophy.

We might even expand on this pattern later by also included the sass styles in the component folder & if we're setting up storybook (see #23 ) this is also a pattern that's common - the storybook integration would then also be a file in the component folder.

PR #49, opened 2021-07-30
Feature - Logout button

SilviaAmAm:
Related to https://github.com/open-formulieren/open-forms/issues/532

Currently this works as follows: if a form has a `loginRequired: true`, a logout button appears. It came to mind though that it is possible to have `loginRequired: false` but still specify an auth backend. In that case, it is still possible to login with DigiD/eHerkenning, but there will be no logout button.

Note: I didn't manage to add a test for `FormStep` to make sure the logout button is displayed :cry: The current problem is how to find a way around mocking `useImmerReducer`. 

Note 2: The difference in size build caused by replacing `import useAsync from 'react-use/esm/useAsync';` with `import {useAsync} from 'react-use';` is from 472.06 KB to 472.15 KB

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -9,7 +9,7 @@ import {
 
 import { ConfigContext } from 'Context';
 
-import { get, post } from 'api';
+import {destroy, get, post} from 'api';
 import usePageViews from 'hooks/usePageViews';
 import ErrorBoundary from 'components/ErrorBoundary';
 import FormStart from 'components/FormStart';
@@ -20,6 +20,7 @@ import RequireSubmission from 'components/RequireSubmission';
 import SubmissionConfirmation from 'components/SubmissionConfirmation';
 import Summary from 'components/Summary';
 import Types from 'types';
+import {useIntl} from 'react-intl';
 
 /**
  * Create a submission instance from a given form instance
@@ -81,6 +82,7 @@ const reducer = (draft, action) => {
  const Form = ({ form }) => {
   const history = useHistory();
   usePageViews();
+  const intl = useIntl();
 
   // extract the declared properties and configuration
   const {steps} = form;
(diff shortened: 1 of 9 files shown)


src/components/Form.js
```diff
@@ -147,7 +149,21 @@ const reducer = (draft, action) => {
...
     });
     history.push('/bevestiging');
-  }
+  };
+
+  const onLogout = async (event) => {
+    event.preventDefault();
+
+    const confirmationQuestion = intl.formatMessage(
+      {id: 'confirmLogout', defaultMessage: 'Are you sure that you want to logout?'}
```
sergei-maertens, line 158: I remember reading somewhere that it's not best practice to specify the `id` yourself, but let the extraction tooling deal with that. The example https://formatjs.io/docs/getting-started/message-declaration#using-imperative-api-intlformatmessage also shows that a description is used rather than id.

src/components/Form.js
```diff
@@ -147,7 +149,21 @@ const reducer = (draft, action) => {
...
+    );
+    if (!window.confirm(confirmationQuestion)) {
+      return;
+    }
+    await destroy(`${config.baseUrl}authentication/session`);
+    history.push('/');
+    window.location.reload();
```
sergei-maertens, line 165: fine for now - but I think later on it'd be cleaner to dispatch an action to reset the state completely without having to do a full page reload.

src/components/LogoutButton.js
```diff
@@ -0,0 +1,25 @@
...
+
+import {Toolbar, ToolbarList} from 'components/Toolbar';
+import Button from 'components/Button';
+
+const LogoutButton = ({onLogout}) => {
+  return (
+    <Toolbar modifiers={['bottom','reverse']}>
+      <ToolbarList>
+        <Button variant="danger" onClick={onLogout}>
+          <FormattedMessage id="logOut" defaultMessage="Log out" />
```
sergei-maertens, line 13: here as well - we should probably use description rather than id

src/components/Summary/test.spec.js
```diff
@@ -0,0 +1,88 @@
...
+        />
+      </IntlProvider>,
+      container
+    );
+  });
+
+  expect(container.textContent).toContain('Uitloggen');
+});
+
+it('Summary displays logout button if loginRequired is false', () => {
```
sergei-maertens, line 60:
```suggestion
it('Summary does not display logout button if loginRequired is false', () => {
```

sergei-maertens (changes requested):

src/components/Form.js
SilviaAmAm: As discussed, removing the `id` causes this issue: https://github.com/formatjs/formatjs/issues/1901
We will remove the ids in the future once we set up the whole translation pipeline. For now, a TODO comment was added next to the messages that use IDs.

sergei-maertens (approved):

PR #50, opened 2021-08-05
Handle all mime types

shea-maykinmedia:
Fixes https://github.com/open-formulieren/open-forms/issues/524

The cause of the bug is that this functionality is not supported in FormIO v4.12.7 though it is supported in version 4.13.x so I copied the code from v4.13.0 into our code base.

Ref:
v4.13.0: https://github.com/formio/formio.js/blob/v4.13.0/src/components/file/File.js#L281-L304
v4.12.7: https://github.com/formio/formio.js/blob/v4.12.7/src/components/file/File.js#L271-L283

We could also bump the version up to v4.13.4 though this would require more testing to make sure the whole project still functions.

**Screenshots**

PDF

Screenshot 2021-08-05 at 14 32 16

Image

Screenshot 2021-08-05 at 14 33 47

### DIFF
--- a/src/formio/components/FileField.js
+++ b/src/formio/components/FileField.js
@@ -1,0 +1,37 @@
+import { Formio } from 'react-formio';
+
+/**
+ * Extend the default file field to modify it to our needs.
+ */
+class FileField extends Formio.Components.components.file {
+
+  get browseOptions() {
+    // This code is copied from https://github.com/formio/formio.js/blob/v4.13.0/src/components/file/File.js#L281-L304
+    // Since we are using FormIO v4.12.7 only image based files are supported
+    // If are are using v4.13.x or higher this can probably be deleted
+
+    const options = {};
+
+    if (this.component.multiple) {
+      options.multiple = true;
+    }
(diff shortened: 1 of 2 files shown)


sergei-maertens:
Oh this is actually unexpected - I was under the impression that the SDK was on 4.13 but apparently we're on 4.12. The backend is on 4.13 though, so we should definitely look into using 4.13.

shea-maykinmedia:
@sergei-maertens Okay, should I close this pull request and make one to update the project to use 4.13?

sergei-maertens:
Discussed during standup - let's upgrade to 4.13 and hope nothing explodes, which should fix this issue. Putting this on-hold in the meantime.

shea-maykinmedia:
@sergei-maertens Related to https://github.com/open-formulieren/open-forms/issues/551 I'm not sure if we want to merge this now and remove it later or just wait until the issue mentioned is resolved but I'll leave this to you for now :)

sergei-maertens:
Yes, let's get this merged!

sergei-maertens (approved):

PR #51, opened 2021-08-11
Apply form logic while the user is filling out the form.

sergei-maertens:
Depends on the backend PRs that provide the API endpoint & form logic implementation.

The idea is that on every change that may affect the state/definition of the form, a check is done towards the backend with the latest form step data (debounced to allow for active typing by the user). The backend can then reply with modified form definition within the context of the existing submission data AND the currently filled out fields, even if these are not persisted yet to the backend.

### DIFF
--- a/src/components/FormStep.js
+++ b/src/components/FormStep.js
@@ -26,18 +26,24 @@ const STEP_LOGIC_DEBOUNCE_MS = 300;
 const initialState = {
   configuration: null,
   data: null,
+  canSubmit: true,
 };
 
 const reducer = (draft, action) => {
   switch (action.type) {
     case 'STEP_LOADED': {
-      const {data, formStep: {configuration}} = action.payload;
+      const {data, formStep: {configuration}, canSubmit} = action.payload;
       draft.configuration = configuration;
       draft.data = data;
+      draft.canSubmit = canSubmit;
       break;
     }
     case 'FORM_CHANGED': {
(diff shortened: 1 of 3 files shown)


src/components/FormStep.js
```diff
@@ -70,13 +96,38 @@ const FormStep = ({ form, submission, onStepSubmitted, onLogout }) => {
...
+
+      // TODO: check custom attributes for submission button control
+      const formInstance = formRef.current.instance.instance;
+
+      // we can't just dispatch this, because Formio keeps references to DOM nodes
+      // which expire when the component re-renders, and that gives React
+      // unstable_flushDiscreteUpdates warnings. However, we can update the form
+      // definition by using the ref to the underlying Formio instance.
+      formInstance.setForm(submissionData.step.formStep.configuration);
+      dispatch({type: 'ENABLE_SUBMIT'});
```
sergei-maertens, line 116: I'm not sure about this - what happens if the action is to disable the step submission?

SilviaAmAm: Mhmm  good point :thinking:

SilviaAmAm:
This last commit *should* fix the problem of being able to continue to the next step (along with this bug fix in OF...https://github.com/open-formulieren/open-forms/pull/656).

However, actions of type 'value' (the ones that should change the value of a component), don't work right now. So for now, stil WIP :sweat_smile:

PR #56, opened 2021-09-06
Feature/changing cancelling appointment

shea-maykinmedia:
SDK Pull request for https://github.com/open-formulieren/open-forms/pull/609

**Screenshots**

Screenshot 2021-09-06 at 10 48 43
Screenshot 2021-09-06 at 10 48 49
Screenshot 2021-09-06 at 11 20 44


src/components/CancelAppointment.js
```diff
@@ -0,0 +1,92 @@
...
+          <p>
+            <FormattedMessage id="uwAfspraakOp" defaultMessage="U staat op het punt om uw afspraak op " />
+            {<FormattedDate
+              value={time}
+              day="numeric"
+              month="long"
+            />}
+            <FormattedMessage id="om" defaultMessage=" om " />
+            {new Date(time).toLocaleTimeString().replace(":", ".").split(":")[0]}
+            <FormattedMessage id="teAnnuleren" defaultMessage="  te annuleren. Vul ter validatie uw e-mailadres in." />
```
shea-maykinmedia, line 38: I'm not sure if there is a better way to build up this sentence.  I tried a few different things but couldn't find a better solution.

I can spend more time on it but didn't want to put too much time into this without a second opinion.

src/components/CancelAppointment.js
```diff
@@ -0,0 +1,92 @@
...
+          </p>
+
+          <VerticalSpacer pixels={60} />
```
shea-maykinmedia, line 41: Not sure how we feel about a pattern like this?  

It seemed like a simple enough solution to add vertical spacing while also making a reusable solution that can be used in other places.

sergei-maertens:
will review this tomorrow

sergei-maertens:
didn't get to it, next week...

sergei-maertens:
image

Server-side validation errors are now styled in the same way as Formio errors.

sergei-maertens (approved):

src/components/CancelAppointment.js
sergei-maertens: as discussed in the office - the approach with `values` and `FormattedDate` components etc. into a bigger `FormattedMessage` is the way to go about this.

PR #57, opened 2021-09-09
Improvements/appointments selecting

shea-maykinmedia:
Fixes some points of https://github.com/open-formulieren/open-forms/issues/597.  Specifically:
- Load products upon form step loading, instead of when clicking the pulldown.
- Load locations in the dropdown upon selecting a product.
- Load dates upon selecting a location.
- Load times when a date is selected.
- Whenever nothing is loaded, make the form elements disabled. Also make them disabled when the load is in progress.
- The initial load of products was slow (3.3s, I presume because of WSDL loading/fetching). Subsequent loads were reasonably fast (250ms) [Note: Sort of, now that the initial loading of the products is done when the step is loaded the end user hopefully won't notice this slowness]
- When clicking a dropdown, it looks like the options are refreshed and it closes/re-opens or re-renders. It's not a pleasant user experience

The other points in the ticket should not be affected by these changes and vice versa so I think it makes sense to make this pull request to keep the pull requests small and specific.

### DIFF
--- a/src/formio/components/Select.js
+++ b/src/formio/components/Select.js
@@ -61,7 +61,7 @@ class Select extends Formio.Components.components.select {
   handleSettingAppointmentDates(data) {
     if (this.component.appointmentsShowDates &&
         this.selectOptions.length === 0 &&
-        [this.component.appointmentsProductForDates] &&
+        data[this.component.appointmentsProductForDates] &&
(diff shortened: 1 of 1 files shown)


sergei-maertens (approved):
Looks alright code-wise

PR #58, opened 2021-09-09
Improve appointments information in summary page

shea-maykinmedia:
Changes needed for points
- Should have human readable values in summary page. Will need a way to save these human readable values after they have been selected.
- On the overview page, the appointment details are rendered weirdly

of https://github.com/open-formulieren/open-forms/issues/597

Related Backend Pull Request: https://github.com/open-formulieren/open-forms/pull/645

By saving both the id and name we can then access the name in the summary page and display it rather than the id.

This also formats the dates and times, both in the summary page and when choosing a date or time.

**Screenshots**

Screenshot 2021-09-09 at 15 21 27

Screenshot 2021-09-09 at 15 21 38

Screenshot 2021-09-09 at 15 22 08

### DIFF
--- a/src/formio/components/Select.js
+++ b/src/formio/components/Select.js
@@ -69,7 +69,7 @@ class Select extends Formio.Components.components.select {
          'location_id': data[this.component.appointmentsLocationForDates]})
         .then(results => {
             this.setItems([]);
-            results.map(result => this.addOption(result.date, getFormattedDateString(result.date)));
+            results.map(result => this.addOption(result.date, getFormattedDateString(this.options.intl, result.date)));
             this.element.lastElementChild.removeAttribute("disabled");
         })
         .catch(error => console.log(error));
@@ -88,7 +88,7 @@ class Select extends Formio.Components.components.select {
          'date': data[this.component.appointmentsDateForTimes]})
         .then(results => {
             this.setItems([]);
-            results.map(result => this.addOption(result.time, getFormattedTimeString(result.time)));
+            results.map(result => this.addOption(result.time, getFormattedTimeString(this.options.intl, result.time)));
             this.element.lastElementChild.removeAttribute("disabled");
         })
         .catch(error => console.log(error));
--- a/src/formio/components/Select.js
+++ b/src/formio/components/Select.js
@@ -2,6 +2,7 @@ import { Formio } from 'react-formio';
 
 import { applyPrefix } from '../utils';
 import { get } from '../../api';
+import {getFormattedDateString, getFormattedTimeString} from '../../utils';
 
 
 /**
@@ -68,7 +69,7 @@ class Select extends Formio.Components.components.select {
          'location_id': data[this.component.appointmentsLocationForDates]})
         .then(results => {
             this.setItems([]);
-            results.map(result => this.addOption(result.date, result.date));
+            results.map(result => this.addOption(result.date, getFormattedDateString(result.date)));
             this.element.lastElementChild.removeAttribute("disabled");
         })
         .catch(error => console.log(error));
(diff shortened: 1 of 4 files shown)


src/utils.js
```diff
@@ -8,6 +8,20 @@ import Image from 'components/Image';
...
+  // Expects YYYY-MM-DD and converts to DD-MM-YYYY
+  return dateString.split('-').reverse().join().replaceAll(',', '-');
+};
+
+
+export const getFormattedTimeString = (dateTimeString) => {
+  // Format an ISO date string to our preferred format
+  // Eg. 2021-09-20T08:00:00+02:00 to 08:00
+  return dateTimeString.split('T')[1].split(':').slice(0, 2).join().replace(',', ':');
+};
```
sergei-maertens, line 22: This is bonkers - in my Brave browser (Chromium based) I can get an actual `Date` instance like this:

```js
new Date('2021-09-09T17:25:00+02:00')
```

This should then be fed to react-intl components or underlying imperative API and be properly formatted given the locale from the browser.

sergei-maertens (changes requested):

src/utils.js
shea-maykinmedia: Ya we can do this.  I am a little worried about the performance impact of this since it's possible there will be a lot of dates or times but for now I suppose we can convert the string to a date and then use react-intl components.

src/utils.js
```diff
@@ -8,6 +8,16 @@ import Image from 'components/Image';
 import Anchor from 'components/Anchor';
 
 
+export const getFormattedDateString = (dateString) => {
+  return new Date(dateString).toLocaleDateString();
```
shea-maykinmedia, line 12: Not sure if we want to change this to `return new Date(dateString).toLocaleDateString('nl-NL');` to force the `locale`.  As it is now the date for me was display as DD/MM/YYYY even when changing my browser settings to Dutch.  If we force the `locale` then the dates will be displayed as DD-MM-YYYY.

src/formio/components/Select.js
```diff
@@ -115,6 +116,16 @@ class Select extends Formio.Components.components.select {
     }
   }
 
+  beforeSubmit() {
+      if (this.component.appointmentsShowProducts || this.component.appointmentsShowLocations) {
+        // For these two types of components we need to send both the identifier and name to the backend
+        const value = this._data[this.component.key].toString();
+        const selectedOption = this.selectOptions.filter(option => option.value === value)[0];
```
shea-maykinmedia, line 123: From what I could see when changing the selected option in the html it did not do anything to change the component itself, just changed what is in `this._data` so this seemed to be the best way to access the data that we would want.

src/utils.js
sergei-maertens: that's not what I said in my review though, I said to use the react-intl package tooling. They have a `FormattedDate` component - which you can't use here because this is not a react component, but you should be able to use the underlying imperative API: https://formatjs.io/docs/react-intl/api#formatdate

sergei-maertens: Needs to use the `react-intl` functions.

src/utils.js
```diff
@@ -8,6 +8,16 @@ import Image from 'components/Image';
...
+};
+
+
+export const getFormattedTimeString = (dateTimeString) => {
+  return new Date(dateTimeString).toLocaleTimeString(undefined,  {hour: '2-digit', minute:'2-digit'});
```
sergei-maertens, line 17: Needs to use the react-intl functions.

sergei-maertens (changes requested):

src/utils.js
```diff
@@ -8,6 +8,20 @@ import Image from 'components/Image';
 import Anchor from 'components/Anchor';
 
 
+// https://formatjs.io/docs/react-intl/api/#createintl
+const intl = createIntl({locale: 'nl-NL',}, createIntlCache());
```
shea-maykinmedia, line 12: Is it okay to hardcode the locale like this?  If no locale is specified an error is logged and says that it'll fallback to english.  I'm not sure if we want to make the locale configurable at this point or if we see it as outside the scope of the MVP.

sergei-maertens: as discussed on Slack - the `useIntl()` hook call should be done in the component and passed as argument to the utility functions.

sergei-maertens (approved):

PR #59, opened 2021-09-13
[#607] Implemented Form/FormStep .publicName

Bartvaderkin:
Complement of https://github.com/open-formulieren/open-forms/pull/652


Bartvaderkin:
Closing this as we went with a different approach.

PR #62, opened 2021-09-16
Check the background processing status/result "while" submitting the form

sergei-maertens:
**Changes**

* Polls status endpoint to check background processing
* Renders confirmation page content + PDF download URL on successful processing (and reference)
* Renders error message if available and processing failed
* Render payment start form to Ogone

### DIFF
--- a/src/components/ErrorBoundary.js
+++ b/src/components/ErrorBoundary.js
@@ -1,8 +1,6 @@
 import React from 'react';
 
-import Body from 'components/Body';
-import FAIcon from 'components/FAIcon';
-import {getBEMClassName} from 'utils';
+import ErrorMessage from 'components/ErrorMessage';
 
 
 const logError = (error, errorInfo) => {
@@ -28,16 +26,8 @@ class ErrorBoundary extends React.Component {
     if (!this.state.hasError) {
       return this.props.children;
     }
-
     return (
-      <div className={getBEMClassName('alert', ['error'])}>
-        <span className={getBEMClassName('alert__icon', ['wide'])}>
-          <FAIcon icon="exclamation-circle" />
-        </span>
-        <Body>
-          Er ging helaas iets fout!
-        </Body>
-      </div>
+      <ErrorMessage>Er ging helaas iets fout!</ErrorMessage>
     );
   }
 }
--- a/src/components/PaymentForm.js
+++ b/src/components/PaymentForm.js
@@ -9,7 +9,11 @@ const AUTOSUBMIT_AFTER = 5000;
 
 /**
  * Renders a plain old HTML form to submit/start the payment.
- * @param  {String} options.method The HTTP method to use (get or post)
+ * @param  {String} method The HTTP method to use (get or post)
+ * @param  {String} url The form submit URL
+ * @param  {String} data The form data as an object of key-value pairs
+ * @param  {Boolean} autoSubmit Whether to submit the form when it's rendered, after
+     a static timeout.
  * @return {JSX}
  */
 const PaymentForm = ({ method, url, data, autoSubmit=true }) => {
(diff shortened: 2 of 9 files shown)


src/components/SubmissionConfirmation.js
```diff
@@ -1,75 +1,160 @@
...
+ * Given a start URL, fetches the payment start information and renders the UI controls
+ * to enter the payment flow.
+ * @param  {String} startUrl The payment start URL received from the backend.
+ * @return {JSX}
+ */
+const StartPayment = ({startUrl}) => {
+  const fullUrl = getStartPaymentUrl(startUrl);
+  const {loading, value} = useAsync(
+    async () => {
+      const resp = await post(fullUrl);
```
SilviaAmAm, line 34: At the moment if this request throws an error, it will break the page and the user might not be able to read their submission reference. Maybe good to have a try/catch that then displays an error?

src/components/Summary/index.js
```diff
@@ -7,13 +7,14 @@ import {useImmerReducer} from 'use-immer';
...
+import Loader from 'components/Loader';
+import LogoutButton from 'components/LogoutButton';
+import PrivacyCheckbox from 'components/PrivacyCheckbox';
 import { Toolbar, ToolbarList } from 'components/Toolbar';
 import Types from 'types';
 import { flattenComponents } from 'utils';
-import LogoutButton from 'components/LogoutButton';
-import PrivacyCheckbox from '../PrivacyCheckbox';
 
 const PRIVACY_POLICY_ENDPOINT = '/api/v1/config/privacy_policy_info/';
```
SilviaAmAm, line 19: I think this is my mistake, but there shouldn't be a trailing slash here?

src/components/ErrorBoundary.js
```diff
@@ -28,16 +26,8 @@ class ErrorBoundary extends React.Component {
...
     return (
-      <div className={getBEMClassName('alert', ['error'])}>
-        <span className={getBEMClassName('alert__icon', ['wide'])}>
-          <FAIcon icon="exclamation-circle" />
-        </span>
-        <Body>
-          Er ging helaas iets fout!
-        </Body>
-      </div>
+      <ErrorMessage>Er ging helaas iets fout!</ErrorMessage>
```
SilviaAmAm, line 30: Should maybe this also be a translatable string?

src/components/PaymentForm.js
```diff
@@ -0,0 +1,56 @@
...
+
+import Button from 'components/Button';
+
+const AUTOSUBMIT_AFTER = 5000;
+
+
+/**
+ * Renders a plain old HTML form to submit/start the payment.
+ * @param  {String} options.method The HTTP method to use (get or post)
+ * @return {JSX}
```
SilviaAmAm, line 13: Missing props in documentation.

src/components/SubmissionConfirmation.js
```diff
@@ -1,75 +1,160 @@
...
+  startUrl: PropTypes.string.isRequired,
+};
+
 
 /**
  * Renders the confirmation page displayed after submitting a form.
- * @param reportDownloadUrl - The URL where the PDF of the submission report can be downloaded
- * @param reportStatusUrl - The URL where to check if the generation of the report is complete
- * @param content - Content to display in the confirmation page
+ * @param statusUrl - The URL where to check if the processing of the submission is complete
```
SilviaAmAm, line 65: missing `onFailure` documentation

SilviaAmAm (changes requested):

src/components/Summary/index.js
sergei-maertens: yeah and I messed up while rebasing, will fix

src/components/ErrorBoundary.js
sergei-maertens: yeah, but so far I've only been adding it for new code because the `id`s everywhere are bothering me. I've got another ticket to set up the actual i18n toolchain with message extraction similar to the open-forms repo.

src/components/SubmissionConfirmation.js
sergei-maertens: Good catch - decided to wrap it in an `ErrorBoundary` and tested by manually `throw`ing in the async callback.

PR #65, opened 2021-09-17
reCAPTCHA Enterprise

SilviaAmAm:
Frontend for https://github.com/open-formulieren/open-forms/issues/244

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -88,6 +88,13 @@ const reducer = (draft, action) => {
       draft.submission = draft.submittedSubmission;
       break;
     }
+    case 'LOAD_RECAPTCHA': {
+      const siteKey = action.payload;
+      const script = document.createElement('script');
+      script.src = `https://www.google.com/recaptcha/enterprise.js?render=${siteKey}`;
+      document.body.appendChild(script);
+      break;
+    }
     default: {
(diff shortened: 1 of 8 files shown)


src/components/Form.js
```diff
@@ -140,6 +147,12 @@ const reducer = (draft, action) => {
       payload: submission,
     });
 
+    // Load the google script that makes ReCAPTCHA work
+    dispatch({
+      type: 'LOAD_RECAPTCHA',
+      payload: config.reCaptchaSiteKey
+    });
```
sergei-maertens, line 154: Please do not use dispatch for that, but write a custom hook which is built using `useEffect` hook. Probably you'll also need `useRef` to keep track of the script element.

src/sdk.js
```diff
@@ -20,6 +20,8 @@ Formio.use(OpenFormsModule);
 // use our own template library
 Templates.current = OFLibrary;
 
+const {REACT_APP_RECAPTCHA_SITE_KEY} = process.env;
```
sergei-maertens, line 23: This means you need a key at _build_ time, which we don't have. This should be an embed option instead of getting from the env.

sergei-maertens (changes requested):
Not sure about the rest of the changes and how reCaptcha works, let's walk through it next week.

PR #71, opened 2021-09-22
[#657] Set up SDK Sentry

SilviaAmAm:
Fixes https://github.com/open-formulieren/open-forms/issues/657

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,7 +1,5 @@
 import classNames from 'classnames';
 import {FormattedNumber} from 'react-intl';
-import * as Sentry from "@sentry/react";
-import { Integrations } from "@sentry/tracing";
 
 import {applyPrefix} from './formio/utils';
 import Body from 'components/Body';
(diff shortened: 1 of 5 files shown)


joeribekker (changes requested):
Search/Replace `sentryDns` with `sentryDsn` :)

src/utils.js
```diff
@@ -142,3 +144,14 @@ const humanFileSize = (size) => {
   const unit = ['byte', 'kilobyte', 'megabyte', 'gigabyte', 'terabyte'][index];
   return {size: newSize, unit};
 };
+
+
+export const initialiseSentry = (sentryDns) => {
+  if (sentryDns) {
+    Sentry.init({
+      dsn: sentryDns,
```
joeribekker, line 152: I forgot to spec in the tasks, but it's nice to also set the environment param so we know which client/environment the issue is about.

joeribekker (changes requested):

sergei-maertens (approved):

PR #88, opened 2021-10-13
Implement saving a form to be completed at a later time

shea-maykinmedia:
SDK PR for https://github.com/open-formulieren/open-forms/pull/804

**Screenshots**

Screenshot 2021-11-17 at 13 39 57

Screenshot 2021-11-17 at 13 39 20


src/components/FAIcon.js
```diff
@@ -3,14 +3,19 @@ import PropTypes from 'prop-types';
...
     'fa',
     'fas',
     'fa-icon',
     `fa-${icon}`,
     ...modifiers.map(mod => `fa-icon--${mod}`),
   );
+
+  if (extraClassname) {
+    className += ` ${extraClassname}`;
+  }
```
sergei-maertens, line 17: no, use the functionality of the classnames lib.

```suggestion
  const className = classNames(
    'fa',
    'fas',
    'fa-icon',
    `fa-${icon}`,
    ...modifiers.map(mod => `fa-icon--${mod}`),
    extraClassname,
  );
```

src/components/FAIcon.js
```diff
@@ -25,6 +30,7 @@ FAIcon.propTypes = {
     PropTypes.string,
     PropTypes.elementType,
   ]),
+  extraClassname: PropTypes.string,
```
sergei-maertens, line 33:
```suggestion
  extraClassName: PropTypes.string,
```
this keep sit consistent with the `className` react prop.

src/components/FormStep.js
```diff
@@ -121,6 +122,9 @@ const FormStep = ({
   const controller = useRef(new AbortController());
   const configurationRef = useRef(configuration);
 
+  const [shouldFormStepSaveModalBeOpen, setShouldFormStepSaveModalBeOpen] = useState(false);
+  const closeFormStepSaveModal = () => setShouldFormStepSaveModalBeOpen(false);
```
sergei-maertens, line 126: we already have an immer reducer for this component, use that instead of adding extra `useState` calls

src/sdk.js
```diff
@@ -18,6 +18,7 @@ import { AddFetchAuth } from 'formio/plugins';
 import {fixIconUrls as fixLeafletIconUrls} from 'map';
 import {loadLocaleData, loadFormioTranslations} from 'i18n';
 import initialiseSentry from 'sentry';
+import ReactModal from "react-modal";
```
sergei-maertens, line 21:
```suggestion
import ReactModal from 'react-modal';
```

src/components/modals/Modal.js
```diff
@@ -0,0 +1,36 @@
...
+import PropTypes from 'prop-types';
+import ReactModal from 'react-modal';
+import classNames from 'classnames';
+
+import FAIcon from 'components/FAIcon';
+
+const CONTENT_CLASS_NAME = 'react-modal__content';
+
+const Modal = ({ isOpen=false, title='', closeModal, children, contentModifiers=[] }) => {
+    const modifiedClassNames = contentModifiers.map(modifier => `${CONTENT_CLASS_NAME}--${modifier}`);
```
sergei-maertens, line 11: don't we have a `getBEMClassName` in utils somewhere?

src/components/modals/Modal.js
```diff
@@ -0,0 +1,36 @@
...
+    const className = classNames(
+        CONTENT_CLASS_NAME,
+        ...modifiedClassNames
+    );
+    return (
+        <ReactModal isOpen={isOpen} onRequestClose={closeModal} className={className} overlayClassName="react-modal__overlay">
+            <header className="react-modal__header">
+                { title ? <h2 className="react-modal__title">{title}</h2> : null }
+                <FAIcon icon="close" extraClassname="react-modal__close" title="Sluiten" onClick={closeModal} />
```
sergei-maertens, line 20: "Sluiten" must be translatable

src/components/modals/FormStepSaveModal.js
```diff
@@ -0,0 +1,115 @@
...
+import {getBEMClassName} from 'utils';
+import Modal from 'components/modals/Modal';
+import ErrorMessage from 'components/ErrorMessage';
+import Body from 'components/Body';
+import Label from 'components/Label';
+import Input from 'components/Input';
+import HelpText from 'components/HelpText';
+import {Toolbar, ToolbarList} from 'components/Toolbar';
+import Button from 'components/Button';
+
```
sergei-maertens, line 18: Please group and order your imports properly. External dependencies first, then our own stuff separated by a blank line and sort alphabetically within a group.

src/components/modals/FormStepSaveModal.js
```diff
@@ -0,0 +1,115 @@
...
+
+const FormStepSaveModal = ({
+    isOpen,
+    closeModal,
+    stepData,
+    saveStepDataUrl,
+    suspendFormUrl,
+    destroySessionUrl,
+}) => {
```
sergei-maertens, line 27: we have a very specific component here tightly coupled to the `FormStep` - please pass the `submission` and `submissionStep` props in their entirety rather than just some attributes. I think we can reduce the number of props for the component that way.

src/components/modals/FormStepSaveModal.js
```diff
@@ -0,0 +1,115 @@
...
+    if (!response.ok) return setFailed(true);
+    response = await post(suspendFormUrl, {email});
+    if (!response.ok) return setFailed(true);
+    try {
+      // Destroy throws an exception if the API is not successful
+      await destroy(destroySessionUrl);
+    } catch (e) {
+      return setFailed(true);
+    }
+    history.push('/');
```
sergei-maertens, line 44: Please provide useful user feedback instead of a generic error message.

* "Saving the data failed, please try again later"
* "Suspending the form failed, please try again later"
* "Logging out failed, please try again later"

etc.

src/components/modals/FormStepSaveModal.js
```diff
@@ -0,0 +1,115 @@
...
+  };
+
+  return (
+      <Modal
+        isOpen={isOpen}
+        closeModal={closeModal}
+      >
```
sergei-maertens, line 51: Please add a title prop, the header now looks weird and without context.

src/components/modals/FormStepSaveModal.js
```diff
@@ -0,0 +1,115 @@
...
+                />
+              </ErrorMessage>
+            )
+            : null
+          }
+
+          <Body modifiers={['big']}>
+            <FormattedMessage
+              description="Form save modal body text"
+              defaultMessage={"Enter your email address to get an email to resume the form at a later date"}
```
sergei-maertens, line 70:
```suggestion
              defaultMessage="Enter your email address to get an email to resume the form at a later date"
```

src/components/modals/FormStepSaveModal.js
```diff
@@ -0,0 +1,115 @@
...
+                description="Form save modal email field help text"
+                defaultMessage="The email address where you will received the email about resuming the form."
+              />
+            </HelpText>
+
+          </div>
+
+          <Toolbar modifiers={['bottom', 'reverse']}>
+            <ToolbarList>
+              <Button type="submit" variant="primary">
```
sergei-maertens, line 95: This button should be disabled while the submissions is happening (`put`, `post` and `delete` calls happening) to prevent double-submission. End-users _will_ double click.

Additionally, you want some kind of user-feedback while the backend is processing, so please display a loader after the user submitted.

src/components/FormStep.js
```diff
@@ -304,59 +309,70 @@ const FormStep = ({
...
+          ) : null
+        }
+      </Card>
+      <FormStepSaveModal
+        isOpen={shouldFormStepSaveModalBeOpen}
+        closeModal={closeFormStepSaveModal}
+        stepData={{...formData.current}}
+        saveStepDataUrl={submissionStep.url}
+        suspendFormUrl={`${submission.url}/_suspend`}
+        destroySessionUrl={`${config.baseUrl}authentication/session`}
```
sergei-maertens, line 373: fetch the `config` from the context in the downstream component rather than passing it as prop, this makes it easier to re-arrange the components and loosens coupling.

sergei-maertens (changes requested):

src/components/modals/Modal.js
shea-maykinmedia: Yes but won't this add `open-forms` to the start of the className, which we don't want since we're overriding the styles coming from React Modal?

src/components/modals/FormStepSaveModal.js
shea-maykinmedia: Do we still want to do this?  As far as I can see the `stepData` needs to be passed in separately which means we would still have five props (I removed the `destroySessionUrl` already) so I think we would have 5 props either way.

Two of the reasons I kept them as just the url:
1. It feels like overkill to pass an entire object when we only want one property of that object
2. It does seem like this would make it harder to reuse this modal in a different component since it'd make it more tightly coupled to the `FormStep` modal.

src/components/modals/Modal.js
shea-maykinmedia: Nevermind, I realise now that react modal does not come with it's own styling.  I'll have this updated in a sec.

src/components/modals/FormStepSaveModal.js
```diff
@@ -0,0 +1,187 @@
...
+  ] = useImmerReducer(reducer, initialState);
+
+  const onSubmit = async (event) => {
+    event.preventDefault();
+    dispatch({type: 'SET_IS_SAVING_TRUE'});
+    dispatch({type: 'SET_ERROR_MESSAGE', payload: ''});
+    let response = await put(saveStepDataUrl, {data: stepData});
+    if (!response.ok) {
+      dispatch({
+        type: 'SET_ERROR_MESSAGE',
```
sergei-maertens, line 81: here too - would use some sort of action type like `API_ERROR` and keep the payload. That the error message state is set is not semantically interesting, it's just a side effect, an implementation detail.

src/components/modals/FormStepSaveModal.js
```diff
@@ -0,0 +1,187 @@
...
+        payload: intl.formatMessage({
+          description: "Modal logging out failed message",
+          defaultMessage: "Logging out failed, please try again later"
+        })
+      });
+      dispatch({type: 'SET_IS_SAVING_FALSE'});
+      return;
+    }
+    dispatch({type: 'SET_IS_SAVING_FALSE'});
```
sergei-maertens, line 116:
```suggestion
    dispatch({type: 'SAVE_SUCCEEDED'});
```
& handle all the relevant side effects in there (= reset the state back to initial state, basically)

src/components/modals/FormStepSaveModal.js
```diff
@@ -0,0 +1,187 @@
...
+    history.push('/');
+  };
+
+  return (
+      <Modal
+        title={<FormattedMessage
+                  description="Form Save Modal Title"
```
sergei-maertens, line 123:
```suggestion
                  description="Form save modal title"
```

src/components/modals/FormStepSaveModal.js
```diff
@@ -0,0 +1,187 @@
...
+                  defaultMessage="Form Save"
```
sergei-maertens, line 124:
```suggestion
                  defaultMessage="Save data"
```

src/components/modals/FormStepSaveModal.js
```diff
@@ -0,0 +1,187 @@
...
+                defaultMessage="The email address where you will received the email about resuming the form."
+              />
+            </HelpText>
+
+          </div>
+
+          <Toolbar modifiers={['bottom', 'reverse']}>
+            <ToolbarList>
+              <Button type="submit" variant="primary" disabled={isSaving}>
+                <FormattedMessage description="Form save modal submit button" defaultMessage="Save Form & Logout" />
```
sergei-maertens, line 169:
```suggestion
                <FormattedMessage description="Form save modal submit button" defaultMessage="Save form & logout" />
```

src/components/modals/Modal.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ReactModal from 'react-modal';
+
+import FAIcon from 'components/FAIcon';
+import {FormattedMessage} from "react-intl";
+import {getBEMClassName} from "../../utils";
```
sergei-maertens, line 7:
```suggestion
import {getBEMClassName} from 'utils';
```
please use absolute imports

src/components/modals/Modal.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ReactModal from 'react-modal';
+
+import FAIcon from 'components/FAIcon';
+import {FormattedMessage} from "react-intl";
```
sergei-maertens, line 6:
```suggestion
import {FormattedMessage} from 'react-intl';
```

src/scss/components/_react-modal.scss
```diff
@@ -0,0 +1,72 @@
...
+
+.#{prefix(react-modal)} {
+
+  &__overlay {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+
+    position: fixed;
+    z-index: 10000000;
```
sergei-maertens, line 11: not sure if we need such a high number

src/scss/components/_react-modal.scss
```diff
@@ -0,0 +1,72 @@
...
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+
+    background-color: rgba(255, 255, 255, 0.75);
+  }
+
+  &__header {
+    margin-bottom: 10px;
```
sergei-maertens, line 21: should use the existing sass mixins/functions that are grid/variable driven instead of hardcoded values. Later we will upgrade this to design tokens

src/scss/components/_react-modal.scss
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  &__content {
+    position: absolute;
+    width: 50vw;
+    height: 50vh;
```
sergei-maertens, line 27: fairly certain this makes it look like trash on mobile

src/scss/components/_react-modal.scss
```diff
@@ -0,0 +1,72 @@
...
+    background: white;
+    border: solid 1px #eee;
+    border-radius: 4px;
+    outline: none;
+    padding: 20px;
```
sergei-maertens, line 32: more magic numbers... should use the existing sass mixins for spacing.

src/scss/components/_react-modal.scss
```diff
@@ -0,0 +1,72 @@
...
+    cursor: pointer;
+    opacity: 0.5;
+
+    &:hover {
+      opacity: 0.8;
+    }
+  }
+
+  &__title {
+    // taken from div#content h1 admin CSS styles
```
sergei-maertens, line 57: didn't know we have admin styles in the SDK :)

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #89, opened 2021-10-13
Upgrade formiojs

SilviaAmAm:
This upgrades formiojs to 4.13.7

We also used react-formio 4.3, which depended on formio 4.12.7. So I upgraded it to @formio/react 5.1.1, which depends on formio 4.13.7 (so that we don't have multiple versions of formiojs as a dependency)

As far as I can see, this does not solve the problem of flickering of the fields.

To do:

- [ ] Uploading files is broken
- [ ] Remove unnecessary code:
     - [ ]  https://github.com/open-formulieren/open-forms-sdk/blob/main/src/formio/components/FileField.js#L8 (This whole file can be deleted)
     - [ ] https://github.com/open-formulieren/open-forms-sdk/blob/main/src/formio/components/Email.js#L19


sergei-maertens:
We'll tackle this in a separate PR, and probably use that as the step to put our custom formio extension into a separate package.

PR #98, opened 2021-11-01
[DRAFT] :recycle: Replace SASS variables with CSS variables

stevenbal:
To support custom themes as defined by municipalities

TODO:
- [ ] Fix variables in `_buttons.scss` (colors, use `hsl` or `brightness` filters?)
- [ ] All uses of `scale_color`
- [x] Fix variables in `_checkbox.scss` (margins)
- [ ] Checked radio buttons / select boxes are currently always blue (svg)
- [ ] Document the vars that can be overridden


sergei-maertens:
Closing in favour of #202

PR #99, opened 2021-11-03
Display payment status

sergei-maertens:
Fixes open-formulieren/open-forms#724

I've decided to keep it simple and possibly we can expand and include the payment button again if payment is not succesful. Given that we wipe all the session/state after completion, it'd be quite a bit of work to fetch that information again, and maybe we can leverage the `status` api endpoint to also display pricing information, but that's all extra complexity.

Screenshots:

Payment cancelled

image

Payment succeeded

image

Banktransfer leads to "processing" status

image

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -1,9 +1,12 @@
 import React, {useContext} from 'react';
+import {useIntl} from 'react-intl';
 import { useImmerReducer } from 'use-immer';
 import {
-  Switch,
+  Redirect,
   Route,
+  Switch,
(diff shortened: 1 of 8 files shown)


SilviaAmAm (approved):
The code seems good! I didn't try it locally though, because I am not sure how to set up the payments.

It might also be nice to add some tests :smile:

sergei-maertens:
@SilviaAmAm tests added... what a pain :smile:

SilviaAmAm (approved):

PR #100, opened 2021-11-03
Map component fixes

sergei-maertens:
* Fixed initial browser location not persisting to step data
* Fixed map flicker and properly using the browser location


package.json
```diff
@@ -114,18 +115,11 @@
       ]
     }
   },
-  "browserslist": {
-    "production": [
-      ">0.2%",
-      "not dead",
-      "not op_mini all"
-    ],
-    "development": [
```
sergei-maertens, line 123: Workaround for https://github.com/PaulLeCam/react-leaflet/issues/891

SilviaAmAm (approved):
Looks good. Just some tiny remarks.

Also for the various Map components it might be good to add some tests :smile:

sergei-maertens:
> Also for the various Map components it might be good to add some tests

I honestly have no idea how to test that with it being Leaflet that does the rendering, but I can give it a try of course.

sergei-maertens:
Tried the whole testing with react-leaflet etc. but it really seems to not easily be possible with functional hook-based components and I'm really not looking to rewrite it into class-based components. Stack-overflow comes up empty as well, react-leaflet library author says to just mock the library completely which entirely beats the point of what we want to test.

No tests for this part then.

Issue #103, opened 2021-11-05
Multiple-value input doesn't work


SilviaAmAm:
How to reproduce:
- Make a form with a text field where the 'multiple values' checkbox is checked.
- Start the SDK and begin the form
- Enter some text in the text field and then click on 'Nog een toevoegen'

This gives the error in the console:
Uncaught TypeError: can't define array index property past the end of an array with non-writable length

PR #105, opened 2021-11-08
[#103] Multivalue input broken

SilviaAmAm:
This seems to have been caused by the fact that the `dataValue` property of the `multivalue` component gets filled with the submission data coming from the react state. For this reason, it was immutable and therefore any push/splice operations were failing.


src/jstests/formio/components/multiple.spec.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import _ from 'lodash';
+import {Formio} from "react-formio";
+import {multipleForm} from './fixtures/multiple';
+import OpenFormsModule from "../../../formio/module";
```
sergei-maertens, line 5:
```suggestion
import OpenFormsModule from '../../../formio/module';
```

src/jstests/formio/components/multiple.spec.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import _ from 'lodash';
+import {Formio} from "react-formio";
+import {multipleForm} from './fixtures/multiple';
```
sergei-maertens, line 4:
```suggestion

import {multipleForm} from './fixtures/multiple';
```

I'm trying to group library vs. own-code imports :)

src/jstests/formio/components/multiple.spec.js
```diff
@@ -0,0 +1,66 @@
...
+import OpenFormsModule from "../../../formio/module";
+
+jest.mock("../../../map/rd", () => jest.fn());
```
sergei-maertens, line 7: my PR #100 removes the need for this mock!

sergei-maertens (changes requested):
Hrm, is there an alternative way to do this without making a custom component?

I'm thinking if we transform the `data` from `{submission: {data}}` to be mutable, that might also resolve the bug maybe?

sergei-maertens (approved):

PR #108, opened 2021-11-18
Add change appointment functionality

shea-maykinmedia:
SDK PR for https://github.com/open-formulieren/open-forms/pull/954


shea-maykinmedia:
Closing since I don't think this code is needed

PR #109, opened 2021-11-18
Validate time on min/max value

SilviaAmAm:
Part of https://github.com/open-formulieren/open-forms/issues/902

### DIFF
--- a/src/formio/components/TimeField.js
+++ b/src/formio/components/TimeField.js
@@ -1,6 +1,7 @@
 import {Formio} from 'react-formio';
-import {getBEMClassName} from '../../utils';
+
 import MinMaxTimeValidator from 'formio/validators/MinMaxTimeValidator';
+import {getBEMClassName} from 'utils';
 
 const Time = Formio.Components.components.time;
 
--- a/src/formio/components/TimeField.js
+++ b/src/formio/components/TimeField.js
@@ -1,15 +1,17 @@
-import {Formio} from "react-formio";
+import {Formio} from 'react-formio';
 import {getBEMClassName} from '../../utils';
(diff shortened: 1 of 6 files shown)


src/jstests/formio/components/time.spec.js
```diff
@@ -0,0 +1,91 @@
+import React from 'react';
+import _ from 'lodash';
+import {Formio} from 'react-formio';
+import {timeForm} from './fixtures/time';
+import OpenFormsModule from '../../../formio/module';
+
+jest.mock('../../../map/rd', () => jest.fn());
```
sergei-maertens, line 7: I think this was no longer needed :)

src/formio/components/TimeField.js
```diff
@@ -1,15 +1,17 @@
-import {Formio} from "react-formio";
+import {Formio} from 'react-formio';
 import {getBEMClassName} from '../../utils';
+import MinMaxTimeValidator from 'formio/validators/MinMaxTimeValidator';
```
sergei-maertens, line 3:
```suggestion
import {Formio} from 'react-formio';

import MinMaxTimeValidator from 'formio/validators/MinMaxTimeValidator';
import {getBEMClassName} from 'utils';
```
let's tidy up our imports :)

sergei-maertens (changes requested):

src/jstests/formio/components/time.spec.js
SilviaAmAm: Aaah yes, bad copy and paste on my part! I will remove it also from the IBAN tests so I don't accidentally copy it anymore :D

sergei-maertens (approved):

PR #110, opened 2021-11-19
Show columns and fieldset in summary page

shea-maykinmedia:
Resolves https://github.com/open-formulieren/open-forms/issues/928

In the first screenshot case of two fieldsets where one comes after the other (like in the screenshot) there will be a double amount of space.  I spent about 30 mins trying to find a solution but thought I should get a second opinion on whether this should be fixed and if there was a possible solution.

I also moved some of the util functions that were only being used by `FormStepSummary` to be in the same directory as the component to make a distinction between util functions used throughout the project and ones only used by this component.

**Screenshots**

Screenshot 2021-11-19 at 14 16 43

**Columns and Fieldsets example with nested columns and nested fieldsets**

Step 1

Screenshot 2021-11-22 at 15 40 40

Step 2

Screenshot 2021-11-22 at 15 38 32

Summary Page

Screenshot 2021-11-22 at 15 39 21


sergei-maertens (approved):

sergei-maertens:
Final version
image

PR #112, opened 2021-11-22
Don't call alert by default

shea-maykinmedia:
Mentioned on Slack that we don't want to display this alert.  While nothing will currently use the alert I thought we should keep the code there in case we want it in the future, though we can also just remove it if that's preferred.


sergei-maertens (approved):
:adhesive_bandage:

PR #116, opened 2021-11-22
:sparkles: Add option to auto redirect user to login

stevenbal:
Fixes https://github.com/open-formulieren/open-forms/issues/1313
related PR in backend: https://github.com/open-formulieren/open-forms/pull/1429

### DIFF
--- a/src/components/FormStart/index.js
+++ b/src/components/FormStart/index.js
@@ -14,7 +14,7 @@ import { Toolbar, ToolbarList } from 'components/Toolbar';
 import useQuery from 'hooks/useQuery';
 import Types from 'types';
 import LoginButton, {
-  START_FORM_QUERY_PARAM, LoginButtonIcon
+  START_FORM_QUERY_PARAM, LoginButtonIcon, getLoginUrl
 } from 'components/LoginButton';
 import {getBEMClassName} from 'utils';
 
@@ -23,6 +23,19 @@ const useStartSubmission = () => {
   const query = useQuery();
   return !!query.get(START_FORM_QUERY_PARAM);
 };
+
+const getLoginRedirectUrl = (form) => {
+  // Automatically redirect the user to a specific login option (if configured)
+  if(form.autoLoginAuthenticationBackend) {
+    let autoLoginOption = form.loginOptions.find(
+      option => option.identifier === form.autoLoginAuthenticationBackend
+    );
+
+    if(autoLoginOption) {
+      return getLoginUrl(autoLoginOption)
+    }
+  }
+}
 
 const FormStartMessage = ({form}) => {
   const intl = useIntl();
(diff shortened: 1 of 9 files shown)


src/sdk.js
```diff
@@ -99,21 +101,35 @@ class OpenForm {
...
+
+    // Automatically redirect the user to a specific login option (if configured)
+    if(!doStart && formObject.autoLoginAuthenticationBackend) {
+      let loginUrl;
+      formObject.loginOptions.forEach(option => {
+        if(option.identifier === formObject.autoLoginAuthenticationBackend) {
+          loginUrl = getLoginUrl(option)
+          window.location = loginUrl;
+        }
+      });
```
sergei-maertens, line 114: you should be able to use `formObject.loginOptions.find(...)` to get the relevant auth option, instead of the `forEach`. You can the flatten the code a bit and make it more readable like that.

I would also structure this as a guard clause - if the login option is found, end with a `return`, but then move the `ReactDOM.render` out of the `else` again and remove the `else`.

So roughly the code would look like this:

```
if shouldRedirecToAutologin
  loginOption = formObject.loginOptions.find(...)
  if loginOption
    doRedirect
    return

doUsualReactRender
```

This makes it a bit more fail-safe in the sense that the form with login options will still render if someone is messing with query params and the relevant login option is not found.

sergei-maertens (changes requested):
I have a bit mixed feelings about doing this in `sdk.js` and not as a hook inside the main App component :thinking: 

@SilviaAmAm do you have any throughts about this?

SilviaAmAm:
@sergei-maertens I agree that it would be nicer to use a hook. Maybe in the `FormStart` component, since we already have the hook to look for the start param: https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/FormStart/index.js#L22  ?

src/components/FormStart/index.js
```diff
@@ -61,12 +74,17 @@ const FormStartMessage = ({form}) => {
...
+  const autoRedirectUrl = getLoginRedirectUrl(form);
   const outagePluginId = useDetectAuthenticationOutage();
   const authErrors = useDetectAuthErrorMessages();
   const hasAuthErrors = !!outagePluginId || !!authErrors;
 
   const onFormStartCalledRef = useRef(false);
 
+  if (!doStart && autoRedirectUrl) {
+    window.location = autoRedirectUrl;
+  }
```
sergei-maertens, line 86: @stevenbal w/r to your 'flickering' questions.

1. This code belongs in the `useEffect` hook - doing redirects in the render phase is an anti-pattern iirc, it should only return JSX
2. We have a [loader](https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/Loader.js)/spinner component you can display - if these conditions are met for auto-redirect, in theory you should be able to just do `return <Loader />`. That should help with the flicker.

sergei-maertens (changes requested):

src/components/FormStart/index.js
stevenbal: Silvia fixed this, now it does show the loading indicator (thanks @SilviaAmAm!)

sergei-maertens (approved):
Code looks great, thanks! I'm going to approve this but not merge it yet until the backend is updated, as this change requires Open Forms API version 1.1.0.

PR #117, opened 2021-11-24
Remove submission data prop and instead use event handler

sergei-maertens:
Setting the submission data as prop was used to process backend data
loading events declaratively (navigating between steps, logic
evaluation...). However, because the prop changed, this caused FormIO
to re-render too aggressively, breaking the cursor position and leading
to a frustrating experience while filling out non-standard fields (
such as time field, currency field...).

This has been removed, and we now sync the FormIO state on backend
data updates (initial step data loading, logic updates...). This
makes the form less jittery and more user-friendly.

Fixes open-formulieren/open-forms#876
Fixes open-formulieren/open-forms#878

Breaks ??? (<- we will find out...)

### DIFF
--- a/src/components/FormStep.js
+++ b/src/components/FormStep.js
@@ -5,11 +5,11 @@ import React, {useRef, useContext} from 'react';
 import PropTypes from 'prop-types';
 import {useIntl} from 'react-intl';
 import { useHistory, useParams } from 'react-router-dom';
+import cloneDeep from 'lodash/cloneDeep';
 import isEqual from 'lodash/isEqual';
 import isEmpty from 'lodash/isEmpty';
 import { useImmerReducer } from 'use-immer';
 import useAsync from 'react-use/esm/useAsync';
-import _ from 'lodash';
 
 import hooks from '../formio/hooks';
 
@@ -54,7 +54,7 @@ class AbortedLogicCheck extends Error {
 
 const initialState = {
   configuration: null,
-  data: null,
+  backendData: {},
   canSubmit: false,
   logicChecking: false,
   isFormSaveModalOpen: false,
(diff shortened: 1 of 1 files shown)


src/components/FormStep.js
```diff
@@ -296,6 +294,23 @@ const FormStep = ({
     history.push(navigateTo);
   };
 
+  const onFormIOInitialized = () => {
```
SilviaAmAm, line 297: In relation to one of the low-prio issues from sprint 9 (https://github.com/open-formulieren/open-forms/issues/974 => Logic not working with checkboxes). 

I think it is due to the fact that on initialisation there is no logic check done (so, for example, logic rules checking if a checkbox is unchecked don't run until something else is changed in the form). So maybe there should also be a check logic here in the `onFormIOInitialised`?

SilviaAmAm (changes requested):

src/components/FormStep.js
sergei-maertens: Hrm, but the `onFormIOInitialised` is triggered in a re-render because of the logic check too :thinking:

src/components/FormStep.js
```diff
@@ -329,10 +344,7 @@ const FormStep = ({
       LOGIC_CHECK_DEBOUNCE,
     );
 
-    dispatch({
-      type: 'STEP_DATA_UPDATED',
-      payload: _.cloneDeep(data),
-    });
+    dispatch({type: 'STEP_DATA_UPDATED'});
```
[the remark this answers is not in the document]
sergei-maertens: Ah shit, I'll set this up in my test form and verify!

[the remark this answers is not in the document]
sergei-maertens: I've added back the cloneDeep and added a comment why it's needed, works in my env now.

src/components/FormStep.js
sergei-maertens: I've created open-formulieren/open-forms#994 for this, let's tackle that separately

PR #121, opened 2021-12-01
API can return null for payment amount

sergei-maertens:
Fixes open-formulieren/open-forms#235


SilviaAmAm (approved):
Thank you for fixing this, so that there is no warning message anymore in the console :D

PR #125, opened 2021-12-06
Co-sign Formio/React component

sergei-maertens:
Fixes open-formulieren/open-forms#296


SilviaAmAm (changes requested):
Couple of comments (also discussed on Slack):

- Currently, using the coSign component with the DigiD Mock doesn't seem to work, because the return URL is not parsed properly and so the submission_uuid is not recognised (which gives a 403 error). This did not seem to be fixes by the front end url clean-up :(
- src/openforms/prefill/contrib/haalcentraal/plugin.py:97 => If the fields are not returned from Haal centraal, then this will raise an error

PR #138, opened 2022-01-07
[#1163] Fixed auth options icon margin

Bartvaderkin:
This change makes the icons separate children of the list instead of all being one item, so they uses the same CSS rule as the label buttons.

Refs #1163


sergei-maertens (approved):
Interesting fix, apparently the `ToolbarList` can't handle fragments. Nice catch!

PR #139, opened 2022-01-07
[#1162] Fixed BSN field validator validating everything not 9 chars long

Bartvaderkin:
Fixes https://github.com/open-formulieren/open-forms/issues/1162

### DIFF
--- a/src/formio/components/BsnField.js
+++ b/src/formio/components/BsnField.js
@@ -10,10 +10,7 @@ const BsnValidator = {
     });
   },
   check(component, setting, value) {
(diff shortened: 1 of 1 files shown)


Bartvaderkin:
Note this will now start invalidating as soon as you type the first digit.

sergei-maertens:
Hrm, I think we should set validation to happen on blur then

sergei-maertens:
I'm approving this in the context of #1171 - the validation trigger should be specified by the form designer, where they can pick between 'change' and 'blur' events.

sergei-maertens (approved):

PR #142, opened 2022-01-12
More fixes w/r to Formio data immutability

sergei-maertens:
open-formulieren/open-forms#1174

There were more places were refs were updated with data that possibly
contained immutable datastructures (because they come from the
immer reducer state). These have now been wrapped with cloneDeep to
make them immutable as well, so that Formio can apply the mutations.

This typically happens when you have two fields for example: a regular
text field, and a multiple=true text field. If you edit the regular
text field, some immutable data ends up in the state, preventing you
from adding extra items to the multiple=true text field.

Additionally, the onFormIOChange event handler has been adapted, as
it did not update the internal view on the data after adding an
item to a multiple=true text field. This is caused by the
modifiedByHuman flag being 'undefined' and the flags containing
'undefined' as values as well, making it unreliable. Instead, we now
compare if the new view on the data is different from the old view
before we effectively detect a change in data.

Finally, some bugs are likely fixed causing the Formio spinner to
spin indefinitely, or the submit button being blocked. We now dispatch
an action when the logic check is aborted or not even performed due
to Formio-level validation errors. An extra side-effect is that invalid
client side forms are now blocked from submission (client-side).

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -151,6 +151,8 @@ const reducer = (draft, action) => {
   const onFormStart = async (event) => {
     event && event.preventDefault();
 
+    // required to get rid of the error message saying the session is expired - once
+    // you start a new submission, any previous call history should be discarded.
     resetSession();
 
     if (state.submission != null) {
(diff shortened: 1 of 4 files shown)


Bartvaderkin (approved):
I tested this locally and pushed a commit with a fix for the `.map is not a function` issue.

PR #150, opened 2022-01-20
[OF#1152] Time validation

SilviaAmAm:
Fixes https://github.com/open-formulieren/open-forms/issues/1152


joeribekker (approved):
Thanks for the tests, that makes it clear this works as described in the issue :)

PR #151, opened 2022-01-21
[OF#1224] files

SilviaAmAm:
Fixes https://github.com/open-formulieren/open-forms/issues/1224
Fixes https://github.com/open-formulieren/open-forms/issues/1226

Note 1224:
The problem seems to be that some components with `multivalue` have as default value `[undefined]`. This broke the summary page.

Note 1226:
The main problem seemed to be that if components are inside hidden fieldsets, then they have a value `undefined` that broke the summary page.

### DIFF
--- a/src/components/FormStepSummary/ComponentValueDisplay.js
+++ b/src/components/FormStepSummary/ComponentValueDisplay.js
@@ -153,6 +153,8 @@ const FileDisplay = ({component, value}) => {
 
 
 const NumberDisplay = ({component, value}) => {
+  if (!value) return <EmptyDisplay/>;
+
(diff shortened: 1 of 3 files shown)


joeribekker (approved):
Had a discussion with @SilviaAmAm about this ticket since the pattern `if (!value) return <EmptyDisplay/>;` is repeated quite alot.

Another approach: Let it fail, catch failure and insert EmptyDisplay on rendering - This can still cause warnings probably due to the use of undefined in functions.

The current approach allows for more finegrained situations when to bail early and fixes real problems. Therefore, accepting as-is.

PR #154, opened 2022-01-27
[OF#1254] Columns go outside the page

SilviaAmAm:
Fixes open-formulieren/open-forms#1254

Screenshot from 2022-01-27 15-03-37


sergei-maertens:
I wonder if this is related to the note mentioned in 38ab8d99394ef86fb88639378ece32a20af864ad?

> Known issue: responsive widths aren't respected yet, now columns just
always display as a column regardless the screen size.

PR #159, opened 2022-02-03
[OF#1261] Fix privacy-checkbox styling

SilviaAmAm:
Screenshot from 2022-02-03 09-30-40
Fixes open-formulieren/open-forms#1261

### DIFF
--- a/src/scss/components/_privacy-checkbox.scss
+++ b/src/scss/components/_privacy-checkbox.scss
@@ -1,16 +1,19 @@
 @import '../mixins/prefix';
+@import '~microscope-sass/lib/typography';
 
 .#{prefix(privacy-checkbox)} {
   @extend .#{prefix(checkbox)};
 
   &__label {
     @extend .#{prefix(checkbox)}__label;
+  }
 
+  &__wyswyg-text {
     // This is needed because the label is populated with the value coming from a WYSIWYG editor
-    // which adds <p></p> around the label text. This tag gets a margin which looks off.
+    // which adds <p></p> around each paragraph in the text. This tag gets a margin which looks off.
     // So here it is overwritten.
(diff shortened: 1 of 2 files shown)


src/scss/components/_privacy-checkbox.scss
```diff
@@ -5,13 +5,17 @@
...
     // This is needed because the label is populated with the value coming from a WYSIWYG editor
-    // which adds <p></p> around the label text. This tag gets a margin which looks off.
+    // which adds <p></p> around each paragraph in the text. This tag gets a margin which looks off.
     // So here it is overwritten.
     p {
-      margin: 0;
+      margin: 0 0 1em 0;
     }
+
+    p:last-child { margin-bottom: 0 !important; }
```
sergei-maertens, line 18: You can simplify this using the [`body` mixin](https://maykinmedia.github.io/microscope-sass/#typography-mixin-body) mixin

```scss
p {
    @include body('auto');  // sets margin-bottom automatically except for the last child
}
```

The advantage is that it doesn't use any magic numbers (`1em`) and instead uses the globally configured spacing configuration. If that is too much (I think it's 20px?) you can use the lower level `@include margin('auto', $value-mobile: 1em);` instead.

sergei-maertens (changes requested):

src/scss/components/_privacy-checkbox.scss
SilviaAmAm: I think it looks okay with the margin given by `@include margin('auto')`, so i didn't use the smaller value! :slightly_smiling_face: 

Screenshot from 2022-02-03 16-57-48

sergei-maertens (approved):

PR #160, opened 2022-02-04
Fixes more event/race conditions while typing

sergei-maertens:
Fixes open-formulieren/open-forms#1274

As suspected already, the formData ref can be refactored out, since the
formRef (if set) contains the same information (the submission data).

The updating of the formData ref lagged behind one tick to what the
actual submission data value was. This _can_ cause race conditions in
the logic check handling where values are set based on formData ref.

The surface for race conditions is now further reduced by relying on
the Formio instance in all cases.

I could reproduce this by putting some `await sleep(5000)` between lines of code that are not IO bound. Hopefully this fixes it, as this is otherwise rather hard to reproduce.

### DIFF
--- a/src/components/FormStep.js
+++ b/src/components/FormStep.js
@@ -68,7 +68,7 @@ const reducer = (draft, action) => {
       draft.logicChecking = false;
       break;
     }
-    case 'STEP_DATA_UPDATED': {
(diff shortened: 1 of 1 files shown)


sergei-maertens:
I also manually checked that `multiple` inputs are not broken again because of this :)

PR #162, opened 2022-02-09
Better file upload errors styling

sergei-maertens:
Fixes open-formulieren/open-forms#1193
Fixes open-formulieren/open-forms#1316

There are some mistakes in the core templates, so it's time to override it anyway. In the future we can use this to properly style this to our liking, as the bootstrap markup/classes makes this hard and annoying to deal with.

image

### DIFF
--- a/src/formio/templates/file.ejs
+++ b/src/formio/templates/file.ejs
@@ -88,21 +88,17 @@
       </div>
       <div class="fileSize col-form-label col-sm-2 text-right">{{ctx.fileSize(status.size)}}</div>
     </div>
-    <div class="row">
-      <div class="col-sm-12">
-        {% if (status.status === 'progress') { %}
-          <div class="progress">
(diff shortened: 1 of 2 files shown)


SilviaAmAm (changes requested):
I notice a few problems when trying out the file field in the form:
- It is now possible to upload multiple files in a file field without a 'multiple' selected.
- If the name of the file is very long, then it looks funny (see screenshot 1)
- If the file component is inside a field set and the name of the updated file is long, the whole field set goes off the page  (see screenshot 2). This problem though was pre-existing: https://github.com/open-formulieren/open-forms/issues/1316


Screenshot from 2022-02-10 13-31-22

Screenshot from 2022-02-10 13-31-57

sergei-maertens:
@SilviaAmAm I think I've fixed fieldsets, can you confirm?

sergei-maertens:
Default formio behaviour seems to be that the file is left there and you have to remove it itself by clicking the cross, and other files dragged into the box are validated/uploaded separately.

SilviaAmAm (approved):

joeribekker (approved):

PR #173, opened 2022-03-02
Hide fieldset header if configured

SilviaAmAm:
Part of open-formulieren/open-forms#1391


sergei-maertens (approved):

src/formio/templates/fieldset.ejs
```diff
@@ -1,7 +1,7 @@
-<fieldset class="{{ctx.ofPrefix}}fieldset">
-  {% if (ctx.component.legend) { %}
+<fieldset class="{{ctx.ofPrefix}}fieldset {% if (!ctx.component.showHeader) { %}{{ctx.ofPrefix}}fieldset--no-header{% } %}">
+  {% if (ctx.component.label && ctx.component.showHeader) { %}
```
sergei-maertens, line 2: as discussed to prevent breaking existing form configs, invert with `hideHeader` instead of `showHeader`

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #177, opened 2022-03-09
[OF#1404] Required fields without asterisks

SilviaAmAm:
Part of https://github.com/open-formulieren/open-forms/issues/1404

### DIFF
--- a/src/scss/components/_form-control.scss
+++ b/src/scss/components/_form-control.scss
@@ -42,6 +42,21 @@
     display: block;
   }
 
+  &--no-asterisks {
+    .required-field {
+      // Override the asterisk
+      .#{prefix(label)}:after {
+        content: "" !important;
+      }
+    }
+
+    .optional-field {
+      .#{prefix(label)}:after {
+        content: "(optioneel)";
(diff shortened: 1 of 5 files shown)


src/scss/components/_form-control.scss
```diff
@@ -42,6 +42,21 @@
...
+    .required-field {
+      // Override the asterisk
+      .#{prefix(label)}:after {
+        content: "" !important;
+      }
+    }
+
+    .optional-field {
+      .#{prefix(label)}:after {
+        content: "(optioneel)";
```
sergei-maertens, line 55: I'd rather see this handled in the label template so we can properly apply translations - if we support English localization (which we already sort of do), then this will look out of place.

sergei-maertens (changes requested):

src/scss/components/_form-control.scss
SilviaAmAm: Yes, this was lazy :stuck_out_tongue:

sergei-maertens (approved):

PR #183, opened 2022-03-25
[Draft] :bug: Fix auto login redirect for embedded forms

stevenbal:
The automatic redirect to login doesn't seem to work for embedded forms. This is caused by the fact that the `next` url to which the user should be redirected after login is now determined later (I think? I'm not really sure how these hooks work. In my initial pull request, `getLoginUrl` was called in `sdk.js`, now this happens in `Form.js`).

Locally for instance, when embedding a form `bezwaar-maken` on path `/`, I would be redirected to `/bezwaar-maken/?_start=1`, which will simply return a 404, because this path doesn't exist.

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -110,7 +110,6 @@ const reducer = (draft, action) => {
  */
  const Form = ({ form }) => {
   const history = useHistory();
-  const shouldAutomaticallyRedirect = useAutomaticRedirect(form);
   const queryParams = useQuery();
   usePageViews();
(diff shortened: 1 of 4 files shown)


stevenbal:
@sergei-maertens I managed to fix this issue locally like this, but it doesn't seem like a good solution :stuck_out_tongue: 

Do you have any ideas perhaps?

stevenbal:
Closing this, as this issue should be fixed by having a catch all URL pattern in the application that embeds OF

PR #190, opened 2022-04-21
Fix broken logic state for formstep submission

sergei-maertens:
Fixes open-formulieren/open-forms#1494

Given that:
 
 1. a form was in a submittable state
 2. user input in the form step is done in quick succession but the net
    end result is no changes (e.g. radio yes/no toggle to the other
    option and quickly back to the originally selected option)
 
 Then the change handler caused the form submission to be blocked as
 a safeguard against bypassing logic checks.
 
 The logic check in turn shortcutted because the data being compared
 was not different than the originally checked data, so no changes in
 state are suspected.
 
 However, this shortcut/guard clause failed to restore the original
 canSubmit state of the form step, which was set to false by the
 change handler.
 
 This patch ensures that on early-exits of logic checks, the state is
 correctly restored. Additionally, the canSubmit state for aborted
 logic checks now no longer only depends on form validity, but also
 on the previous state. Even though a client-side form is valid,
 it may still be blocked from submission by logic rules.

### DIFF
--- a/src/components/FormStep.js
+++ b/src/components/FormStep.js
@@ -1,5 +1,25 @@
 /**
  * Render a single form step, as part of a started submission for a form.
+ *
+ * Functional requirements:
+ *
+ *   - the backend form configuration must be displayed when a new step is loaded
+ *   - users must be able to fill out the form step
+ *   - whenever there is user input, this requires to be validated/checked by both
+ *     frontend and backend
+ *   - the default reaction for (new) user input is blocking the submit button
+ *   - as long as there is any unvalidated/unchecked input, the submit button must remain
+ *     blocked
+ *   - while the submit button is blocked, we run client-side validation. if that fails,
+ *     the submit button remains blocked
+ *   - user input schedules a backend logic check. after this is completed, the submit
+ *     button state is updated with the result from the backend
+ *   - backend logic checks are cancelled/debounced on new user input
+ *   - backend logic checks are skipped if the input data is empty or unchanged from the
+ *     last check that did execute
+ *   - on quick user input changes with a net zero result, the submit button must be
+ *     restored to its original state
+ *
  */
 import React, {useRef, useContext} from 'react';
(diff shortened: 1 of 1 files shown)


Bartvaderkin:
I tested this and while the Next button-state does update to the expected enabled/disabled state there is a small ~1 second window where a race-condition allows that after I select the option that would disable the button I can quickly click the still enabled button and proceed to the next step.

sergei-maertens:
Did some more tweaking and passing around data from older events, I think it's now alright.

It did highlight for me again the complexity of this state management and the need to refactor this to make the code understandable again.

Bartvaderkin (approved):
I tested the new changes with my reproduction and my earlier issue with the debounce seems fixed.

src/components/FormStep.js
```diff
@@ -395,7 +440,15 @@ const FormStep = ({
...
 
-    dispatch({type: 'BLOCK_SUBMISSION'});
+    // backend logic leads to changes in FormIO configuration, which triggers onFormIOInitialized.
+    // This in turn triggers the onFormIOChange event because the submission data is set
+    // programmatically. Without checking for human interaction, this would block the
+    // submission again for LOGIC_CHECK_DEBOUNCE ms, for the logic check to eventually
+    // be interrupted inside the evaluateFormLogic handler because the data hasn't
+    // changed. We can skip this particular block-unblock cycle by only blocking the
+    // submission because of human input.
+    modifiedByHuman && dispatch({type: 'BLOCK_SUBMISSION'});
```
Bartvaderkin, line 450: not a huge fan of this boolean construct over a simple visible if-statements

SilviaAmAm (approved):

src/components/FormStep.js
sergei-maertens: changed!

sergei-maertens:
Backported via 1bdf03a9d20887113e0cdb9d57b1c255b3d8436f

PR #191, opened 2022-04-21
Do not give the illusion that the login button icon is a button

sergei-maertens:
Fixes open-formulieren/open-forms#1523

The cursor is now the default instead of pointer, and if you do click
the icon, the referenced href is opened in a new tab instead.


SilviaAmAm (approved):
Maybe we should make an issue of the FIXME comment?

sergei-maertens:
> Maybe we should make an issue of the FIXME comment?

I think it'll happen naturally when we look more into NL DS components!

PR #196, opened 2022-05-02
File upload component improvements

SilviaAmAm:
Fixes https://github.com/open-formulieren/open-forms/issues/1555
Translations: https://github.com/open-formulieren/open-forms/pull/1557

While uploading:
Screenshot from 2022-05-02 12-13-21


sergei-maertens (approved):
This probably also fixes some CSP reports of inline styles, so pretty happy with it.

PR #198, opened 2022-05-02
[OF#1533] Infinite redirects

SilviaAmAm:
Fixes open-formulieren/open-forms#1533

The problem was that the hook `useAutomaticRedirect` was not checking the current location. So on any pathname without the `_start=1` queryparam it was redirecting to login.

### DIFF
--- a/src/hooks/useAutomaticRedirect.js
+++ b/src/hooks/useAutomaticRedirect.js
@@ -1,20 +1,32 @@
-import {useEffect} from 'react';
+import {useContext, useEffect} from 'react';
+import {useLocation} from 'react-router-dom';
 
 import useStartSubmission from './useStartSubmission';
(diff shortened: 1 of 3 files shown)


src/hooks/useAutomaticRedirect.js
```diff
@@ -1,14 +1,18 @@
...
+import {useLocation} from 'react-router-dom';
 
 import useStartSubmission from './useStartSubmission';
 import {getLoginRedirectUrl} from 'components/utils';
 
 
 const useAutomaticRedirect = (form) => {
+  const location = useLocation();
+
+  const isStartPage = location.pathname === "/";
```
sergei-maertens, line 11: This probably doesn't work on subpaths :grimacing:

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #201, opened 2022-05-10
Fix DELETE call not being made for removed file uploads

sergei-maertens:
Fixes open-formulieren/open-forms#1539

Sigh... Formio pls.

So much global stuff leaks into component behaviour that isn't even
necessary.


sergei-maertens:
Backported in f41462d96b6a7f1d972820eb6e5d09b641929589

PR #203, opened 2022-05-11
Upgrade CRA build tooling

sergei-maertens:
- [x] Revert eject
- [ ] Upgrade to newest version of react-scripts
- [ ] Apply build changes again


package.json
```diff
@@ -3,98 +3,88 @@
...
+    "webpack": "^5.64.4",
+    "webpack-dev-server": "^4.6.0",
+    "webpack-manifest-plugin": "^4.0.2",
+    "workbox-webpack-plugin": "^6.4.1"
   },
   "scripts": {
     "start": "node scripts/start.js",
     "build": "node scripts/build.js",
-    "test": "node scripts/test.js --modulePaths=src",
+    "test": "node scripts/test.js",
```
sergei-maertens, line 87: Check if --modulePaths is still need (there's config in package.json now)

sergei-maertens: doesn't look like it with the `package.json` entry!

PR #209, opened 2022-06-13
Always call backend logic, but only with valid data

sergei-maertens:
Fixes open-formulieren/open-forms#1526

On input data change, the backend logic evaluation is now always
performed (i.e. even if the client-side validation reports an
invalid form). However, only the data that validated is sent to the
backend.

This fixes situations where the backend form logic is required to
resolve the client-side validation errors, typically when a former
form field is returned to which triggers some logic.


SilviaAmAm (approved):

sergei-maertens:
Backported in 2b6887f05993f15e6d0978f76bc934dd7a9bff33 and 534c1a8e656be2030b5937039c97e8a0c973dd7e

PR #211, opened 2022-06-30
Implement warning for session expiration — [#509]

dariomory:
Fixes open-formulieren/open-forms#509


src/components/Sessions/RequireSession.js
```diff
@@ -1,42 +1,77 @@
...
+    {display: false, accepted: false});
+
+  useEffect(() => {
+    if (expiryDate) {
+      const ms = untilWarningMilliseconds(expiryDate);
+      const timeout = setTimeout(() => {
+        setExpiryWarning({display: true});
+      }, ms);
+      return () => clearTimeout(timeout);
+    }
```
sergei-maertens, line 32: there might be an edge case here where the component is unmounted but there are still stale references and `setExpiryWarning` may still be called. Though that should probably clear up with the `clearTimeout`.

I also suspect that multiple hook calls can be scheduled to set the warning display to true for every render of this component where it does not get unmounted :thinking:

src/components/Sessions/RequireSession.js
```diff
@@ -1,42 +1,77 @@
...
 import PropTypes from 'prop-types';
 import {Link} from 'react-router-dom';
 import {FormattedMessage} from 'react-intl';
 
 import Anchor from 'components/Anchor';
 import Card from 'components/Card';
 import ErrorMessage from 'components/ErrorMessage';
+import Modal from "components/modals/Modal";
 
-// import SessionExpiry from './SessionExpiry';
```
sergei-maertens, line 10: code is not used anywhere and you removed the last reference to the comment, so that component can be deleted alltogether

sergei-maertens (changes requested):
This isn't quite there yet in its current form:

image

* The modal styling is not consistent at all with the rest of the page - the title is huge and there's a lot of blank space. We can definitely present a nicer message.
* It just statically displays "your session will expiry in XX seconds" once it triggers, but the value does not count down, so by the time you see this message you actually don't have a clue how much time is left before the session expires
* There are no buttons in the modal to be able to actually mitigate the situation. _How_ do I actually extend the session as an end user? How do I confirm that I want to log out/let it expire?

Please take a look at the backend implementation of this exact same functionality, it also already tackled a lot of edge cases with the `useEffect` hooks and uses the relative time delta formatter from react-intl properly.

I'm also not seeing the mechanism to actually extend the session, am I missing that somewhere?

src/components/Sessions/RequireSession.js
```diff
@@ -1,42 +1,113 @@
...
+        }}
+      >
+        <ErrorMessage>
+          <FormattedMessage
+            description="Session expiry warning message (in modal)"
+            defaultMessage="Your session will expire in {seconds} seconds. Extend your session if you wish to continue."
+            values={{
+              seconds: expiryWarning.seconds,
+            }}/>
+        </ErrorMessage>
```
sergei-maertens, line 94: This should be a warning message rather than an error. No error has occurred yet, but there's one about to.

src/components/Sessions/RequireSession.js
```diff
@@ -1,42 +1,124 @@
...
+
+  useEffect(() => {
+    /*
+    Count down to the session expiry time.
+     */
+    const interval = setInterval(() => {
+      setSecondsToExpiry(Math.floor((expiryDate - new Date()) / 1000));
+    }, 1000);
+    return () => clearInterval(interval);
+  }, [expiryDate]);
```
sergei-maertens, line 61: This is quite a naive implementation because you're effectively doing a state update and component re-render every second, while we know that the warning only needs to be triggered in `90% of the remaining time`. With a session timeout of 15 minutes, that's 810 state updates that should have been only 1.

src/scss/components/_alert.scss
```diff
@@ -20,6 +20,14 @@
     }
   }
 
+  @include modifier('warning') {
+    background-color: #fff3cd;
```
sergei-maertens, line 24: This has to be a design token, not a hardcoded color.

src/components/Sessions/RequireSession.js
```diff
@@ -1,42 +1,124 @@
...
-          defaultMessage="Your session has expired. Click <link>here</link> to restart."
-          values={{
-            link: chunks => <Link to="/" component={Anchor}>{chunks}</Link>,
-          }}
-        />
-      </ErrorMessage>
-    </Card>
-  );
+    <>
+      <Modal
```
sergei-maertens, line 87: Hovering over the close icon displays `[Object object]`, so something is going wrong with the translations/title props there.

sergei-maertens (changes requested):

sergei-maertens:
I also noticed during testing that I can dismiss the warning only once, after that if the session is about to expire again, I don't get a modal/warning at all anymore.

PR #212, opened 2022-07-01
Handle unusable autofilled field — [open-formulieren/open-forms#1717]

dariomory:
Fixes open-formulieren/open-forms#1717

### DIFF
--- a/src/formio/components/TextField.js
+++ b/src/formio/components/TextField.js
@@ -47,14 +47,18 @@ class TextField extends Formio.Components.components.textfield {
     const isValidPostcode = POSTCODE_REGEX.test(data[this.component.derivePostcode]);
 
     if (isValidHouseNumber && isValidPostcode) {
-      if (this.component.deriveStreetName) {
+      // Fill data if it is not set yet or if the field is readonly (i.e. Formio's disabled).
+      // Unrelated to the HTML 'disabled' attribute.
+      const mayAutofillValue = !this.getValue() || this.component.disabled;
+
+      if (this.component.deriveStreetName && mayAutofillValue) {
         this.setLocationData(
           data[this.component.derivePostcode],
           data[this.component.deriveHouseNumber],
           'streetName'
         );
       }
-      if (this.component.deriveCity) {
+      if (this.component.deriveCity && mayAutofillValue) {
         this.setLocationData(
           data[this.component.derivePostcode],
           data[this.component.deriveHouseNumber],
--- a/src/jstests/formio/components/textfield.spec.js
+++ b/src/jstests/formio/components/textfield.spec.js
@@ -32,8 +32,10 @@ describe('TextField Component', () => {
     }).catch(done);
   });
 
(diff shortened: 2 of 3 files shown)


sergei-maertens (changes requested):
Formio dispatches change events for every change, both by humans and automatically (because of logic for example). I'm not sure exactly how or where, but I think that we should be able to tap into these events and keep track of manual changes (by end users) to prevent overwriting those values again by changed addresses.

I find the entire behaviour ambiguous though - what should happen if the end user fills out a postal code + house number, modifies the city/street and then changes the postal code and/or house number? Should it autofill? Should it overwrite? I'm also wondering if/how we can leverage the `readOnly` option at form-design time and maybe base behaviour off that? Maybe we should only set the value from autofill if the current value is empty and then not touch it anymore?

src/jstests/formio/components/textfield.spec.js
```diff
@@ -72,7 +72,7 @@ describe('TextField Component', () => {
     }).catch(done);
   });
 
-  test('TextField with address prefill refreshes street on invalid data', (done) => {
+  test('TextField with address prefill refreshes street if already filled', (done) => {
```
sergei-maertens, line 75: * Please make it explicit that this only applies to editable textfields
* The behaviour is now the opposite of what the test name is - it should NOT refresh the street

sergei-maertens (changes requested):
The tests you modified should be _new_ tests, because you have now effectively deleted the tests for what happens when the API endpoint returns an empty object.

src/formio/components/TextField.js
```diff
@@ -26,7 +26,7 @@ class TextField extends Formio.Components.components.textfield {
     return info;
   }
 
-  checkComponentValidity(data, dirty, row, options = {}){
+  checkComponentValidity(data, dirty, row, options = {}) {
```
sergei-maertens, line 29: Revert this, this has nothing to do with the actual bug or bug fix. See earlier remarks about prettier.

src/formio/components/TextField.js
```diff
@@ -47,20 +47,23 @@ class TextField extends Formio.Components.components.textfield {
     const isValidPostcode = POSTCODE_REGEX.test(data[this.component.derivePostcode]);
 
     if (isValidHouseNumber && isValidPostcode) {
-      if (this.component.deriveStreetName) {
+
+      // Fill data if it is not set yet or if the field is readonly (disabled)
```
sergei-maertens, line 51:
```suggestion
      // Fill data if it is not set yet or if the field is readonly. Note that Formio's "disabled" effectively means
      // readonly - do not confuse this with the HTML 'disabled' attribute/behaviour.
```

sergei-maertens (changes requested):
You still have effectively removed the tests that check/validate what happens when the API endpoint returns an empty response.

It cannot be that behaviour/logic receives new nuances/variants and that the test count remains the same.

sergei-maertens (approved):

PR #215, opened 2022-07-08
[OF#1710] Repeating groups

SilviaAmAm:
Part of https://github.com/open-formulieren/open-forms/issues/1710

The summary page now looks like this:
Screenshot from 2022-07-13 17-47-28

After writing the renderer in the backend, im in doubt if I should have gone for the same approach (i.e. the fields being displayed under rather than next to the label) :thinking:

### DIFF
--- a/src/components/FormStepSummary/SummaryEditGrid.js
+++ b/src/components/FormStepSummary/SummaryEditGrid.js
@@ -1,0 +1,81 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Utils as FormioUtils} from 'formiojs';
+import _ from 'lodash';
+
+import {getComponentLabel} from './utils';
+import {getBEMClassName} from '../../utils';
+import {TableCell, TableHead, TableRow} from '../Table';
+import ComponentValueDisplay from './ComponentValueDisplay';
+
+const ItemInGroup = ({component, value, isLast=false}) => {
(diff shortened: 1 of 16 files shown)


SilviaAmAm:
Now it looks more similar to the backend:
Screenshot from 2022-07-14 10-57-00

src/components/FormStepSummary/SummaryEditGrid.js
```diff
@@ -0,0 +1,88 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Utils as FormioUtils} from 'formiojs';
+import _ from 'lodash';
+
+import {getComponentLabel} from './utils';
+import {getBEMClassName} from '../../utils';
```
sergei-maertens, line 7: we can use "absolute" imports (relative to the `src` directory) so we don't end up in path-hell:

```suggestion
import {getBEMClassName} from 'utils';
```

src/components/FormStepSummary/SummaryEditGrid.js
```diff
@@ -0,0 +1,88 @@
...
+import {TableCell, TableHead, TableRow} from '../Table';
```
sergei-maertens, line 8:
```suggestion
import {TableCell, TableHead, TableRow} from 'components/Table';
```

PR #217, opened 2022-07-21
Handle HTTP 403 errors (and others)

sergei-maertens:
This PR adds some robustness for backend errors with special attention for permission errors (HTTP 403).

* ensure that components that talk to the backend are wrapped in the error boundary
* throw HTTP 400-599 errors as exceptions
* track backend errors in component callbacks/event handlers/async calls
* display relevant information to the end-user

Example HTTP 403 error during login:

image

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -258,51 +258,57 @@ const reducer = (draft, action) => {
         <Switch>
 
           <Route exact path="/">
-            <ErrorBoundary>
+            <ErrorBoundary useCard>
               <FormStart form={form} onFormStart={onFormStart} />
             </ErrorBoundary>
           </Route>
 
           <Route exact path="/overzicht">
-            <RequireSession expired={sessionExpired} expiryDate={expiryDate}>
-              <RequireSubmission
-                submission={state.submission}
-                form={form}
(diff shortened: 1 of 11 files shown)


src/components/Form.js
```diff
@@ -258,32 +258,36 @@ const reducer = (draft, action) => {
...
+              <RequireSubmission
+                submission={state.submittedSubmission}
+                statusUrl={state.processingStatusUrl}
+                onFailure={onProcessingFailure}
+                onConfirmed={() => dispatch({type: 'PROCESSING_SUCCEEDED'})}
+                component={SubmissionConfirmation} />
+              </ErrorBoundary>
           </Route>
 
           <Route exact path="/betaaloverzicht">
```
SilviaAmAm, line 293: This route now has only `<ErrorBoundary>` without the `useCard` prop. Is it intentional?

sergei-maertens: ehm no, the `useCard` prop is present - though it looks like I messed up the indentation which probably caused the confusion!

sergei-maertens:
Backported to `stable/1.0.x` and `stable/1.1.x`

PR #220, opened 2022-08-23
Fix location autofill flicker

sergei-maertens:
(Hopefully) fixes open-formulieren/open-forms#1832

* Only query location data when field is read-only or field has no value yet
* Debounce the location data querying

The backend has a patch which adds caching to further reduce the amount of requests sent to the BAG API.

Needs backporting to stable/1.0.x as well, and the debounce functionality should probably be ported to the main branch.

### DIFF
--- a/src/formio/components/TextField.js
+++ b/src/formio/components/TextField.js
@@ -1,4 +1,5 @@
 import { Formio } from 'react-formio';
+import debounce from 'lodash/debounce';
 
 import { applyPrefix } from '../utils';
 import { get } from '../../api';
@@ -8,6 +9,7 @@ import enableValidationPlugins from "../validators/plugins";
 const POSTCODE_REGEX = /^[0-9]{4}\s?[a-zA-Z]{2}$/;
 const HOUSE_NUMBER_REGEX = /^\d+$/;
 
+const LOCATION_AUTOCOMPLETE_DEBOUNCE = 200;
 
 /**
  * Extend the default text field to modify it to our needs.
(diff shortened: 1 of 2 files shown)


src/formio/components/TextField.js
```diff
@@ -30,22 +32,41 @@ class TextField extends Formio.Components.components.textfield {
...
 
   handleSettingLocationData(data) {
     const isValidHouseNumber = HOUSE_NUMBER_REGEX.test(data[this.component.deriveHouseNumber]);
     const isValidPostcode = POSTCODE_REGEX.test(data[this.component.derivePostcode]);
 
+    // Fill data if it is not set yet or if the field is readonly (i.e. Formio's disabled).
+    // Unrelated to the HTML 'disabled' attribute.
+    // See #1717 and #1832 - we only want to make API calls if there is useful work to
+    // be done.
+    const mayAutofillValue = !this.getValue() || this.component.disabled;
```
SilviaAmAm, line 65: Does this change the streetname/city if the user entered a house number/postcode and then changes it to another one?

src/formio/components/TextField.js
```diff
@@ -30,22 +32,41 @@ class TextField extends Formio.Components.components.textfield {
...
+  get setLocationData() {
+    if (!this._debouncedSetLocationData) {
+      this._debouncedSetLocationData = debounce((postcode, house_number, key) => {
+        get(`${this.options.baseUrl}location/get-street-name-and-city`, {postcode, house_number})
+          .then(result => {
+            if (result[key]) {
+              this.setValue(result[key]);
+            } else {
+              this.setValue('');
+            }
```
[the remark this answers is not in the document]
sergei-maertens: yes indeed, we need to handle the empty backend response situation :)

src/formio/components/TextField.js
sergei-maertens: it doesn't - that's the biggest problem with this patch (and #1717). If they clear the field completely, then it will be autofilled again.

I'm thinking that we could probably streamline this in the future (read: another PR) where we keep track if the value was set automatically by autofill, or if it was set by the end-user (manually). In the first case, we could then update again even though there's a value set already, in the second case we leave it untouched.

PR #222, opened 2022-08-24
Re-add non validated data to step data

SilviaAmAm:
Fixes open-formulieren/open-forms#1868

### DIFF
--- a/src/components/FormStep.js
+++ b/src/components/FormStep.js
@@ -98,13 +98,15 @@ const doLogicCheck = async (stepUrl, data, invalidKeys=[], signal) => {
   const url = `${stepUrl}/_check_logic`;
   // filter out the invalid keys so we only send valid (client-side) input data to the
   // backend to evaluate logic.
-  if (invalidKeys.length) {
-    data = omit(data, invalidKeys);
-  }
-  const stepDetailData = await post(url, {data}, signal);
+  let dataForLogicCheck = invalidKeys.length ? omit(data, invalidKeys) : data;
+  const stepDetailData = await post(url, {data: dataForLogicCheck}, signal);
   if (!stepDetailData.ok) {
     throw new Error('Invalid response'); // TODO -> proper error & use ErrorBoundary
   }
+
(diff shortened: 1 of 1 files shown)


src/components/FormStep.js
```diff
@@ -98,14 +98,20 @@ const doLogicCheck = async (stepUrl, data, invalidKeys=[], signal) => {
...
   if (!stepDetailData.ok) {
     throw new Error('Invalid response'); // TODO -> proper error & use ErrorBoundary
   }
-  return stepDetailData.data;
+
+  // Re-add any invalid data to the step data that was not sent for the logic check. Otherwise, any previously saved
+  // data in the step will overwrite the user input
+  let {submission, step} = stepDetailData.data;
+  step.data = {...step.data, ...data};
+  return {submission, step};
```
sergei-maertens, line 114: can't we do:

```suggestion
  Object.assign(stepDetailData.data.step.data, data);
  return stepDetailData.data;
```

src/components/FormStep.js
```diff
@@ -98,14 +98,20 @@ const doLogicCheck = async (stepUrl, data, invalidKeys=[], signal) => {
   const url = `${stepUrl}/_check_logic`;
   // filter out the invalid keys so we only send valid (client-side) input data to the
   // backend to evaluate logic.
+  let dataForLogicCheck = cloneDeep(data);
   if (invalidKeys.length) {
-    data = omit(data, invalidKeys);
+    dataForLogicCheck = omit(dataForLogicCheck, invalidKeys);
   }
```
sergei-maertens, line 104: as far as I can see from the docs, `_.omit` should not mutate the first argument, so I think we can avoid the `cloneDeep`?

```js
const dataForLogicCheck = invalidKeys.length ? omit(data, invalidKeys) : data;
```

sergei-maertens (changes requested):

sergei-maertens (approved):

sergei-maertens:
Backported in b7b73d09464bdb296cc09b04bd47729fdfcdf1f7 and 0e11bd5e14e1adb371d733c12bcb7d3e2d3500c5

PR #227, opened 2022-09-06
docs: add Form.io components to Storybook

Robbert:
Add the first batch of Form.io components to Storybook, so we can better track how integrating NL Design System will affect the rendering.


package.json
```diff
@@ -28,6 +28,7 @@
   "license": "EUPL-1.2",
   "dependencies": {
     "@formio/protected-eval": "^1.2.1",
+    "@formio/react": "^5.2.4-rc.1",
```
sergei-maertens, line 31: The SDK currently uses `react-formio` ^4.3 and `formio` 4.12 - I'd stick to those to get an accurate representation. As far as I know, `@react/formio` is the future, but pulls in a newer version of Formio that has known issues with the SDK.

src/formio/components/Button.stories.mdx
```diff
@@ -0,0 +1,38 @@
...
+      label,
+}}/>);
+
+<Canvas>
+  <Story name="Default" args={{ id: 'submit', label: 'Verzenden'}}>
+    {Template.bind({})}
+  </Story>
+</Canvas>
+
+<Props story="Default" />
```
sergei-maertens, line 38: I think we should remove this as it's misleading - we don't use the formio button component(s) (they are not available in the form designer in the backend).

src/formio/components/story-util.js
```diff
@@ -0,0 +1,16 @@
...
+
+export const FormioComponent = ({ component }) => (
+  <Form form={{
+    type: 'form',
+    components: [component]
+  }}/>);
+
+export const FormDecorator = (Story) => (
+  <div
+    className="utrecht-theme utrecht-document"
```
sergei-maertens, line 13: For the future - what would be _really_ cool is a storybook addon that lets you select the theme in a dropdown so we can toggle between them right in the canvas/docs.

sergei-maertens (changes requested):
Thanks for this, that will be super useful!

src/formio/components/story-util.js
Robbert: Awesome, I was hoping we could to do that too. Would you be okay with having `@utrecht/design-tokens` and `@gemeente-denhaag/design-tokens-components` in `devDependencies`?

src/formio/components/Button.stories.mdx
Robbert: So I guess `src/formio/components/templates/button.ejs` is not used at all. Should we even keep it then? Having it in the codebase is also a bit misleading perhaps.

src/formio/components/story-util.js
sergei-maertens: go ahead!

src/formio/components/Button.stories.mdx
sergei-maertens: I *could* be wrong - it's possible it is used for `multiple: true` components to add another entry and possibly also in the `datagrid` component.

@SilviaAmAm can you confirm this?

sergei-maertens: @Robbert we looked into it, and the buttons for data grid and `multiple` components are hard `<button>` elements in their templates (in `src/formio/templates`, the files `multiValueRow.ejs` and `multiValueTable.ejs`).

So I think we can remove the `Button` component. Let's exclude it from storybook in this PR, and then make a separate PR/commit to remove the component - if we're wrong, then it's easily reverted.

Robbert:
@sergei-maertens I have addressed your concerns, and included the theme switcher 😃

sergei-maertens:
> @sergei-maertens I have addressed your concerns, and included the theme switcher smiley

That's really cool! LGTM :tada:

sergei-maertens (approved):

PR #230, opened 2022-09-08
Handle maintenance mode forms & deactivated forms

sergei-maertens:
Fixes open-formulieren/open-forms#1967

* Form designers get a warning but can continue (maintenance mode)
* Everyone gets an error message for deactivated forms

### DIFF
--- a/src/components/FormStart/index.js
+++ b/src/components/FormStart/index.js
@@ -13,6 +13,7 @@ import MaintenanceMode from 'components/MaintenanceMode';
 import { Toolbar, ToolbarList } from 'components/Toolbar';
 import Types from 'types';
 import LoginButton, {LoginButtonIcon} from 'components/LoginButton';
+import {UnprocessableEntity} from 'errors';
 import {getBEMClassName} from 'utils';
 import useStartSubmission from 'hooks/useStartSubmission';
 
@@ -78,6 +79,10 @@ const FormStart = ({ form, onFormStart }) => {
         <Loader modifiers={['centered', 'only-child']} />
       </Card>
     );
+  }
+
+  if (!form.active) {
+    throw new UnprocessableEntity('Unprocessable Entity', 422, 'Form not active', 'form-inactive');
   }
 
   if (form.maintenanceMode) {
--- a/src/components/FormStart/index.js
+++ b/src/components/FormStart/index.js
@@ -2,7 +2,7 @@ import React, {useEffect, useRef, useState} from 'react';
 import PropTypes from 'prop-types';
 import {FormattedMessage} from 'react-intl';
 
-import AuthenticationOutage, { useDetectAuthenticationOutage } from 'components/auth/AuthenticationOutage';
+import AuthenticationOutage, {useDetectAuthenticationOutage} from 'components/auth/AuthenticationOutage';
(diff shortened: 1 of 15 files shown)


src/components/RequireSubmission.js
```diff
@@ -10,11 +13,23 @@ import { Redirect } from 'react-router-dom';
...
+
   if (!submission || !Object.keys(submission).length) {
     return (<Redirect to="/" />);
   }
   return (
-    <Component submission={submission} {...props} />
+    <>
+      {maintenanceModeAlert}
+      <Component submission={submission} {...props} />
+    </>
```
sergei-maertens, line 32: this is not ideal, I'd rather have it under the title, but that messes with the structure/re-usability of components so I took the path of least resistance.

src/components/RequireSubmission.js
```diff
@@ -10,11 +13,23 @@ import { Redirect } from 'react-router-dom';
...
 const RequireSubmission = ({ submission, component: Component, ...props }) => {
+  const maintenanceMode = props?.form?.maintenanceMode;
+  const isFormDesigner = IsFormDesignerHeader.getValue();
+  let maintenanceModeAlert = null;
+  if (!isFormDesigner && maintenanceMode) {
+    throw new ServiceUnavailable('Service Unavailable', 503, 'Form in maintenance', 'form-maintenance');
+  } else if (isFormDesigner && maintenanceMode) {
+    maintenanceModeAlert = <MaintenanceMode />;
+  }
+
```
SilviaAmAm, line 24: Wouldn't it make more sense to do this after the check for whether there is a submission?

SilviaAmAm (changes requested):

src/components/FormStart/index.js
```diff
@@ -101,6 +108,8 @@ const FormStart = ({ form, onFormStart }) => {
     <LiteralsProvider literals={form.literals}>
       <Card title={form.name}>
 
+        {isFormDesigner && <MaintenanceMode asToast />}
```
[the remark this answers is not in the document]
sergei-maertens: I think you're right, I'll verify it!

src/components/FormStep.js
```diff
@@ -66,12 +65,19 @@ const submitStepData = async (stepUrl, data) => {
...
+    try {
+      validateResponse = await post(validateUrl, data);
+    } catch(error) {
+      onBackendError(error);
+      next([{path: '', message: error.detail, code: error.code}]);
+      return;
+    }
 
     // process the errors
     if (validateResponse.status === 400) {
```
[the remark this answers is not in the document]
sergei-maertens: you're right again - I cleaned up some other unreachable code that similarly threw an error. I'll have to do type checking in the `catch` block and handle it appropriately.

Will try it again and fix it :-)

src/components/RequireSubmission.js
sergei-maertens: It was originally like that, but then the page rendered the form step briefly (half a second or so), to then throw and go into the error boundary, which was a really janky user experience.

In terms of UX, if you can detect an error situation before the user even does any interaction (explicit or implicit), you want to handle it asap.

PR #238, opened 2022-09-22
[OF#2058] Summary page shouldnt display fields inside hidden fieldsets

SilviaAmAm:
Fixes open-formulieren/open-forms#2058


src/components/FormStepSummary/utils.js
```diff
@@ -3,7 +3,7 @@ import {Utils as FormioUtils} from 'formiojs';
 
 const iterComponentsWithData = (components, data) => {
```
sergei-maertens, line 4: I think this can be deleted

sergei-maertens (approved):

PR #241, opened 2022-09-26
Remove accidental styling of content components

SilviaAmAm:
Fixes open-formulieren/open-forms#2073

Fixes open-formulieren/open-forms#2067

### DIFF
--- a/src/scss/components/_form-control.scss
+++ b/src/scss/components/_form-control.scss
@@ -16,28 +16,28 @@
     display: none;
   }
 
-  .error {
+  &.error {
     @include border(left, var(--of-color-danger), 4px);
     padding-left: $grid-margin-3;
   }
 
-  .warning {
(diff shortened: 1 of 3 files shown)


SilviaAmAm:
Noticed that this shows a red rectangle by (valid) repeating groups. Need to fix.

src/scss/components/_form-control.scss
```diff
@@ -16,26 +16,33 @@
     display: none;
   }
 
-  &[class*="error"] {
+  .error {
```
sergei-maertens, line 19: This results in the CSS rule `.openforms-form-control .error` rather than `.openforms-form-control.error` - is that intentional?

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #244, opened 2022-10-03
[OF#2035] Scroll to top on step load

SilviaAmAm:
Fixes open-formulieren/open-forms#2035

### DIFF
--- a/src/components/FormStep.js
+++ b/src/components/FormStep.js
@@ -247,6 +247,7 @@ const FormStep = ({
         payload: stepDetail,
       });
       configurationRef.current = stepDetail.formStep.configuration;
+      window.scrollTo(0, 0, {behavior: 'smooth'})
     },
     [submissionStep.url]
   );


src/components/FormStep.js
```diff
@@ -135,6 +135,8 @@ const reducer = (draft, action) => {
       draft.canSubmit = canSubmit;
       draft.logicChecking = false;
       draft.isNavigating = false;
+      
+      window.scrollTo({ top: 0, behavior: 'smooth' });
```
sergei-maertens, line 139: you should _never_ apply side-effects in state-management code - it's possible that these state updates run multiple times before React even re-renders.

Side-effects like this belong in the `useEffect` hook during rendering, and we have a custom hook for this even: https://github.com/open-formulieren/open-forms-sdk/blob/main/src/hooks/useScrollIntoView.js

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #251, opened 2022-10-05
ci: visual regression test with Chromatic

Robbert:
Perhaps we can use our Frameless account to do visual regression tests for a while, that seems like a better approach than @sergei-maertens spending his precious time doing visual regression tests for our changes 😅

I suspect I need to be a contributor or maintainer of this repo to be able to set this up from Chromatic, otherwise the repo won't show up in my list of options. Then someone also needs to configure the project token. Let's discuss this idea later this week.

### DIFF
--- a/.github/workflows/sb.yml
+++ b/.github/workflows/sb.yml
@@ -10,12 +10,6 @@ on:
       - main
   workflow_dispatch:
 
-# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
-permissions:
-  contents: read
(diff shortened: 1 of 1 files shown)


sergei-maertens (commented):
Cool, this looks useful! I assume we'll discuss this a bit tomorrow, and then I can take it to Joeri as well :)

sergei-maertens:
Merged via 2fb1ef3f8da22f078641eb6430a24dbd3827b3f9 (after rebase on main)

PR #254, opened 2022-10-06
Fix style progress indicator

SilviaAmAm:
Fixes https://github.com/open-formulieren/open-forms/issues/2127

Screenshot from 2022-10-06 17-19-11

Screenshot from 2022-10-06 17-21-49


src/components/ProgressIndicator/index.js
```diff
@@ -26,24 +27,31 @@ const getLinkModifiers = (active, isApplicable) => {
...
 
-const LinkOrDisabledAnchor = ({ to, useLink, children, ...props }) => {
+const LinkOrSpan = ({isActive, isApplicable, to, useLink, children, ...props}) => {
   if (useLink) {
-    return <Link to={to} component={Anchor} {...props}>{children}</Link>
+    return <Link to={to} component={Anchor} modifiers={ getLinkModifiers(isActive, isApplicable) } {...props}>{children}</Link>
   }
-  return <Anchor component="span" {...props}>{children}</Anchor>
+
+  return <Span modifiers={['muted']} {...props}>{children}</Span>
```
sergei-maertens, line 35: I don't think we need an entire new `Span` component for this, it looks to me that the `Body` component should be sufficient:

```jsx
<Body component="span" modifiers={['muted']}>{children}</Body>
```
maybe some extra modifier will be needed to remove padding/margin, but I'd rather make the `Body` component more flexible than adding stuff that's only used in a single place.

Another option would be to use an actual anchor that's disabled via CSS:

```css
.anchor.anchor--disabled {
    pointer-events: none;
}
```

but I think that doesn't block tab-events :thinking: However, the `onClick` in React may possible fix that?

sergei-maertens (changes requested):

src/components/ProgressIndicator/index.js
SilviaAmAm: The body component works nicely!

sergei-maertens (approved):

PR #265, opened 2022-10-13
feat: add NL Design System class names to components

Robbert:
This PR replaces #228, but a branch the main repo to have access to visual regression tests. Will still address comments from other PR of course.


src/components/nl-design-system-community.scss
```diff
@@ -0,0 +1,13 @@
+@import '@utrecht/components/dist/button/css/index.css';
```
sergei-maertens, line 1: I think this file rather belongs in `src/scss` than `src/components` for consistency sake

src/components/Button.js
```diff
@@ -36,16 +36,60 @@ const Button = ({ type='', component: Component='button', variant='', onDisabled
...
+          }
+          case 'button': {
+            // LinkButton link uses <button> element
+            Component = UtrechtLinkButton;
+            break;
+          }
+          default: {
+            // Render `UtrechtButton` or custom component
+            break;
+          }
```
sergei-maertens, line 68: this *might* have to default to `case 'button'`, but for now I've worked around it

sergei-maertens (approved):

PR #268, opened 2022-10-14
[OF#2126] Make crosses reachable via tab navigation

SilviaAmAm:
Fixes open-formulieren/open-forms#2126


src/formio/templates/file.ejs
```diff
@@ -17,7 +17,11 @@
       <li class="list-group-item">
         <div class="row">
           {% if (!ctx.disabled) { %}
-            <div class="col-md-1"><i class="{{ctx.iconClass('remove')}}" ref="removeLink"></i></div>
+            <div class="col-md-1">
+              <button type="button" class="{{ctx.ofPrefix}}button {{ctx.ofPrefix}}button--invisible" ref="removeLink">
+                <i class="{{ctx.iconClass('remove')}}"></i>
+              </button>
+            </div>
```
sergei-maertens, line 24: Can't we make this an `a` tag instead? Semantically that makes more sense and we then wouldn't need an additional variant of the button.

sergei-maertens (changes requested):

src/formio/templates/file.ejs
SilviaAmAm: We could. I just followed what Formio does in their template (the removal icons are inside buttons)

SilviaAmAm: After discussion: 
- Looked at both the [button that looks like a link with icon](https://nl-design-system.github.io/utrecht/storybook/?path=/docs/css-link-button--with-icon) and the [button with icon](https://nl-design-system.github.io/utrecht/storybook/?path=/docs/css-button--icon-after) but they don't really seem like what we need here?

sergei-maertens (changes requested):
I don't think the `invisible` variant name is a good one - I'd expect that nothing is displayed then. Probably `icon-only` as variant name is more appropriate.

I also think we should remove the `color: $color-danger` styles from the file-upload/multi-value-row components, and instead apply both button modifiers: `button button--icon-only button--danger` and style appropriately in the button component. We then should not need the `!important` overrides _and_ we can keep using design tokens to control the colors in a consistent manner.

SilviaAmAm:
I notice that while in the SDK the buttons don't change when you hover them, they do change in Storybook :thinking:

src/scss/components/_button.scss
```diff
@@ -169,4 +169,22 @@ $button-padding-h: $grid-margin-2;
...
+
+    &:hover, &:active {
+      --utrecht-button-hover-color: var(--of-button-fg);
+      color: var(--of-button-fg);
+      background-color: transparent;
+      border-style: none;
+    }
+
+    &.#{prefix(button)}--danger {
+      color: var(--of-button-danger-bg)
```
sergei-maertens, line 187:
```suggestion
      color: var(--of-button-danger-bg)
```

PR #269, opened 2022-10-17
[OF#1180] Fix gtag calls

SilviaAmAm:
Fixes open-formulieren/open-forms#1180

### DIFF
--- a/src/hooks/usePageViews.js
+++ b/src/hooks/usePageViews.js
@@ -6,7 +6,12 @@ const isDev = process.env.NODE_ENV === 'development';
 
 const ANALYTICS_PROVIDERS = {
   debug: async (location) => (isDev && console.log(`Tracking navigation to ${location.pathname}`)),
-  googleAnalytics: async (location) => (window.ga && window.ga.send(['pageview', location.pathname])),
+  gtag: async (location) => {
+    return window.gtag && window.gtag('event', 'page_view', {
+      page_location: location.href,
+      page_path: location.pathname,
+    });
+  },
   // https://support.siteimprove.com/hc/en-gb/articles/115001615171-Siteimprove-Analytics-Custom-Visit-Tracking
   // unsure if SiteImprove subscribes to URL changes or not, if they do - we'll fire double events here.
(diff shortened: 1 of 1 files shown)


sergei-maertens (changes requested):
As discussed in the office:

* there's probably a mistake in the `usePageviews` hook for GA causing it to not send pageviews
* we should be a bit more conservative with the window title - on load/SDK init, read the page title as is, and then append our own bit to it, otherwise we break the page titles for embedded forms.

SilviaAmAm:
Things I noticed:
- windows.ga (from the hook) sends data to www.google-analytics.com. However, the gtag documentation suggests we should use windows.gtag (https://developers.google.com/analytics/devguides/collection/gtagjs/sending-data#send_data_with_the_event_command)
- To send the pathname, an extra parameter has to be added to the script in the `google.html` template: `gtag('config', '{{ analytics_tools_config.ga_code|escapejs }}', {'page_path': location.pathname});`. We can also add 'page_title' here (https://developers.google.com/analytics/devguides/collection/gtagjs/pages#default_behavior).

src/hooks/usePageViews.js
```diff
@@ -6,7 +6,12 @@ const isDev = process.env.NODE_ENV === 'development';
 
 const ANALYTICS_PROVIDERS = {
   debug: async (location) => (isDev && console.log(`Tracking navigation to ${location.pathname}`)),
-  googleAnalytics: async (location) => (window.ga && window.ga.send(['pageview', location.pathname])),
+  gtag: async (location) => {
+    return window.gtag && window.gtag('event', 'page_view', {
+      page_location: location.href,
+      page_path: location.pathname,
+    });
```
SilviaAmAm, line 13: What I understood from the [documentation](https://developers.google.com/analytics/devguides/collection/gtagjs/pages#manual_pageviews) is that the default behaviour would be to have the snippet on each page where a view needs to be tracked. 

To have manual events sent to the `/collect` endpoint, we have to deactivate the call made by the snippet (`gtag('config', '{{ analytics_tools_config.ga_code|escapejs }}', {send_page_view: false});`) and then add the code in this diff to manually track the page_view event.

PR #270, opened 2022-10-24
:art: emit own styles/design tokens in openforms-theme scope

sergei-maertens:
Relates to open-formulieren/open-forms#2225

To avoid overriding design tokens from *other* themes, include our own scoped design token values only if the openforms theme is set.


SilviaAmAm (approved):

Robbert (approved):

sergei-maertens:
Chromatic UI isn't entirely happy yet, so I'll check that before merging :grimacing:

PR #272, opened 2022-10-31
:truck: [#2232] Rename i18nEnabled -> translationEnabled

stevenbal:
since the djangorestframework-camel-case parser does not seem to parse i18nEnabled properly (it results in i_18n_enabled)

### DIFF
--- a/src/sdk.js
+++ b/src/sdk.js
@@ -12,7 +12,7 @@ import OFLibrary from './formio/templates';
 
 import './styles.scss';
 
-import { get } from 'api';
+import { get, post } from 'api';
 import { ConfigContext, FormioTranslations } from 'Context';
 import App from 'components/App';
 import {CSPNonce} from 'headers';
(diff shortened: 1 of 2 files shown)


stevenbal:
@sergei-maertens it seems that the camelcase parser for DRF parses `i18nEnabled` to `i_18n_enabled` and I didn't want to deal with that, so I renamed the field

src/sdk.js
```diff
@@ -99,6 +99,11 @@ class OpenForm {
 
     this.formObject = formObject;
 
+    // Explicitly set the default language if translations are disabled
+    if(!formObject.translationEnabled) {
+      post(`${formObject.url}/set_default_language`);
```
sergei-maertens, line 104:
```suggestion
      await post(`${formObject.url}/set_default_language`);
```

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #273, opened 2022-11-03
[#2253] Language Selection Component

CharString:
Closes open-formulieren/open-forms#2253


src/components/LanguageSelection/LanguageSelection.js
```diff
@@ -0,0 +1,100 @@
...
+
+  // unpack props
+  const { heading, headingLevel, onLanguageChange } = props;
+  const headingId = uuidv4(); // XXX: useId() from 'react'
+
+  const changeLanguage = (lang_code) => async () => {
+    if (lang_code === current) return;
+    await put(`${config.baseUrl}i18n/language`, { code: lang_code });
+    setCurrent(lang_code); // state change
+    onLanguageChange(lang_code);
```
CharString, line 59: Even after reading https://github.com/formatjs/formatjs/issues/243 it's not crystal clear to me how to properly trigger a locale change in the containing `IntlProvider` of the storybook.

src/components/LanguageSelection/LanguageSelection.stories.mdx
```diff
@@ -0,0 +1,135 @@
...
+    }}
+    play={async ({args, canvasElement}) => {
+    const canvas = within(canvasElement);
+    const frysk_button = await waitFor(() => canvas.findByText(/^fy$/i)); // wait for api info call to return
+    await userEvent.click(frysk_button);
+    await userEvent.click(frysk_button); // click twice
+    await waitFor(() => expect(args.onLanguageChange).toHaveBeenCalledTimes(0)); // change once
+    }}
+  >{Template.bind({})}</Story>
+</Canvas>
```
CharString, line 135: It would be nice if we could render the Mock panel in the docs tab.
Also the show code block looks particularly ugly.

yarn.lock
```diff
@@ -5134,29 +5210,29 @@
     eslint-visitor-keys "^3.0.0"
 
 "@utrecht/component-library-css@^1.0.0-alpha.343":
-  version "1.0.0-alpha.343"
-  resolved "https://registry.yarnpkg.com/@utrecht/component-library-css/-/component-library-css-1.0.0-alpha.343.tgz#87fdb158b63804dc6f0f1e03875fc0bb6531c9db"
-  integrity sha512-a2+b5ewCN/kbVQYHuSORB3vYwaa6o0Oj8zNd17z+9IJ9tQE2VTOeElaVXBQ6Y1EnkT+8i3rTVn3H2ckHifceaw==
+  version "1.0.0-alpha.365"
```
CharString, line 5213: image
This is what it looks like in the [Utrecht storybook](https://nl-design-system.github.io/utrecht/storybook/?path=/docs/css-alternate-lang-nav--button-group). I think for visual a11y, there should be a visual separator between the buttons.

Robbert: I think you could use the following, similar to the alternate lang link approach:

```jsx
<Button inline>EN</Button><span aria-hidden="true">{' | '}</span><Button inline>NL</Button>
```

src/components/LanguageSelection/LanguageSelection.js
```diff
@@ -0,0 +1,104 @@
+import React, { useContext, useState } from "react";
+import { ConfigContext } from "Context";
+import { useAsync } from "react-use";
+import Loader from "components/Loader";
+import { FormattedMessage, useIntl } from "react-intl";
+import { v4 as uuidv4 } from "uuid";
+import {
+  ButtonGroup,
+  Heading,
+  LinkButton,
```
sergei-maertens, line 10: For the time being we should use our own `Button` component here with the appropriate variants:

```jsx
<Button variant="anchor" component="button" ... />
```
should result in the same markup, but have the advantage that our own styling layer is applied on top.

src/components/LanguageSelection/LanguageSelection.js
```diff
@@ -0,0 +1,104 @@
...
+} from "@utrecht/component-library-react";
+import { get, put } from "../../api";
```
sergei-maertens, line 12:
```suggestion
import { get, put } from "api";
```
we use "absolute" imports (relative to the `src` directory)

src/components/LanguageSelection/LanguageSelection.js
```diff
@@ -0,0 +1,104 @@
...
+import PropTypes from "prop-types";
```
sergei-maertens, line 13: unfortunately we don't have an automated tool for this (yet), but please group your imports similar to the backend:

1. Third party packages (from `node_modules`)
2. blank line
3. Own package (absolute imports from the `src` folder)
4. blank line
5. Relative imports within the same (sub) package (e.g. `./utils`)

src/components/LanguageSelection/LanguageSelection.js
```diff
@@ -0,0 +1,104 @@
...
+
+const propTypes = PropTypes.shape({
+  heading: PropTypes.string,
+  headingLevel: PropTypes.number,
+  onLanguageChange: PropTypes.func,
+});
```
sergei-maertens, line 19: preferable we define those directly on the component after the component definition:

```jsx
LanguageSelection.propTypes = ...
```

src/components/LanguageSelection/LanguageSelection.js
```diff
@@ -0,0 +1,104 @@
...
+import {
+  ButtonGroup,
+  Heading,
+  LinkButton,
+} from "@utrecht/component-library-react";
+import { get, put } from "../../api";
+import PropTypes from "prop-types";
+
+const propTypes = PropTypes.shape({
+  heading: PropTypes.string,
```
sergei-maertens, line 16:
```suggestion
  heading: PropTypes.node,
```
React elements aren't strings, but they can be rendered, so this is a more appropriate prop type (see the default prop)

src/components/LanguageSelection/LanguageSelection.js
```diff
@@ -0,0 +1,104 @@
...
+  headingLevel: PropTypes.number,
+  onLanguageChange: PropTypes.func,
+});
+
+const LanguageSelection = (props) => {
```
sergei-maertens, line 21: destructuring the props in the callback allows you to provide default values for optional props, so you don't need to specify `defaultProps`:

```jsx
const LanguageSelection = ({
  heading=<FormattedMessage description="Language selection heading" defaultMessage="Choose language" />,
  headingLevel=2,
  onLanguageChange=console.log,
}) => {
```

the `onLanguageChange` prop will probably be replaced with something we get from the config context instead, so I wouldn't worry about that one and/or the default value too much now.

src/components/LanguageSelection/LanguageSelection.js
```diff
@@ -0,0 +1,104 @@
...
+  const getAvailableLanguages = async () => {
+    const result = await get(`${config.baseUrl}i18n/info`);
+    setCurrent(result.current); // state change if current changed
+    return result.languages.map(({ code, name }) => ({
+      lang: code,
+      textContent: code.toUpperCase(),
+      label: name,
+      current: code === result.current,
+    }));
+  };
```
sergei-maertens, line 35: We won't need local state once the callback cycle is implemented, as that will lead to a re-render and different value for `locale` from the `useIntl` hook.

Then when `result.current != locale`, we'll call the `onLanguageChanged` callback which informs the root component state of the active language and everything will render accordingly.

src/sdk.js
```diff
@@ -60,6 +59,22 @@ class OpenForm {
...
+      }
+      case "object": {
+        this.languageSelectorTarget = languageSelectorTarget;
+        break;
+      }
+      default:
+        this.languageSelectorTarget = undefined;
+        break;
+    }
+
```
CharString, line 77: Where should tests for this live? I've tested it manually by toying around with `public/sdk.html` but that didn't feel quite right.

src/components/LanguageSelection/LanguageSelection.js
```diff
@@ -0,0 +1,104 @@
...
+  // Hook uses
+  const config = useContext(ConfigContext);
+  const { locale } = useIntl();
+  const [current, setCurrent] = useState(locale);
+  const getAvailableLanguages = async () => {
+    const result = await get(`${config.baseUrl}i18n/info`);
+    setCurrent(result.current); // state change if current changed
+    return result.languages.map(({ code, name }) => ({
+      lang: code,
+      textContent: code.toUpperCase(),
```
sergei-maertens, line 31: probably better to keep the actual text content lower case & do the transform in CSS? @Robbert - do you have any input here?

src/components/LanguageSelection/LanguageSelection.js
CharString: This actually belongs in open-formulieren/open-forms#2256

src/components/LanguageSelection/LanguageSelection.js
```diff
@@ -0,0 +1,104 @@
...
+      current: code === result.current,
+    }));
+  };
+  const {
+    loading,
+    value: items,
+    error,
+  } = useAsync(
+    getAvailableLanguages,
+    [config.baseUrl] // actually current too?!
```
sergei-maertens, line 42: yeah, but it should rather be `locale` then and that's what we should compare against.

We will do the GET all again on language changes because that will change the API response `current`, but if that wouldn't be used in comparisons, we could split this in two hooks so that the GET call would only be called once :thinking: 

This is okay for now though.

src/components/LanguageSelection/LanguageSelection.js
```diff
@@ -0,0 +1,104 @@
...
+  const [current, setCurrent] = useState(locale);
+  const getAvailableLanguages = async () => {
+    const result = await get(`${config.baseUrl}i18n/info`);
+    setCurrent(result.current); // state change if current changed
+    return result.languages.map(({ code, name }) => ({
+      lang: code,
+      textContent: code.toUpperCase(),
+      label: name,
+      current: code === result.current,
+    }));
```
sergei-maertens, line 34: This transform should be moved outside of the hook - it's side-effect free and  it allows you to use other hook return values/state (for the `current` bool) without having to pass that to the hook dependencies. Eventually we'll then be able to avoid hitting the API when it's not absolutely needed.

src/components/LanguageSelection/LanguageSelection.js
```diff
@@ -0,0 +1,100 @@
...
+  if (error) {
+    throw error; // bubble up to boundary
+  }
+  if (loading) {
+    return <Loader modifiers={["small"]} />;
+  }
+
+  // unpack props
+  const { heading, headingLevel, onLanguageChange } = props;
+  const headingId = uuidv4(); // XXX: useId() from 'react'
```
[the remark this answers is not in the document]
sergei-maertens: Cool that they have a thing built in for that - at the moment we're not React 18 compatible yet so the TODO is fine.

I don't see HTML IDs in the Utrecht storybook, but if it's a required prop I suppose we can just use a static value too: `'of-langselection-title'` seems fine.

edit: I see it's used for accessibility reasons :)

src/components/LanguageSelection/LanguageSelection.js
```diff
@@ -0,0 +1,104 @@
...
+
+  const changeLanguage = (lang_code) => async () => {
+    if (lang_code === current) return;
```
sergei-maertens, line 56: should compare against `locale` when we remove local state

src/components/LanguageSelection/LanguageSelection.js
```diff
@@ -0,0 +1,104 @@
...
+
+  return (
+    <section className="utrecht-alternate-lang-nav" aria-labelledby={headingId}>
+      <Heading
+        level={headingLevel}
+        className="utrecht-alternate-lang-nav__heading"
+        id={headingId}
+      >
+        {heading}
+      </Heading>
```
sergei-maertens, line 70: we'll probably want to only have this visible for screen readers - I can see in the Utrecht SB that they have a CSS rule:

```css
.utrecht-alternate-lang-nav__heading {
    display: none;
}
```
self-reminder: figure out why the CSS is not applied

src/components/LanguageSelection/LanguageSelection.js
```diff
@@ -0,0 +1,104 @@
...
+  }
+  if (loading) {
+    return <Loader modifiers={["small"]} />;
+  }
+
+  // unpack props
+  const { heading, headingLevel, onLanguageChange } = props;
+  const headingId = uuidv4(); // XXX: useId() from 'react'
+
+  const changeLanguage = (lang_code) => async () => {
```
sergei-maertens, line 55:
```suggestion
  const onChangeLanguage = (lang_code) => async () => {
      event.preventDefault();
```
1. event callbacks should be prefixed with `on` (convention)

I initially also had the `event` argument added to prevent accidental navigations/submissions, but that shouldn't be a problem when you use `<button type="button">` elements.

src/components/LanguageSelection/LanguageSelection.js
```diff
@@ -0,0 +1,104 @@
...
+      </Heading>
+      <ButtonGroup>
+        {items.map(({ current, label, lang, textContent }, i, a) => (
+          <>
+            <LinkButton
+              pressed={current}
+              lang={lang}
+              aria-label={label}
+              key={lang}
+              onClick={changeLanguage(lang)}
```
sergei-maertens, line 79: I'd tweak the function signature a bit so this is more in line with the conventional approach to bind event handlers:

```jsx
const onChangeLanguage = async (languageCode) => { ... };
```

(p.s. - in JS we typically use camelCase)

and then use a closure:

```suggestion
              onClick={() => onChangeLanguage(lang)}
```

src/components/LanguageSelection/LanguageSelection.js
```diff
@@ -0,0 +1,104 @@
...
+              inline
+            >
+              {textContent}
+            </LinkButton>
+            {i + 1 < a.length ? <span aria-hidden="true">{" | "}</span> : <></>}
```
sergei-maertens, line 84:
```suggestion
            {i + 1 < a.length ? <span aria-hidden="true">{" | "}</span> : null}
```

src/components/LanguageSelection/LanguageSelection.js
```diff
@@ -0,0 +1,104 @@
...
+    <FormattedMessage
+      description="Language selection heading"
+      defaultMessage="Choose language"
+    />
+  ),
+  headingLevel: 2,
+  onLanguageChange: (new_language_code) => {},
+};
+
+export default LanguageSelection;
```
sergei-maertens, line 104: I'd like to see this broken out still into a dumb and smart component:

1. the smart component has all the hooks stuff happening
2. the dumb component is basically just rendering from props:

```jsx
const LanguageSelectionDisplay = ({ 
  heading,
  headingLevel,
  headingId, 
  items, 
  onChangeLanguage,
}) => (
  <section className="utrecht-alternate-lang-nav" aria-labelledby={headingId}>
    <Heading
      level={headingLevel}
      className="utrecht-alternate-lang-nav__heading"
      id={headingId}
    >
      {heading}
    </Heading>
    <ButtonGroup>
      {items.map(({ current, label, lang, textContent }, i, a) => (
        <>
          <LinkButton
            pressed={current}
            lang={lang}
            aria-label={label}
            key={lang}
            onClick={() => onChangeLanguage(lang)}
            inline
          >
            {textContent}
          </LinkButton>
          {i + 1 < a.length ? <span aria-hidden="true">{" | "}</span> : null}
        </>
      ))}
    </ButtonGroup>
  </section>
);
```

advantage of that is that we can isolate the logic from presentation and then SDK users that use the NPM package can easily swap in another `LanguageSelectionDisplay` implementation.

src/components/LanguageSelection/LanguageSelection.stories.mdx
sergei-maertens: the show code block is probably ugly because of the explicit `ConfigContext.Provider` - that should rather be implemented as a story decorator I think: https://storybook.js.org/docs/react/writing-stories/decorators

src/components/App.js
```diff
@@ -1,35 +1,42 @@
...
-const App = (props) => {
+const App = ({ languageSelectorTarget, ...formProps }) => {
+  const LanguageSwitcher = () =>
+    languageSelectorTarget ? (
+      ReactDOM.createPortal(<LanguageSelection />, languageSelectorTarget)
+    ) : (
+      <LayoutRow>
+        <LanguageSelection />
+      </LayoutRow>
+    );
```
sergei-maertens, line 22: maybe break this out in a separate component (within `App.js`) and pass down the `languageSelectorTarget` prop. This closure isn't necessary.

src/components/App.js
```diff
@@ -1,35 +1,42 @@
...
 import { Layout, LayoutRow } from 'components/Layout';
 import ManageAppointment from 'components/appointments/ManageAppointment';
 
+import LanguageSelection from "components/LanguageSelection";
 
 /*
 Top level router - routing between an actual form or supporting screens.
  */
-const App = (props) => {
+const App = ({ languageSelectorTarget, ...formProps }) => {
```
sergei-maertens, line 14:
```suggestion
const App = ({ languageSelectorTarget, ...props }) => {
```
will also still work - don't think this rename is particularly needed

src/sdk.js
sergei-maertens: `sdk.spec.js` would be appropriate and be picked up - you can find some inspiration in `jstests/formio/components/`

sergei-maertens (changes requested):

src/components/LanguageSelection/LanguageSelection.js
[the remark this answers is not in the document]
CharString: I think by default the first h… element in a section labels the document outline, so it may be unnecessary. I just followed Robbert's example.

src/components/LanguageSelection/LanguageSelection.stories.mdx
CharString: I retried using a decorator, but no real change... There is a difference in what code comes out between using `<Story />` or `{Story()}`. But neither will render useful example code.

.storybook/preview.js
```diff
@@ -34,7 +43,11 @@ export const parameters = {
       { name: 'Gemeente Utrecht', class: 'utrecht-theme', color: '#cc0000' }
     ],
   },
-}
+};
+
+export const decorators = [
+  ConfigDecorator,
+];
```
sergei-maertens, line 50: looks like there is a better option after all via https://github.com/storybookjs/storybook/issues/16999#issuecomment-1127503783, one more thing to try

Applying the decorators globally does render the source code preview properly, while doing it at the Meta/Story level breaks in various ways depending on how you use it.

sergei-maertens (approved):
commit history is a mess, but I'm really done with this PR :stuck_out_tongue:

PR #281, opened 2022-11-14
Load translations onLanguageChanged

CharString:
- Adds I18NManager to cantain all language string loading and signalling re-renders.
 - Moves `onLanguageChanged` callback from `LanguageSelection` params to `ConfigContext`
   
 Closes open-formulieren/open-forms#2255

### DIFF
--- a/src/components/LanguageSelection/LanguageSelection.stories.mdx
+++ b/src/components/LanguageSelection/LanguageSelection.stories.mdx
@@ -1,7 +1,7 @@
 import {useState} from 'react';
 import { Meta, Canvas, Story, ArgsTable } from '@storybook/addon-docs';
 import { within, userEvent, waitFor } from '@storybook/testing-library';
-import { expect, fn } from '@storybook/jest';
+import { expect } from '@storybook/jest';
 import { IntlProvider } from 'react-intl';
 
 import { ConfigContext } from 'Context';
--- a/src/components/LanguageSelection/LanguageSelection.stories.mdx
+++ b/src/components/LanguageSelection/LanguageSelection.stories.mdx
@@ -1,7 +1,7 @@
 import {useState} from 'react';
 import { Meta, Canvas, Story, ArgsTable } from '@storybook/addon-docs';
 import { within, userEvent, waitFor } from '@storybook/testing-library';
-import { expect } from '@storybook/jest';
+import { expect, fn } from '@storybook/jest';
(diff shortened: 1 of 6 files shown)


CharString:
Flatpickr seems to automagically change. I don't know what triggers it, but on every language change it will add a new node to the DOM with a new flatpickr script with the new locale. It even keeps working if I remove this line: 
https://github.com/open-formulieren/open-forms-sdk/blob/5e41548abd4ddae4e236909ae04ae43275d4db0a/src/sdk.js#L94
I tried to find the magic, because it doesn't clean up the previous nodes, but I gave up.

Also tried to get
https://github.com/open-formulieren/open-forms-sdk/blob/5e41548abd4ddae4e236909ae04ae43275d4db0a/src/i18n.js#L14-L21
to use `import()` such that it will load strings dynamically and async to reduce main bundle size. But `try {} catch {}` doesn't catch the import errors that can occur.

src/components/LanguageSelection/LanguageSelection.stories.mdx
```diff
@@ -33,7 +33,6 @@ export const ConfigDecorator = (Story, { args }) => (
         type: { summary: 'number' },
       },
     },
-    onLanguageChanged: { table: { disable: true } },
```
sergei-maertens, line 36: this arg is not relevant for the `LanguageSelectionDisplay` component and you can't change it in the controls anyway, that's why I disabled it for the canvas.

src/i18n.js
```diff
@@ -1,19 +1,99 @@
...
 
 const loadLocaleData = (locale) => {
-    switch (locale) {
-        case 'nl':
-            return messagesNL;
-        default:
-            return messagesEN;
-    }
+  switch (locale) {
+    case "nl":
```
sergei-maertens, line 16: please keep the code style consistent with existing code style in the file - we'll implement prettier for the SDK at some point, but javascript strings are generally single quote, while JSX props use double quotes.

These sort of formatting changes really clutter the diff and make reviewing harder.

sergei-maertens (changes requested):
I'm not a fan of hardcoding the App in this container component, why not use `props.children` instead?

We now have a component called `I18NManager` doing much more than that - it's also rendering the app, loading the form details from the API which is really the responsibility of other components that need to be composed.

I'll see if I can come to a set up that is more loosely coupled

PR #289, opened 2022-11-21
card-titles as page-titles

jiromaykin:
fixes open-formulieren/open-forms#2272


src/components/FormStep.js
```diff
@@ -233,6 +233,11 @@ const FormStep = ({form, submission, onLogicChecked, onStepSubmitted, onLogout})
   // look up the form step via slug so that we can obtain the submission step
   const formStep = form.steps.find(s => s.slug === slug);
   const currentStepIndex = form.steps.indexOf(formStep);
+
+  // use meaningful document titles
+  const pageTitle = `${config.titlePrefix} - ${formStep.formDefinition}`;
+  useTitle(pageTitle);
```
sergei-maertens, line 239: I would wrap this in our own `useTitle` hook to encapsulate the prefix behaviour.

```jsx
// in hooks/useTitle.js

import {useContext} from 'react';
import {useTitle as reactUseTitle} from 'react-use';

const useTitle = (localTitle) => {
    const {titlePrefix} = useContext(ConfigContext);
    return titlePrefix ? `${titlePrefix} / ${localTitle}` : localTitle;
};

export default useTitle;
```

Usage would then be:

```jsx
import useTitle from 'hooks/useTitle';

const MyComponent = (...) => {
  ...
  useTitle(formStep.formDefinition);
  ...
};
```

1. This ensure that the prefix is _always_ applied without having to concatenate the strings manually
2. If no existing document title was given, the prefix is empty and then we don't include the separator

src/components/ProgressIndicator/index.js
```diff
@@ -30,19 +32,30 @@ const ProgressIndicator = ({
   const applicableCompleted =
     hasSubmission && applicableSteps.length === applicableAndCompletedSteps.length;
 
+  // meaningful titles based on steps
+  const config = useContext(ConfigContext);
+  let currentStepTitle = '';
```
sergei-maertens, line 37: see above remark about custom `useTitle` hook to abstract this away

src/components/ProgressIndicator/index.js
```diff
@@ -30,19 +32,30 @@ const ProgressIndicator = ({
...
   } else if (summaryMatch) {
     activeStepTitle = STEP_LABELS.overview;
+    currentStepTitle = STEP_LABELS.overview.props.defaultMessage[0].value;
   } else if (confirmationMatch) {
     activeStepTitle = STEP_LABELS.confirmation;
+    currentStepTitle = STEP_LABELS.confirmation.props.defaultMessage[0].value;
   } else {
     const step = steps.find(step => step.slug === stepSlug);
     activeStepTitle = step.formDefinition;
   }
```
sergei-maertens, line 54: This doesn't seem right for a couple of reasons:

1. `STEP_LABELS.foo` are `FormattedMesage` components and the internal API is private. Better would be to instead change that definition to use `intl.defineMessages` so we can get the string variant in the calling code by using `intl.formatMessage`
2. The progress-indicator is not guaranteed to be rendered, it can be disabled at the form configuration level, so this is the wrong place to hook into setting the title.

Instead, I'd expect the following components to be updated to set the title:

* `FormStart`
* `Summary`
* `SubmissionConfirmation`
* `PaymentOverview`

These are the components inside of each `Route` in the `Form` component

sergei-maertens (changes requested):

src/components/FormStep.js
jiromaykin: @sergei-maertens  An elegant idea, but "import {useTitle as reactUseTitle} from 'react-use';" in this way is not used anywhere, so where would we set "reactUseTitle" in this structure?

jiromaykin (commented):
This branch diverges too much from main so will be deleted eventually.

jiromaykin:
This branch diverges too much from main so will be deleted eventually.

PR #293, opened 2022-11-24
🐛 [open-formulieren/open-forms#2384] Fix initial language switch

CharString:
Don't start a new submission if none were already started. The user may not have logged in yet, and if the form requires login, user won't have permission to go to first step yet.

Stopped my attempt at writing `Form.spec.js` when after 3 mocks and 2 contexts in, I still didn't come past the Loader. There are way to many moving parts to make Form testable in isolation. :disappointed: 

Fixes open-formulieren/open-forms#2384

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -163,14 +163,14 @@ const Form = ({form}) => {
   useEffect(
     () => {
       if (prevLocale === undefined) return;
-      if (intl.locale !== prevLocale) {
+      if (intl.locale !== prevLocale && state.submission) {
         removeSubmissionId();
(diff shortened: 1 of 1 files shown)


sergei-maertens:
> Stopped my attempt at writing Form.spec.js when after 3 mocks and 2 contexts in, I still didn't come past the Loader. There are way to many moving parts to make Form testable in isolation.

it's a heavily composited component and almost at the top of the tree, testing this in isolation makes little sense. It'd be better to have selenium tests covering this in the backend project.

sergei-maertens (approved):

PR #294, opened 2022-11-25
Set semantic headings

jiromaykin:
fixes https://github.com/open-formulieren/open-forms/issues/2267

### DIFF
--- a/src/components/FormStepSummary/index.js
+++ b/src/components/FormStepSummary/index.js
@@ -42,7 +42,9 @@ const FormStepSummary = ({slug, name, data, editStepText}) => {
   return (
     <div className={getBEMClassName('summary')}>
       <div className={getBEMClassName('summary__step-header')}>
-        <div className={getBEMClassName('summary__step-name')}>{name}</div>
+        <div className={getBEMClassName('summary__step-name')}>
+          <h3>{name}</h3>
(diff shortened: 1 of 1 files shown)


src/components/FormStepSummary/index.js
```diff
@@ -42,7 +42,9 @@ const FormStepSummary = ({slug, name, data, editStepText}) => {
   return (
     <div className={getBEMClassName('summary')}>
       <div className={getBEMClassName('summary__step-header')}>
-        <div className={getBEMClassName('summary__step-name')}>{name}</div>
+        <div className={getBEMClassName('summary__step-name')}>
+          <h3>{name}</h3>
+        </div>
```
sergei-maertens, line 47:
```suggestion
        <h3 className={getBEMClassName('summary__step-name')}>{name}</h3>
```
is there a reason we can't do this?

We really don't want to rely on global/browser-default styles for UI elements and every component needs to be explicitly styled via a BEM class name. Haven't checked how this looks, but this should look the same and be semantic, no?

jiromaykin: Yes, I think still works for accessibility - I might still need to set margins and font sizes.

sergei-maertens (approved):

PR #295, opened 2022-11-25
Add support for pluggable/configurable replacement components

sergei-maertens:
Fixes open-formulieren/open-forms#1517

- [x] Split up the `Form` component into smart and "dumb" part
- [x] Provide mechanism to override default components

Quick demo with some alternative layout (ugly [code](https://github.com/open-formulieren/open-forms-sdk/pull/295/files#diff-e2e7a692ca6f6636dc97a0ca1fd80791aecfe66b5c6e2fcd4cfdcb9987c07138)!):

image

(also 
image
)

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -4,12 +4,15 @@ import {Switch, Route} from 'react-router-dom';
 
 import AppDebug from 'components/AppDebug';
 import Form from 'components/Form';
-import {Layout, LayoutRow} from 'components/Layout';
-import ManageAppointment from 'components/appointments/ManageAppointment';
 import LanguageSelection from 'components/LanguageSelection';
+import {LayoutRow} from 'components/Layout';
+import ManageAppointment from 'components/appointments/ManageAppointment';
+import {ConfigContext} from 'Context';
 import {I18NContext} from 'i18n';
 import Types from 'types';
 import {DEBUG} from 'utils';
+
+import AppDisplay from './AppDisplay';
 
 const LanguageSwitcher = () => {
   const {languageSelectorTarget: target} = useContext(I18NContext);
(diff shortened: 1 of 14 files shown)


src/components/App.js
```diff
@@ -26,32 +29,30 @@ const LanguageSwitcher = () => {
...
 const App = ({...props}) => {
+  const config = useContext(ConfigContext);
   const {
     form: {translationEnabled},
     noDebug = false,
   } = props;
+
+  const AppDisplayComponent = config?.displayComponents?.app ?? AppDisplay;
+
+  const languageSwitcher = translationEnabled ? <LanguageSwitcher /> : null;
```
SilviaAmAm, line 40: I am wondering what is the reason of passing the jsx as prop to AppDisplay, instead of the component (`<LanguageSwitcher />` vs `LanguageSwitcher`)? (also, I guess there will come a `LanguageSwitcherDisplay`?)

Because if we used just the component, we could also specify here which component should be used for the `ProgressIndicatorDisplay`?

src/index.js
```diff
@@ -1,5 +1,6 @@
 import {OpenForm} from './sdk';
 import './index.scss';
+// import displayComponents from './custom-display-components';
```
SilviaAmAm, line 3:
```suggestion

```

src/index.js
```diff
@@ -10,6 +11,7 @@ window.onload = () => {
     baseUrl: REACT_APP_BASE_API_URL,
     formId: formId || REACT_APP_FORM_ID,
     basePath: '/',
+    // displayComponents,
```
SilviaAmAm, line 14:
```suggestion

```

src/components/App.js
sergei-maertens: `LanguageSelectionDisplay` already exists, see https://open-formulieren.github.io/open-forms-sdk/?path=/story/composites-language-selection--display&globals=locale:nl;locales.nl:Nederlands;locales.en:English

We deliberately pass an element (component bound with props, in this case none) instead of a component for which the props need to be provided, so that the layout component just needs to deal with re-organizing the layout of what goes where instead of calling the nested components with the correct props (which would require us to provide a mechanism to access those props).

src/index.js
sergei-maertens: it's there deliberately - disabled by default, but easily enabled by uncommenting the commented lines

src/index.js
sergei-maertens: see above

PR #296, opened 2022-11-25
🐛 Fix loading translated literals and progress steps

CharString:
Closes open-formulieren/open-forms#2391

### DIFF
--- a/src/i18n.js
+++ b/src/i18n.js
@@ -53,7 +53,7 @@ const I18NContext = React.createContext({
 });
 I18NContext.displayName = 'I18NContext';
 
-const I18NManager = ({languageSelectorTarget, children}) => {
+const I18NManager = ({languageSelectorTarget, onLanguageChangeDone, children}) => {
   const {baseUrl} = useContext(ConfigContext);
   const [languageCode] = useGlobalState(currentLanguage);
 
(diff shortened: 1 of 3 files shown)


src/i18n.js
```diff
@@ -92,6 +95,7 @@ const I18NManager = ({languageSelectorTarget, children}) => {
 
 I18NManager.propTypes = {
   languageSelectorTarget: PropTypes.instanceOf(Element),
+  onLanguageChangeDone: PropTypes.instanceOf(Function),
```
sergei-maertens, line 98:
```suggestion
  onLanguageChangeDone: PropTypes.func,
```
see https://reactjs.org/docs/typechecking-with-proptypes.html

src/i18n.js
```diff
@@ -75,14 +75,17 @@ const I18NManager = ({languageSelectorTarget, children}) => {
 
   const {messages, formioTranslations} = value;
 
-  const onLanguageChangeDone = newLanguageCode => {
+  const logLanguageChangeDone = newLanguageCode => {
+    onLanguageChangeDone(newLanguageCode);
     console.log(`Changed language to: ${newLanguageCode}`);
   };
```
sergei-maertens, line 81: we can drop the `console.log`, so you can just forward the `onLanguageChangeDone` prop

sergei-maertens (changes requested):
nice tests :+1:

sergei-maertens (approved):

PR #297, opened 2022-11-25
Issue/2272 show unique page titles

jiromaykin:
fixes https://github.com/open-formulieren/open-forms/issues/2272


sergei-maertens (approved):
Made a couple minor tweaks, the general idea is:

* formatted message description is intended for translators and should provide sufficient context how/where the message is used
* group imports and sort them alphebetically (by module and name) within a group: third party packages, first-party packages and then local imports
*

PR #301, opened 2022-12-01
:wheelchair: nr.2266 Text-Alternatives with aria-labels and ALT 

jiromaykin:
Fixes open-formulieren/open-forms#2266

by adding appropriate aria-labels and alt tags in images where possible; sometimes with use of title attribute if needed.


src/components/ProgressIndicator/CompletionMark.js
```diff
@@ -1,11 +1,34 @@
...
 import PropTypes from 'prop-types';
+import {useIntl} from 'react-intl';
 
 import FAIcon from 'components/FAIcon';
 
 const CompletionMark = ({completed = false}) => {
+  const intl = useIntl();
+  // Wrapper may be a DOM element, which can't handle <FormattedMessage />
+  const ariaLabel = intl.formatMessage({
+    description: 'Completion message',
```
sergei-maertens, line 11: Try to be as specific as possible in these descriptions - they are used to generate the ID of a message.

```suggestion
    description: 'Step completion marker icon label',
```

src/components/ProgressIndicator/ProgressIndicatorDisplay.js
```diff
@@ -115,7 +117,11 @@ const ProgressIndicatorDisplay = ({
         onClick={() => setExpanded(!expanded)}
       >
         <FAIcon icon={expanded ? 'chevron-up' : 'chevron-down'} modifiers={['normal']} />
-        <span className={getBEMClassName('progress-indicator__active-step')}>
+        <span
+          className={getBEMClassName('progress-indicator__active-step')}
+          aria-label="activeStepTitle"
+          title="active step title"
```
sergei-maertens, line 123: these should also be translatable, no?

src/formio/templates/signature.ejs
```diff
@@ -9,7 +9,10 @@
     class="btn btn-sm btn-light signature-pad-refresh utrecht-button utrecht-button--html-button utrecht-button--subtle"
     ref="refresh"
   >
-    <i class="{{ctx.iconClass('refresh')}}"></i>
+    <i class="{{ctx.iconClass('refresh')}}">
+      <!-- For accessibility a Button element must contain text content -->
+      <span class="sr-only">Refresh</span>
```
sergei-maertens, line 14: pass this through `{{ctx.t('Refresh')}}` to support translations/internationalization

sergei-maertens (changes requested):

src/formio/templates/signature.ejs
jiromaykin: @sergei-maertens Should _<span class="sr-only">_ then become _<span class="{{ctx.iconClass('sr-only')}}">Refresh</span>_ here? (Since that is where the text is situated?)
Or should the end-result simply be {{ctx.**iconClass.t(**'Refresh')}} that surrounds all?

sergei-maertens: nah, just stick to hardcoded `sr-only` here. Formio supports other icon libraries than font-awesome (technically), but we're not using that feature/functionality.

The end result should thus be:

```suggestion
      <span class="sr-only">{{ctx.t('Refresh')}}</span>
```

src/components/ProgressIndicator/ProgressIndicatorDisplay.js
```diff
@@ -108,17 +110,37 @@ const ProgressIndicatorDisplay = ({
...
+  });
+  const toggleStepsTitle = intl.formatMessage({
+    description: 'Active step toggle titles',
+    defaultMessage: 'Show/hide steps',
+  });
+
   return (
     <Card blockClassName="progress-indicator" modifiers={expanded ? [] : ['mobile-collapsed']}>
-      <div
+      {/* Turn a Div with an onClick event into a button for accesibility */}
```
jiromaykin, line 126: @sergei-maertens I am assuming that: if a visually impaired user Zooms the screen, the mobile view will become triggered, where they should still be able to navigate the form-steps, so I am proposing to turn the div into a real button so it can have focus and can be operated with the space-bar etc.

sergei-maertens: I checked, had to scroll to zoomlevel 200% before mobile view was triggered, but that is indeed what happens. Proposal is okay.

src/components/ProgressIndicator/ProgressIndicatorDisplay.js
```diff
@@ -108,17 +110,37 @@ const ProgressIndicatorDisplay = ({
...
+      <button
         className={getBEMClassName('progress-indicator__mobile-header')}
+        aria-pressed={expanded ? 'false' : 'true'}
```
sergei-maertens, line 129: Reading on [mdn](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-pressed):

> Toggle buttons require a full press-and-release cycle to change their value. Pressing and releasing it once changes the value to true. If it's pressed and released again, the value changes back to false.

Pressing the button expands the menu, making it visible, so shouldn't this be swapped?

```suggestion
        aria-pressed={expanded ? 'true' : 'false'}
```

Also following the recommendations, the icon inside the button changes, so that should probably be `aria-hidden="true"` so a screen reader will be reporting the correct thing? Might also need some aria-label to give a more meaningful label to the button instead of the active step title.

src/components/ProgressIndicator/ProgressIndicatorDisplay.js
```diff
@@ -108,17 +110,37 @@ const ProgressIndicatorDisplay = ({
...
+
   return (
     <Card blockClassName="progress-indicator" modifiers={expanded ? [] : ['mobile-collapsed']}>
-      <div
+      {/* Turn a Div with an onClick event into a button for accesibility */}
+      <button
         className={getBEMClassName('progress-indicator__mobile-header')}
+        aria-pressed={expanded ? 'true' : 'false'}
         onClick={() => setExpanded(!expanded)}
       >
```
sergei-maertens, line 131: please add some CSS to make this `width: 100vw` - I now have to tab the icon or step text instead of the "bar" to toggle it.

src/components/ProgressIndicator/ProgressIndicatorDisplay.js
```diff
@@ -108,17 +110,37 @@ const ProgressIndicatorDisplay = ({
...
-        <FAIcon icon={expanded ? 'chevron-up' : 'chevron-down'} modifiers={['normal']} />
-        <span className={getBEMClassName('progress-indicator__active-step')}>
+        <FAIcon
+          icon={expanded ? 'chevron-up' : 'chevron-down'}
+          modifiers={['normal']}
+          aria-label={ariaIconLabel}
+        />
+        <span
+          className={getBEMClassName('progress-indicator__active-step')}
+          title={toggleStepsTitle}
```
sergei-maertens, line 139: shouldn't this be `aria-label` instead of `title` so that screen readers pick this up?

src/scss/components/_label.scss
```diff
@@ -11,7 +11,7 @@
   &.field-required:after {
     content: '*';
     color: var(--of-color-danger);
-    z-index: 1; // otherwise this is above the progress indicator
+    z-index: 0; // otherwise this is above the progress indicator
```
sergei-maertens, line 14: both seem to work visually - what is/was the problem with the z-index being `1`?

src/scss/components/_progress-indicator.scss
```diff
@@ -18,6 +18,14 @@
     @include body--big;
     @include margin(-$grid-container-margin, $properties: margin-left);
 
+    // style for replacing div with a button for accessibility
+    background: none;
+    color: inherit;
+    border: none;
+    padding: 0;
+    font-family: $typography-font-family-base !important;
```
sergei-maertens, line 26: don't think this rule is required, since the `@body` mixin sets the font-family correctly to `font-family: var(--of-typography-sans-serif-font-family);`

src/scss/components/_progress-indicator.scss
```diff
@@ -18,6 +18,14 @@
     @include body--big;
     @include margin(-$grid-container-margin, $properties: margin-left);
 
+    // style for replacing div with a button for accessibility
+    background: none;
+    color: inherit;
```
sergei-maertens, line 23: the `@include body` sets the color already to `var(--of-color-fg)` which is correct.

src/scss/components/_progress-indicator.scss
```diff
@@ -39,10 +47,8 @@
 
     // style layout
     @at-root .#{prefix('layout__row')} & {
-      @include margin(
-        calc(-1 * var(--of-progress-indicator-mobile-margin)),
-        $properties: (margin-left, margin-right)
-      );
+      @include margin(calc(-1 * var(--of-progress-indicator-mobile-margin)),
+        $properties: (margin-left, margin-right));
```
sergei-maertens, line 51: this looks weird and the previous version was already what prettier outputs, so please restore.

src/scss/components/_progress-indicator.scss
```diff
@@ -77,4 +83,4 @@
   @include element('label') {
     flex-grow: 1;
   }
-}
+}
```
sergei-maertens, line 86: please restore the newlines at end of files (cfr. `.editorconfig` file)

sergei-maertens (changes requested):

src/scss/components/_progress-indicator.scss
jiromaykin: In this case the 'styled-like-a-div' button kept on forcing its default text-styles and I needed !important to really set the font-family.

src/scss/components/_progress-indicator.scss
jiromaykin: I will double-chek but in this case the Button kept on forcing its default button background.

src/scss/components/_progress-indicator.scss
sergei-maertens: hm, maybe that's a FF-quirk then?

src/scss/components/_progress-indicator.scss
sergei-maertens: could be a FF vs. Brave quirk :)

src/scss/components/_label.scss
jiromaykin: Amazingly this differs in Firefox and Chrome on Ubuntu: in FF the asterisks for required fields remain visible when scrolling through the progress indicator mobile view. I think I will need to test this on PC and Mac before making a decision.

jiromaykin: Indeed testing FF on Mac and Windows did not reproduce the same bug - so I'm switching to Chrome for front-end.

sergei-maertens (approved):

PR #303, opened 2022-12-06
:wheelchair: autocomplete assistance

jiromaykin:
Add autocomplete to form fields wherever possible.


jiromaykin:
Unnecessary solution.

Issue #304, opened 2022-12-09
Add Signature component to storybook


sergei-maertens:
We currently don't have all formio components hooked up to storybook, which excludes them from visual regression testing.

1. Add a story for the signature component in `src/formio/components/Signature.stories.mdx`
2. Check the backend form designer for the required JSON for this component
3. Configure/document the story similarly to the other formio components

PR #312, opened 2022-12-09
:memo: [#304] add Storybook Signature component

jiromaykin:
fixes open-formulieren/open-forms-sdk#304


src/formio/components/Signature.stories.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+## Default
+
+export const Template = ({
+  backgroundColor = '',
+  description = '',
+  footer = '',
+  id = '',
+  label = '',
+  penColor = '',
```
sergei-maertens, line 44: while these are properties that can be specified at the formio level, we don't expose them in our own flavour of the form builder, so we should leave them out to not give wrong impressions:

* backgroundColor
* footer
* id
* penColor

These args (and their argtype definitions) should be removed.

sergei-maertens (approved):

PR #313, opened 2022-12-09
[OF#2447] Display login buttons

SilviaAmAm:
Fixes open-formulieren/open-forms#2447

Related Backend PR: https://github.com/open-formulieren/open-forms/pull/2462

Desktop:
Screenshot from 2022-12-09 13-32-27

Mobile:
Screenshot from 2022-12-09 13-32-46


src/components/LoginButton/index.js
```diff
@@ -1,32 +1,30 @@
...
-      defaultMessage="Login with {provider}"
-      values={{provider: option.label}}
-    />
-  </Button>
+  <div className={getBEMClassName('login-button')}>
+    <Button
+      variant="primary"
+      component="a"
+      href={option.url}
+      onClick={option.onClick || noop}
```
sergei-maertens, line 17: anchors shouldn't have `onClick` specified at all

src/components/LoginButton/test.spec.js
```diff
@@ -43,11 +43,8 @@ it('Login button has the right URL after cancelling log in', () => {
     );
   });
 
-  const expectedUrl = new URL(option.url);
-  expectedUrl.searchParams.set('next', 'https://open-forms.nl/digid-form/?_start=1');
-
   const actualUrl = container.getElementsByClassName('openforms-button')[0].href;
 
-  expect(actualUrl).toEqual(expectedUrl.toString());
+  expect(actualUrl).toEqual(option.url);
```
sergei-maertens, line 48: doesn't this break the whole redirect flow if the `next` param is not in the URL anymore?

src/components/LoginOptions/LoginOptionsDisplay.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import {FormattedMessage} from 'react-intl';
+import PropTypes from 'prop-types';
+
+import {getBEMClassName} from 'utils';
+import LoginButton from 'components/LoginButton';
+import FormattedLoginOption from 'types/FormattedLoginOption';
```
sergei-maertens, line 7: please sort the imports alphabetically

src/components/LoginOptions/index.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import {FormattedMessage} from 'react-intl';
+
+import Types from 'types';
+import PropTypes from 'prop-types';
+import Literal from 'components/Literal';
+import {getLoginUrl} from 'components/utils';
```
sergei-maertens, line 7: import ordering

src/components/LoginOptions/index.js
```diff
@@ -0,0 +1,55 @@
...
+    }
+  });
+
+  if (!form.loginRequired) {
+    loginAsYourselfOptions.push({
+      identifier: 'anonymous',
+      label: <Literal name="beginText" />,
+      url: '#',
+      onClick: onFormStart,
+    });
```
sergei-maertens, line 39: maybe we should make this `onFormStart` thing more semantic:

1. if login is not required, render the form start page between `form` tags, otherwise use a `div`
2. render the anonymous login option/button as an actual `type="submit"` button, when it's clicked, it submits the outer `form`
3. hook up the outer `form` with `onSubmit={onFormStart}`. This decouples the form start event from the actual buttons/links. Because the login buttons are links to the backend, they will not result in a form start submit and instead invoke the form start via the URL query parameter.

This approach allows us to not have to deal with `onClick` at all in login button components.

src/scss/components/_login-button-logo.scss
```diff
@@ -0,0 +1,21 @@
+@use 'sass:math';
+
+@import '~microscope-sass/lib/typography';
+
+@import '../mixins/prefix';
+
+.#{prefix(login-button-logo)} {
+  display: flex;
+  align-items: center;
+  margin-left: math.div($grid-container-margin-base, 2);
```
sergei-maertens, line 10: can't we use the `gap` flex property in the parent component instead?

src/scss/components/_login-button.scss
```diff
@@ -0,0 +1,18 @@
...
+
+@import '~microscope-sass/lib/typography';
+
+@import '../mixins/prefix';
+
+.#{prefix(login-button)} {
+  display: flex;
+  align-items: center;
+
+  margin-bottom: math.div($grid-container-margin-base, 2);
```
sergei-maertens, line 11: this now creates extra white space if there are no login button icons at all - I'd only apply this if it's not the last child (`:not(:last-child)`)

This is also encapsulated in the microscope-sass `@include margin(auto)` which uses the correct/relevant variables for this.

sergei-maertens (changes requested):

src/components/CoSign.js
```diff
@@ -35,26 +34,27 @@ const CoSignAuthentication = ({form, submissionUuid, saveStepData, authPlugin})
...
 
-  // add the co sign submission parameter to the login URL
+  // add the co-sign submission parameter to the login URL
   const loginUrl = new URL(loginOption.url);
   loginUrl.searchParams.set('coSignSubmission', submissionUuid);
-  const modifiedLoginOption = {...loginOption, url: loginUrl.toString()};
+  const modifiedLoginOption = {
+    ...loginOption,
+    url: loginUrl.toString(),
+    onClick: noop,
```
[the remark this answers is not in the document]
SilviaAmAm: We have 2 behaviours for the buttons:
- For the anonymous start: here we want the `onFormStart` to be called when the button is clicked, but there is no URL to go to.
- For the login buttons: here we have a URL, but we don't need a `onClick` handler.

So I had the `onClick` doing nothing here on purpose

src/components/LoginButton/LoginButtonIcon.js
```diff
@@ -1,28 +1,26 @@
...
+  const {href, imageSrc, title} = logo;
+  return (
+    <a
+      className={getBEMClassName('login-button-logo')}
+      href={href}
+      key={identifier}
+      onClick={event => {
+        event.preventDefault();
+        window.open(href, '_blank', 'noopener,noreferrer');
+      }}
```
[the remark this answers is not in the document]
SilviaAmAm: Without the `event.preventDefault()`, the link to DigiD/eHerkenning.. was being opened both in the current tab _and_ in a new tab :thinking:

src/components/LoginOptions/index.js
SilviaAmAm: I changed it, although now there is more logic in the `LoginButton` component to figure out the URLs/type of component/... :thinking: But I like not having the `onClick`

src/components/CoSign.js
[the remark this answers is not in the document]
SilviaAmAm: After changing the behaviour as you suggested below, the onClick is no longer needed

src/components/LoginButton/test.spec.js
SilviaAmAm: It is not the `LoginButton` now that makes the URL but the 'smart' `LoginOptions` component. So the `LoginButton` just uses the URL that is passed as prop.

Maybe this test makes no sense anymore and I should test the `LoginOptions` component.

src/components/LoginOptions/LoginOptions.stories.mdx
```diff
@@ -0,0 +1,178 @@
...
+
+## Functional
+
+The `LoginOptions` component handles the 'smart' part of the component. It separates the login options between
+the ones that can be used to 'login as yourself' and the ones that can be used to log in 'on behalf of someone else'.
+It also formats the labels that will be displayed on the buttons and creates the right URL to which the user will be
+redirected upon clicking one of the buttons.
+
+export const Template = ({form, onFormStart}) => (
+  <LiteralsProvider literals={{'beginText': {resolved: 'Begin Form'}}}>
```
sergei-maertens, line 98: Probably would be a bit neater to use the decorator approach here for the provider, see for example:

* https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/LanguageSelection/LanguageSelection.stories.mdx?plain=1#L160
* https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/LanguageSelection/LanguageSelection.stories.mdx?plain=1#L166

src/components/LoginOptions/LoginOptions.stories.mdx
```diff
@@ -0,0 +1,178 @@
...
+      loginAsYourselfOptions: [
+        {
+          identifier: "digid",
+          label: "DigiD",
+          logo: {
+              href: "https://www.digid.nl/",
+              title: "DigiD",
+              imageSrc: "/digid.png"
+          },
+          url: "http://localhost:8999/auth/form-slug/digid/start?next=http%3A%2F%2Flocalhost%3A3000%2F%3F_start%3D1",
```
sergei-maertens, line 42: maybe make these URLs `#` instead for the stories - clicking the links now gives errors and takes you out of storybook, while with the `#` it'll open up the underlying iframe instead

src/components/LoginOptions/LoginOptions.stories.mdx
```diff
@@ -0,0 +1,178 @@
...
+              href: "https://www.eherkenning.nl/",
+              title: "eIDAS",
+              imageSrc: "/eidas.png"
+          },
+          url: "http://localhost:8999/auth/form-slug/eidas/start?next=http%3A%2F%2Flocalhost%3A3000%2F%3F_start%3D1",
+        },
+        {
+          identifier: "anonymous",
+          label: "Begin form",
+          url: "#",
```
sergei-maertens, line 67: this breaks the `if (!url)` check in the `LoginButton` - is that deliberate? I'd say it's probably more representative to leave the URL empty here so that the DOM in storybook correctly renders a `button` element instead of an anchor.

See also the remark below

src/components/LoginOptions/LoginOptions.stories.mdx
```diff
@@ -0,0 +1,178 @@
...
+        logo: {
+            href: "https://www.digid.nl/",
+            title: "DigiD Machtigen",
+            imageSrc: "/digid.png"
+        },
+        url: "http://localhost:8999/auth/issue-2447-style-login-page/digid_machtigen/start?next=http%3A%2F%2Flocalhost%3A3000%2F%3F_start%3D1",
+      }]
+    }}
+  >
+    {DisplayTemplate.bind({})}
```
sergei-maertens, line 82: perhaps adding a `<form>` container around this (possibly as a decorator?) makes sense to simulate realistic behaviour:

```suggestion
    <form onSubmit={e => {
      e.preventDefault();
      alert('form start triggered');
    }}>
      {DisplayTemplate.bind({})}
    </form>
```

This also better documents the expected usage (inside of a `<form>` container)

src/components/LoginOptions/LoginOptions.stories.mdx
```diff
@@ -0,0 +1,178 @@
...
+      onFormStart={onFormStart}
+    />
+  </LiteralsProvider>
+);
+
+<Canvas>
+  <Story
+    name="Functional"
+    args={{
+      form: {
```
sergei-maertens, line 110: check the console in storybook too - it warns about the missing `uuid` prop` for `form`

src/components/LoginOptions/LoginOptions.stories.mdx
```diff
@@ -0,0 +1,178 @@
...
+            logo: {
+              title: "eHerkenning bewindvoering",
+              imageSrc: "/eherkenning.png",
+              href: "https://www.eherkenning.nl/"
+            },
+            isForGemachtigde: true
+          },
+        ]
+      },
+      onFormStart: () => {},
```
sergei-maertens, line 169: this now submits for real, leading to a nonsense error/page, so I'd also implement an `event.preventDefault()` and `alert` strategy here purely for storybook purposes

I'd also expect this to be picked up as an interaction which could be tested via the `play` function :-) For an example of such tests, see: https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/LanguageSelection/LanguageSelection.stories.mdx?plain=1#L117

src/components/LoginOptions/index.js
```diff
@@ -0,0 +1,69 @@
+import React, {useContext} from 'react';
+import {FormattedMessage} from 'react-intl';
+import PropTypes from 'prop-types';
+
+import Literal from 'components/Literal';
+import {getLoginUrl} from 'components/utils';
+import {ConfigContext} from 'Context';
+
+import Types from 'types';
+import LoginOptionsDisplay from './LoginOptionsDisplay';
```
sergei-maertens, line 10:
```suggestion
import Literal from 'components/Literal';
import {getLoginUrl} from 'components/utils';
import {ConfigContext} from 'Context';
import Types from 'types';

import LoginOptionsDisplay from './LoginOptionsDisplay';
```

src/components/LoginOptions/index.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  return (
+    <form onSubmit={onFormStart}>
+      <LoginDisplayComponent
+        loginAsYourselfOptions={loginAsYourselfOptions}
+        loginAsGemachtigdeOptions={loginAsGemachtigdeOptions}
+      />
+    </form>
+  );
```
sergei-maertens, line 61: you can simplify this a bit to avoid repetition:

```jsx
const Container = form.loginRequired ? React.Fragment : 'form';
return (
  <Container onSubmit={onFormStart}>
    <LoginDisplayComponent
        loginAsYourselfOptions={loginAsYourselfOptions}
        loginAsGemachtigdeOptions={loginAsGemachtigdeOptions}
      />
  </Container>
);
```

I'm not sure if you can just pass whatever props to `React.Fragment` or not, if not then the container props need to be defined in a similar manner.

src/scss/components/_login-button-logo.scss
```diff
@@ -0,0 +1,20 @@
+@use 'sass:math';
```
sergei-maertens, line 1: unused import

src/scss/components/_login-button-logo.scss
```diff
@@ -0,0 +1,20 @@
...
+
+@import '~microscope-sass/lib/typography';
```
sergei-maertens, line 3: also unused, and I think it should be `responsive` instead of `typography` for the `mobile-only` mixin, but it kinda works because another file already loaded it.

We should refactor into using proper sass modules at some point :grimacing:

src/scss/components/_login-button.scss
```diff
@@ -0,0 +1,20 @@
+@use 'sass:math';
+
+@import '~microscope-sass/lib/typography';
```
sergei-maertens, line 3:
```suggestion
@import '~microscope-sass/lib/responsive';
```

src/components/CoSign.js
```diff
@@ -7,22 +7,21 @@ import {get} from 'api';
...
 
 const getCosignStatus = async (baseUrl, submissionUuid) => {
   const endpoint = `${baseUrl}submissions/${submissionUuid}/co-sign`;
-  const response = await get(endpoint);
-  return response;
+  return await get(endpoint);
 };
 
 const CoSignAuthentication = ({form, submissionUuid, saveStepData, authPlugin}) => {
+  const config = useContext(ConfigContext);
```
sergei-maertens, line 23: the idea is nice, however this won't work because the formio component calls `ReactDOM.render` and I don't think this shares the context from the top-level react root (our structure is `react root -> formio root -> react root` and the react roots are independent).

The display config should be passed down in `FormStep.js` to the formio config, which the `src/formio/components/CoSign.js` should then use in the `ConfigContext.Provider` in the `renderReact` method.

src/components/LoginButton/LoginButtonIcon.js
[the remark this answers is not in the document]
sergei-maertens:
```jsx
    <a
      className={getBEMClassName('login-button-logo')}
      href={href}
      key={identifier}
      target="_blank"
      rel="noreferrer nofollower"
    >
      <img className={getBEMClassName('login-button-logo__image')} alt={title} src={imageSrc} />
    </a>
```
this works for me in Chromium based browsers :thinking:

sergei-maertens (changes requested):

src/components/CoSign.js
```diff
@@ -7,22 +7,21 @@ import {get} from 'api';
 import Body from 'components/Body';
 import ErrorMessage from 'components/ErrorMessage';
 import Loader from 'components/Loader';
-import LoginButton, {LoginButtonIcon} from 'components/LoginButton';
 import {ConfigContext, SubmissionContext} from 'Context';
-import {Toolbar, ToolbarList} from 'components/Toolbar';
 import Types from 'types';
-
 import {getBEMClassName} from 'utils';
+import LoginOptionsDisplay from 'components/LoginOptions/LoginOptionsDisplay';
```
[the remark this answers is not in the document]
SilviaAmAm: :see_no_evil:

src/scss/components/_login-button-logo.scss
SilviaAmAm: I thought that typography was needed for the `$grid-row-height`, but that's in `grid` and `typography` imports `grid`. So I removed `typography`, added `grid` and `responsive`

src/components/LoginOptions/LoginOptions.stories.mdx
```diff
@@ -0,0 +1,221 @@
...
+    loginAsGemachtigdeOptions={loginAsGemachtigdeOptions}
+  />
+);
+
+<Canvas>
+  <Story
+    name="Display"
+    decorators={[(Story) => (
+      <form>
+        {Story()}
```
SilviaAmAm, line 35: I used `{Story()}` instead of `<Story/>` because I was having this issue: https://github.com/storybookjs/storybook/issues/12596

No idea why it works fine in the language component story.

src/components/LoginOptions/LoginOptions.stories.mdx
```diff
@@ -0,0 +1,221 @@
...
+## Functional
+
+The `LoginOptions` component handles the 'smart' part of the component. It separates the login options between
+the ones that can be used to 'login as yourself' and the ones that can be used to log in 'on behalf of someone else'.
+It also formats the labels that will be displayed on the buttons and creates the right URL to which the user will be
+redirected upon clicking one of the buttons.
+
+export const LiteralDecorator = (Story) => (
+  <LiteralsProvider literals={{'beginText': {resolved: 'Begin Form'}}}>
+    {Story()}
```
sergei-maertens, line 104: `<Story />` seems to work, but note that I also added some `parameter` config in the `Meta`:

```jsx
<Meta
  title="Composites/Login Options"
  component={LoginOptions}
  parameters={{
    docs: {
      source: {
        type: 'dynamic',
        excludeDecorators: true,
      }
    }
  }}
/>
```

`excludeDecorators` is crucial for accurate code previews

src/components/LoginOptions/LoginOptions.stories.mdx
```diff
@@ -0,0 +1,272 @@
...
+        url: "#",
+      }]
+    }}
+  >
+    {DisplayTemplate.bind({})}
+  </Story>
+</Canvas>
+
+Note that the anonymous form start option is implemented as a button with `type="submit"` and is therefore
+expected to be nested inside of a `<form>` element with an appropriate `onSubmit` handler.
```
sergei-maertens, line 101: I've added an explicit note here

src/components/LoginOptions/LoginOptions.stories.mdx
```diff
@@ -0,0 +1,272 @@
...
+The `LoginOptions` component handles the 'smart' part of the component. It separates the login options between
+the ones that can be used to 'login as yourself' and the ones that can be used to log in 'on behalf of someone else'.
+It also formats the labels that will be displayed on the buttons and creates the right URL to which the user will be
+redirected upon clicking one of the buttons.
+
+export const LiteralDecorator = (Story, { args }) => (
+  <LiteralsProvider literals={{'beginText': {resolved: args.beginText}}}>
+    <Story />
+  </LiteralsProvider>
+);
```
sergei-maertens, line 118: I've made it so that the button text can be specified via the story args

src/components/LoginOptions/LoginOptions.stories.mdx
```diff
@@ -0,0 +1,272 @@
...
+    name="Functional"
+    decorators={[LiteralDecorator]}
+    play={async ({args, canvasElement}) => {
+      const canvas = within(canvasElement);
+      let anonymousStartButton = await canvas.getByRole('button');
+      await userEvent.click(anonymousStartButton);
+      await waitFor(() => expect(args.onFormStart).toHaveBeenCalled());
+    }}
+    argTypes={{
+      form: {table: {disable: true}},
```
sergei-maertens, line 162: the prop types are automatically added to the canvas controls, but since we don't use this, we can explicitly remove/disable it from the canvas controls.

src/components/LoginOptions/LoginOptions.stories.mdx
```diff
@@ -0,0 +1,272 @@
...
+      onFormStart: {
+        action: true,
+      }
+    }}
+    args={{
+      beginText: 'Begin form (anonymous)',
+      loginRequired: false,
```
sergei-maertens, line 169: I've added this arg so that we get a nice toggle true/false and see how the story reacts to that, as it has an effect on the container element + additional button.

src/components/LoginOptions/LoginOptions.stories.mdx
```diff
@@ -0,0 +1,272 @@
...
+      onFormStart: jest.fn(e => e.preventDefault()),
+    }}
+  >
+    {Template.bind({})}
+  </Story>
+</Canvas>
+
+<Canvas>
+  <Story
+    name="No \'machtigen\' options"
```
sergei-maertens, line 236: I added a story for the situation without machtigen options so that we have visual regression testing to ensure the title is absent too if there are none.

src/components/LoginOptions/index.js
```diff
@@ -0,0 +1,62 @@
...
+  if (!form.loginRequired) {
+    loginAsYourselfOptions.push({
+      identifier: 'anonymous',
+      label: <Literal name="beginText" />,
+    });
+  }
+
+  const LoginDisplayComponent = config?.displayComponents?.loginOptions ?? LoginOptionsDisplay;
+  const Container = form.loginRequired ? React.Fragment : 'form';
+  const containerProps = form.loginRequired ? {} : {onSubmit: onFormStart};
```
sergei-maertens, line 45: console was complaining about these props after all, so I've broken this up too.

sergei-maertens (approved):

Issue #306, opened 2022-12-09
Add the Content component to storybook


sergei-maertens:
We currently don't have all formio components hooked up to storybook, which excludes them from visual regression testing.

1. Add a story for the content component in `src/formio/components/Content.stories.mdx`
2. Check the backend form designer for the required JSON for this component ("Vrije tekst")
3. Configure/document the story similarly to the other formio components

PR #315, opened 2022-12-10
:memo: [#306] add Storybook Content component

jiromaykin:
Fixes open-formulieren/open-forms-sdk#306

### DIFF
--- a/src/formio/components/Content.stories.mdx
+++ b/src/formio/components/Content.stories.mdx
@@ -12,13 +12,6 @@ import {FormioComponent, FormDecorator} from './story-util';
     },
     customClass: {
       options: ['default', 'warning', 'info', 'error', 'success'],
-      mapping: {
-        default: false,
-        warning: false,
-        info: true,
-        error: false,
-        success: false,
-      },
       control: {
         type: 'radio',
         labels: {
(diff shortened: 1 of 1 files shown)


sergei-maertens (changes requested):
I'd also like to see an arg for the configuration option "Aangepaste CSS class" from the form designer.

src/formio/components/Content.stories.mdx
```diff
@@ -0,0 +1,54 @@
...
+      type: 'content',
+    }}
+  />
+);
+
+<Canvas>
+  <Story
+    name="Default"
+    args={{
+      customClass: 'breed',
```
sergei-maertens, line 44: this arg will have to be added to the `argTypes` above - the allowed list of values is limited in the backend: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/js/components/form/content.js#L13

You can use the control type of `radio` for this, see https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/ProgressIndicator/index.stories.mdx?plain=1#L13 for an example

sergei-maertens (changes requested):

jiromaykin:
Squashed some commits.

Issue #307, opened 2022-12-09
Add the fieldset component to storybook


sergei-maertens:
We currently don't have all formio components hooked up to storybook, which excludes them from visual regression testing.

1. Add a story for the fieldset component in `src/formio/components/Fieldset.stories.mdx`
2. Check the backend form designer for the required JSON for this component
3. Configure/document the story similarly to the other formio components - ensure there are some nested components and the JSON of nested components can be edited in the storybook canvas

PR #316, opened 2022-12-10
:memo: [#307] add Storybook Fieldset component

jiromaykin:
Fixes open-formulieren/open-forms-sdk#307

### DIFF
--- a/src/formio/components/Fieldset.stories.mdx
+++ b/src/formio/components/Fieldset.stories.mdx
@@ -128,15 +128,20 @@ export const Template = ({ descriptionFamilyName, descriptionGivenName, id = '',
   <Story
     name="Default"
     args={{
-      descriptionFamilyName: 'family-name',
-      descriptionGivenName: 'given-name',
-      id: 'veldengroep1',
-      label: 'Uw keuzevelden',
-      labelFirstName: 'Voornaam',
-      labelLastName: 'Achternaam',
-      labelPrivacy: 'Ik ga akkoord met de privacy-voorwaarden',
-      placeholderFirstName: 'voornaam...',
-      placeholderLastName: 'achternaam...',
+      id: 'fieldset',  // really not relevant or interesting at all tbh, we just have to provide something
+      label: 'The group of fields',  // English first for code/documentation!
+      hideHeader: false,
+      nestedComponents: [
+        {
+          key: 'given-name',
+          type: 'textfield',
+          input: true,
+          label: 'Given name',
+          description: 'Given name as in your passport/identity card.',
+          placeholder: 'Willy',
+        },
+        ...
+      ],
     }}
   >
     {Template.bind({})}
--- a/src/formio/components/Fieldset.stories.mdx
+++ b/src/formio/components/Fieldset.stories.mdx
@@ -13,6 +13,27 @@ import {FormioComponent, FormDecorator} from './story-util';
     label: {
       type: {name: 'string', required: false},
     },
+    descriptionFamilyName: {
+      type: {name: 'string'},
+    },
+    descriptionGivenName: {
+      type: {name: 'string'},
+    },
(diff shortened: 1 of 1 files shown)


src/formio/components/Fieldset.stories.mdx
```diff
@@ -0,0 +1,146 @@
...
+  parameters={{
+    controls: {sort: 'requiredFirst'},
+  }}
+/>
+
+# Field Set
+
+A Field Set can be used to create a title of an area of the form. This is useful to put inside
+Layout components or in between lots of related components. The Field Set is for display only and
+will not be saved to the API,
```
sergei-maertens, line 48: not entirely correct?

A fieldset is used to group a number of fields. The provided label will be used as title of the fieldset. Optionally, you can [hide](https://github.com/open-formulieren/open-forms/blob/master/src/openforms/js/components/form/fieldset.js#L18) this title.

A fieldset is a presentation-only component - no data is saved to the API.

src/formio/components/Fieldset.stories.mdx
```diff
@@ -0,0 +1,146 @@
...
+
+Additionally the use of a field set increases accessibility: grouping related form controls makes forms more understandable for all users, as related controls are easier to identify, makign it easier for people to focus on smaller and more manageable groups.
+
+**Legend**
+
+A Field Set label that will appear as a Legend for the component in the form.
```
sergei-maertens, line 54: there should be an arg/control for `hideHeader`, default is `false` (so it is visible).

src/formio/components/Fieldset.stories.mdx
```diff
@@ -0,0 +1,146 @@
...
+
+## Default
+
+export const Template = ({ descriptionFamilyName, descriptionGivenName, id = '', label = '', labelFirstName, labelLastName, labelPrivacy, placeholderFirstName, placeholderLastName, ...args }) => (
```
sergei-maertens, line 58: In this case I would structure the args quite differently - there are a number of args/controls directly related to the fieldset:

* `label`: content of the fieldset legend
* `id` -> mapped to formio `key`: `fieldset` as default value is fine
* `hideHeader`: controls whether the label is visible or not

And then there's the nested components inside the fieldset - you have a number of args that related to these nested components, but they now prevent me from pasting completely different components into the configuration in storybook to see how that looks. What if I wanted to check a `select` component?

So it'd be better to provide an arg `nestedComponents: [...]` with the current nested components, and leave it so you can edit that as one big blob.

src/formio/components/Fieldset.stories.mdx
```diff
@@ -0,0 +1,146 @@
...
+      descriptionGivenName: 'given-name',
+      id: 'veldengroep1',
+      label: 'Uw keuzevelden',
+      labelFirstName: 'Voornaam',
+      labelLastName: 'Achternaam',
+      labelPrivacy: 'Ik ga akkoord met de privacy-voorwaarden',
+      placeholderFirstName: 'voornaam...',
+      placeholderLastName: 'achternaam...',
+    }}
+  >
```
sergei-maertens, line 141:
```suggestion
  <Story
    name="Default"
    args={{
      id: 'fieldset',  // really not relevant or interesting at all tbh, we just have to provide something
      label: 'The group of fields',  // English first for code/documentation!
      hideHeader: false,
      nestedComponents: [
        {
          key: 'given-name',
          type: 'textfield',
          input: true,
          label: 'Given name',
          description: 'Given name as in your passport/identity card.',
          placeholder: 'Willy',
        },
        ...
      ],
    }}
  >
```
you can also strip the bloat of the component JSON to only keep the essentials.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #317, opened 2022-12-12
:memo: [#308] add Storybook File-Upload component

### LINKED ISSUE #308: Add the file upload component to storybook
We currently don't have all formio components hooked up to storybook, which excludes them from visual regression testing.

1. Add a story file for the file component in `src/formio/components/FileField.stories.mdx`
2. Check the backend form designer for the required JSON for this component
3. Ensure that a URL is set in `component.url` and a mock is set up for this endpoint

The following variants should be set up/documented:

* Happy flow - successful upload
* Error flow - backend error responds with a HTTP 400 with validation errors

The following controls should be available:

* single/multiple upload
* `filePattern` validation (multiple checkboxes, you can find the default list of extensions/mime types here: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/config/constants.py#L60

jiromaykin:
Closes #308

### DIFF
--- a/src/formio/components/FileField.stories.mdx
+++ b/src/formio/components/FileField.stories.mdx
@@ -1,0 +1,128 @@
+import {ArgsTable, Canvas, Meta, Story} from '@storybook/addon-docs';
+import {getWorker} from 'msw-storybook-addon';
+
+export const worker = getWorker();
+
+import {FileField} from './FileField';
+import {
+  UPLOAD_URL,
+  mockFileUploadDelete,
+  mockFileUploadErrorPost,
+  mockFileUploadPost,
+} from './FileUpload.mocks';
+import {FormDecorator, FormioComponent} from './story-util';
+
+export const FILEPATTERN = [
+  'application/pdf',
+  'application/msword',
+  'application/vnd.rar',
+  'application/zip',
+  'image/heic',
+  'image/jpeg',
+  'image/png',
+  'text/csv',
+  'text/plain',
+];
+
+export const ERRORS = ['File is too big', 'File is of the wrong type'];
+
+<Meta
+  title="Form.io components/FileUpload"
+  component={FileField}
+  argTypes={{
+    filePattern: {
+      options: FILEPATTERN,
+      control: {type: 'check'},
+    },
+  }}
+  decorators={[FormDecorator]}
+  parameters={{
+    controls: {sort: 'requiredFirst'},
+    msw: {
+      handlers: [mockFileUploadPost, mockFileUploadDelete],
+    },
+  }}
+/>
+
+# File Upload
+
+**Note** - file uploads on the Docs page will always error due to the way
+the network mocking library works. Use the canvas instead.
+
+export const Template = ({
+  id,
+  label,
+  required,
+  description,
+  filePattern=[],
+  multiple,
+  backendErrors=[],
+}) => {
+  if (backendErrors.length) {
+    worker.use(mockFileUploadErrorPost(backendErrors));
+  }
+  return (
+    <FormioComponent
+      component={{
+        type: 'file',
+        id: 'ef2k87',
+        key: id,
+        label,
+        input: true,
+        description,
+        validate: {
+          required,
+        },
+        multiple,
+        filePattern: filePattern.join(','),
+        storage: 'url',
+        url: UPLOAD_URL,
+      }}
+    />
+  );
+}
+
+## Happy flow
+
+<Canvas>
+  <Story
+    name="Happy flow"
+    args={{
+      id: 'fileUpload1',
+      label: 'Attachment',
+      description: 'Klik om een bestand te selecteren en in te zenden',
+      required: false,
+      multiple: false,
+    }}
+  >
+    {Template.bind({})}
+  </Story>
+</Canvas>
+
+<ArgsTable story="Happy flow" />
+
+## Error flow
+
+The backend performs a number of checks on file uploads and may return validation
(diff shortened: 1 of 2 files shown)


src/formio/components/FileField.stories.mdx
```diff
@@ -0,0 +1,124 @@
+import {Canvas, Meta, Props, Story} from '@storybook/addon-docs';
```
sergei-maertens, line 1: please use `ArgsTable` instead of `Props`

src/formio/components/FileField.stories.mdx
```diff
@@ -0,0 +1,124 @@
...
+
+import {FileField} from './FileField';
+import {FormDecorator, FormioComponent} from './story-util';
+import {
+  BASE_URL,
+  ENDPOINT,
+  mockFileUploadError,
+  mockFileUploadPost,
+} from '/src/mocks/FileUploadMocks';
```
sergei-maertens, line 10: please use 'absolute' imports relative to the `src` folder: `mocks/FileUploadMocks`

related: I prefer keeping the mocks and components closely together, so in this case `./FileUpload.mocks.js` is probably a better location.

#369 also adds some more general structure (exposes `BASE_URL` in `story-utils/mocks.js`) so let's wait until that is merged and then rebase this branch on that.

src/mocks/FileUploadMocks.js
```diff
@@ -0,0 +1,17 @@
+import {rest} from 'msw';
+
+export const BASE_URL = 'http://localhost:8000/api/v2/';
+export const ENDPOINT = 'formio/fileupload/';
```
sergei-maertens, line 4: maybe better to just define `const UPLOAD_URL = ${BASE_URL}formio/fileupload`?

src/mocks/FileUploadMocks.js
```diff
@@ -0,0 +1,17 @@
...
+  const body = {
+    url: `${BASE_URL}${ENDPOINT}/test`,
+    name: 'test',
+    size: 'test',
+  };
+  return res(ctx.status(200), ctx.json(body));
+});
+
+export const mockFileUploadError = rest.post(`${BASE_URL}${ENDPOINT}`, (req, res, ctx) => {
+  return res(ctx.status(400));
```
sergei-maertens, line 16: please add an example error in the response body - I remember this requires some cruft so you can trigger an invalid upload on a real form and then look at the response data for an example via the browser debugger.

src/mocks/FileUploadMocks.js
```diff
@@ -0,0 +1,17 @@
+import {rest} from 'msw';
+
+export const BASE_URL = 'http://localhost:8000/api/v2/';
+export const ENDPOINT = 'formio/fileupload/';
+
+export const mockFileUploadPost = rest.post(`${BASE_URL}${ENDPOINT}`, (req, res, ctx) => {
+  const body = {
+    url: `${BASE_URL}${ENDPOINT}/test`,
+    name: 'test',
+    size: 'test',
```
sergei-maertens, line 10: this should be an integer - does not conform to the API docs: http://localhost:8000/api/v2/docs/#tag/formio/operation/formio_fileupload_create

src/formio/components/FileField.stories.mdx
```diff
@@ -0,0 +1,124 @@
...
+
+<Props story="Happy flow" />
+
+## Error flow
+
+<Canvas>
+  <Story
+    name="Error flow"
+    args={{
+      componentUrl: `${BASE_URL}${ENDPOINT}`,
```
sergei-maertens, line 108: this should not be an editable arg - probably we can just bake this directly into the `Template` instead of using an arg for it

src/formio/components/FileField.stories.mdx
```diff
@@ -0,0 +1,124 @@
...
+      description: 'Klik om een bestand te selecteren en in te zenden',
+      id: 'option',
```
sergei-maertens, line 110: this maps to the component `key` of the formio component, so I'd use something more appropriate like `fileUpload` here

src/formio/components/FileField.stories.mdx
```diff
@@ -0,0 +1,124 @@
...
+  'txt',
+  'doc',
+  'docx',
+  'open_office',
+  'zip',
+  'rar',
+  'tar',
+  'msg',
+  'dwg',
+];
```
sergei-maertens, line 30: the component itself requires mime types, no? instead of extensions

src/formio/components/FileField.stories.mdx
```diff
@@ -0,0 +1,124 @@
...
+  ...args
+}) => (
+  <FormioComponent
+    component={{
+      key: 'bestand',
+      type: 'file',
+      input: true,
+      label,
+      description,
+      storage: setStorage,
```
sergei-maertens, line 71: I think this should be hardcoded to `url` and not an arg

sergei-maertens (changes requested):

src/formio/components/FileField.stories.mdx
pi-sigma: The component itself takes mime types, not extensions, but the user who creates forms selects mime types only indirectly, viz. by selecting file extensions. If the purpose is purely illustrative, then it doesn't make sense to list mime types.

sergei-maertens (approved):

jiromaykin:
@sergei-maertens Nice job! I'll take some time to try and understand this one :-D

PR #328, opened 2022-12-19
:wheelchair: [open-formulieren/open-forms#2360] set h1 headings

jiromaykin:
Set semantically ordered headings, fixes open-formulieren/open-forms#2360


sergei-maertens (commented):
I'll take a look at the style changes this introduces

PR #329, opened 2022-12-22
Fixed some storybook bugs

### LINKED ISSUE #325: Storybook controls/theme switcher is broken in formio components
Until currency it seems fine, but from that one the theme selector suddenly is no longer available and styling seems to break too.

sergei-maertens:
Fixes #325 

Took some time to find the root case: the `argTypes.customClass.control.labels` with an empty string is not happily accepted by storybook.


jiromaykin (approved):
Approved debugging.

PR #330, opened 2022-12-22
:wheelchair: nr.2267 Progress indicator semantics

jiromaykin:
Extra fixes for https://github.com/open-formulieren/open-forms/issues/2267

### DIFF
--- a/src/components/ProgressIndicator/ProgressIndicatorDisplay.js
+++ b/src/components/ProgressIndicator/ProgressIndicatorDisplay.js
@@ -115,70 +115,72 @@ const ProgressIndicatorDisplay = ({
     defaultMessage: 'Progress step indicator icon',
   });
   const toggleStepsTitle = intl.formatMessage({
-    description: 'Active step toggle titles',
-    defaultMessage: 'Show/hide steps',
+    description: 'Active step title',
+    defaultMessage: 'Current step',
   });
 
   return (
     <Card blockClassName="progress-indicator" modifiers={expanded ? [] : ['mobile-collapsed']}>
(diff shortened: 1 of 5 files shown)


src/components/ProgressIndicator/ProgressIndicatorDisplay.js
```diff
@@ -115,70 +115,72 @@ const ProgressIndicatorDisplay = ({
...
-        </span>
-      </button>
+          <FAIcon
+            icon={expanded ? 'chevron-up' : 'chevron-down'}
+            modifiers={['normal']}
+            aria-label={ariaIconLabel}
+          />
+          <span
+            className={getBEMClassName('progress-indicator__active-step')}
+            aria-label={toggleStepsTitle + ': ' + activeStepTitle}
```
sergei-maertens, line 138: this interpolation needs to be handled at the translation level rather than string concatenating - future languages may have a different sentence structure.

I'll pick this up myself to wrap up the PR.

jiromaykin: Thanks, I am very interested to see - could this be a right-to-left language issue?

sergei-maertens (approved):

src/components/ProgressIndicator/ProgressIndicatorDisplay.js
sergei-maertens: not necessarily, but there is potential that a language exists where a different structure is more natural.

Translators also have an option to then do something like `{activeStepTitle} (huidige stap)` instead of `Huidige stap: {activeStepTitle}`

PR #336, opened 2022-12-30
[OF#2377] Make links hover colours more consistent

SilviaAmAm:
Fixes open-formulieren/open-forms#2377

TODO/check:

- [ ] Update submodule to latest design tokens
- [ ] Update package.json min version of design tokens
- [ ] Can we get rid of of-anchor completely and instead use `.utrecht-link--openforms`?


src/scss/components/_button.scss
```diff
@@ -222,4 +208,46 @@ $button-padding-h: $grid-margin-2;
       color: var(--of-button-danger-bg);
     }
   }
+
+  @include modifier('anchor') {
```
sergei-maertens, line 212: I can't easily tell from the diff - but why was this broken out of the `button-theme` mixin? Is it possible to add a `@content` to that button and include additional styles here as part of the mixin call then?

e.g.

```scss
@include button-theme('anchor', ...) {
  additional-property: value;
}
```

src/scss/mixins/_anchor.scss
```diff
@@ -0,0 +1,10 @@
+@mixin of-anchor() {
+  color: var(--of-link-color);
+  cursor: pointer;
+  text-decoration: var(--utrecht-link-text-decoration, underline);
+
+  &:hover {
+    color: var(--of-link-hover-color);
+    text-decoration: var(--utrecht-link-hover-text-decoration, none);
+  }
+}
```
sergei-maertens, line 10: we actually completely want to avoid this, as it introduces OF-specific design tokens and doesn't allow us to lift on existing design tokens. See https://nl-design-system.github.io/utrecht/storybook/?path=/docs/css-link-design-tokens--page for a reference of existing design tokens.

In particular, we should move away from `--of-link-color` and `--of-link-hover-color` in favour of `--utrecht-link-color` and `--utrecht-link-hover-color`. This implies:

* update the backend links to have the `.utrecht-link` class name
* update the backend documentation to include the `utrecht.link.color` design token example
* add a datamigration in the backend to update the config design tokens, mapping from of-specific value to utrecht component
* update the e-mail/PDF templates to use the utrecht design token rather than our own


For the WYSIWYG fields, I would avoid the mixin structure in this PR, instead rather re-use the available styles with some sass features:

```scss
.#{prefix(body)} {
  @include modifier(wysiwyg) {
    @include wysiwyg;

    a {
      @extend .utrecht-link;
    }
  }
}
```
(see the [`@extend`](https://sass-lang.com/documentation/at-rules/extend) docs)

This will also require to add the `@import '@utrecht/components/dist/link/css/index';` into the `_anchor.scss` component (and then it can be removed from `nl-design-system-community.scss`)

sergei-maertens (changes requested):

src/scss/components/_button.scss
SilviaAmAm: When we added the "image" button we said that we wanted to move out of the "button-theme". So, since I was modifying it, I thought I'd refactor it so that we have fewer `!important`

sergei-maertens: hrm, I think it'd be nicer to have a single refactor PR to tackle _all_ of them, in this in-between state I think it increases confusion for Jiro for example who's working on accessibility things related to buttons, amongst other :thinking:

SilviaAmAm: Ok, I will make a PR with the refactor separately from this PR!

src/scss/mixins/_anchor.scss
SilviaAmAm: Backend PR: https://github.com/open-formulieren/open-forms/pull/2529

src/scss/components/_button.scss
SilviaAmAm: https://github.com/open-formulieren/open-forms-sdk/issues/339

sergei-maertens (commented):
I'm not entiry sure that all this is needed to get where we want to get, but I also can't put my finger on it right now. I'll take a look tomorrow morning at this PR and see if I can make things easier to follow!

PR #338, opened 2023-01-06
:memo: - Improve documentation and readability of FormStep.js

svenvandescheur:


src/components/FormStep.js
```diff
@@ -264,47 +341,82 @@ const FormStep = ({form, submission, onLogicChecked, onStepSubmitted, onLogout})
...
+   *
+   * During evaluation, the following actions may be dispatched:
+   *
+   *   - `LOGIC_CHECK_INTERRUPTED` When there's no point/change in state to be expected by the logic
+   *       check
+   *   - `BLOCK_SUBMISSION` When the the logic is checking and the form should be disabled for
+   *       submission.
+   *   - `LOGIC_CHECK_DONE` When the logic check is done and the form should be enabled for
+   *       submission.
+   *   - `LOGIC_CHECK_INTERRUPTED` When an error occurred during the logic check.
```
sergei-maertens, line 400: you are tying the dispatch action states with the state mutations that happen in the reducer here, this will definitely go out of sync if/when we update internal state aspects in the reducers.

The action types are intended to be sufficiently descriptive to have a broad idea of what is happening/being affected, the detailed documentation belongs in the reducer.

These descriptions also don't capture all nuances, e.g. `LOGIC_CHECK_INTERRUPTED` is not because some error happened, it may just have been part of the debounce mechanism. You probably mean that a technical exception was raised to communicate that, but that's just an implementation detail that is not relevant for the functional side of things.

src/components/FormStep.js
```diff
@@ -339,16 +451,20 @@ const FormStep = ({form, submission, onLogicChecked, onStepSubmitted, onLogout})
 
     // now the actual checking *can* be aborted, which results in an exception being thrown.
     try {
-      // call the backend to do the check
+      // `checkAbortedLogicCheck` throws error if checko should be aborted, skipping `doLogicCheck`.
       checkAbortedLogicCheck(controller.signal);
-      // update our view of the data, which may have been changed by now because of user input
+
```
sergei-maertens, line 456: a bunch of these newlines break up the code more, and hurt readability in the sense that the entire function no longer fits in the editor window to try to grok it. Again - please refrain from making code style updates for the sake of updating the code style.

src/components/FormStep.js
```diff
@@ -399,12 +515,25 @@ const FormStep = ({form, submission, onLogicChecked, onStepSubmitted, onLogout})
...
       if (e.name !== 'AbortError') {
         throw e; // re-throw on unexpected errors
       }
     }
   };
 
+  /**
+   * A form has been submitted.
+   * @see {onReactSubmit} We wrap the submit so that we control our own submit buttonLOGIC_CHECK_DONE
```
sergei-maertens, line 527: copy paste problem with the LOGIC_CHECK_DONE here?

src/components/FormStep.js
```diff
@@ -399,12 +515,25 @@ const FormStep = ({form, submission, onLogicChecked, onStepSubmitted, onLogout})
...
+   *
+   * During evaluation, the following actions may be dispatched:
+   *
+   *   - `NAVIGATE` Immediately (if submission is available).
+   *   - `ERROR` When an error occurred while loading the form step configuration.
```
sergei-maertens, line 532: this really doesn't make things clearer for me looking just at the docstring - I actually have no idea at the moment what it means unless I go looking at the code anyway. Should probably be inline comments.

src/components/FormStep.js
```diff
@@ -481,16 +624,39 @@ const FormStep = ({form, submission, onLogicChecked, onStepSubmitted, onLogout})
     }
   };
 
+  /**
+   * Gets called when the user confirms saving the form.
```
sergei-maertens, line 628: adds very little value over the function name

src/components/FormStep.js
```diff
@@ -481,16 +624,39 @@ const FormStep = ({form, submission, onLogicChecked, onStepSubmitted, onLogout})
...
-    const response = await submitStepData(submissionStep.url, {...getCurrentFormData()});
-    return response;
+    return await submitStepData(submissionStep.url, {...getCurrentFormData()});
   };
 
+  /**
+   * Get called when the user presses the save button.
+   * @see {onSaveConfirm} Opens a modal, `onSaveConfirm` after user confirms.
+   *
+   * During evaluation, the following actions may be dispatched:
```
sergei-maertens, line 639: seems to imply that sometimes it's dispatched, sometimes not, but it's always dispatched. & see earlier remarks about this documentation style for action dispatching -> I'd leave this out of the docblock.

src/components/FormStep.js
```diff
@@ -481,16 +624,39 @@ const FormStep = ({form, submission, onLogicChecked, onStepSubmitted, onLogout})
...
+   * @return {Promise}
+   */
   const onFormSave = async event => {
     event.preventDefault();
     dispatch({type: 'TOGGLE_FORM_SAVE_MODAL', payload: {open: true}});
   };
 
+  /**
+   * Gets called when the user presses the previous page button.
+   * @param {PointerEvent} event
```
sergei-maertens, line 654:
```suggestion
   * Handler to navigate back to the previous step or page
   * @param {PointerEvent} event
```

src/components/FormStep.js
```diff
@@ -504,6 +670,19 @@ const FormStep = ({form, submission, onLogicChecked, onStepSubmitted, onLogout})
...
+   * loop.
+   * @see {@link Form.io documentation} https://help.form.io/developers/form-renderer#form-events
+   *
+   *  - We cannot filter 'blank' values to prevent Formio validation from running, as Formio will
+   *    use the default values in that case which have been explicitly unset.
+   *  - In the situation that we have invalid backend data (loading a submission with a required
+   *    field with default value that was cleared, for example), we _need_ to see the validation
+   *    errors since the data is not valid.
+   *  - For the initial, empty form load, no validation errors are displayed as there is no
+   *    respective backend data.
```
sergei-maertens, line 684: this really belongs in the context _close_ to where it's relevant, so please put it back as inline comment.

src/components/FormStep.js
svenvandescheur: I'd much rather leave them in as I'm going to need them in the near future and I expect code to be largely rewritten.

src/components/FormStep.js
```diff
@@ -228,19 +301,23 @@ const FormStep = ({form, submission, onLogicChecked, onStepSubmitted, onLogout})
...
-  // state updates on unmounted components.
+  /**
+   * Fetches the form step data from the backend.
+   * @todo something is causing the FormStep.js to render multiple times, leading to state updates
+   *   on unmounted components.
+   *
+   * During evaluation, the following actions may be dispatched:
+   *
+   *   - `ERROR` When an error occurred while loading the form step configuration.
+   *   - `STEP_LOADED` When the form step is fetched fom the  backend.
```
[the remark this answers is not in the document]
svenvandescheur: Please see my previous note regarding future refactors, as well the functions/methods are quite chunky and (too) complex so IMO (during refactoring) they are adding value.

src/components/FormStep.js
svenvandescheur: Fro me "action types are intended to be sufficiently descriptive" and "don't capture all nuances" are conflicting and a clear indication that they do need documentation. I understand the risk of them going out of sync and I'm opposed to remove them after refactor given that the code is sufficiently self descriptive, which is, IMO, at the current stage not the case.

svenvandescheur: Discussed: keep for now, may be removed after refactor.

src/components/FormStep.js
svenvandescheur: Discussed: keep for now, may be removed after refactor.

PR #340, opened 2023-01-10
Ensure utrecht-link styles/states are applied everywhere

sergei-maertens:
Closes #336 
Related to open-formulieren/open-forms#2377
Depends on https://github.com/open-formulieren/design-tokens/pull/14/files

**Changes**

* `openforms-anchor` class + variants is gone, replaced with `.utrecht-link` and variants thereof
* wysywig styles that can contain `a` tags without classname -> added styling extending utrecht-link component
* button--anchor variants -> best effort

Backend PR tackling similar issues: https://github.com/open-formulieren/open-forms/pull/2536


src/components/Button.js
```diff
@@ -8,7 +8,7 @@ import React from 'react';
 
 import {getBEMClassName} from 'utils';
 
-export const VARIANTS = ['', 'primary', 'secondary', 'anchor', 'danger', 'image', 'icon-only'];
+export const VARIANTS = ['', 'primary', 'secondary', 'anchor', 'danger', 'icon-only'];
```
sergei-maertens, line 11: `image` variant was cleaned up an not used anymore as far as I can tell -> one less thing to support

src/scss/components/_button.scss
```diff
@@ -222,4 +203,26 @@ $button-padding-h: $grid-margin-2;
       color: var(--of-button-danger-bg);
     }
   }
+
+  // anchor variants, these map to utrecht components:
+  // * UtrechtButton (the default)
+  // * UtrechtButtonLink (`a` component)
+  // * UtrechtLinkButton (`button` component)
+  @include bem.modifier('anchor') {
+    @include button-theme(
```
sergei-maertens, line 212: currently the best way to handle all the accessibility stuff, we'll revisit later

src/scss/nl-design-system-community.scss
```diff
@@ -3,7 +3,6 @@
 @import '@utrecht/components/dist/form-field-description/css/index.css';
 @import '@utrecht/components/dist/form-label/css/index.css';
 @import '@utrecht/components/dist/img/css/index.css';
-@import '@utrecht/components/dist/link/css/index.css';
```
sergei-maertens, line 6: promoted into our `_anchor.scss` file so it's included in our build artifact instead of only used for SB

src/scss/components/_button.scss
```diff
@@ -222,4 +198,32 @@ $button-padding-h: $grid-margin-2;
...
+      $force-colors: true // there will not be an anchor variant
+    );
+
+    // the OPPOSITE of what we normally do
+    text-decoration: var(--utrecht-link-hover-text-decoration, none);
+    &:hover {
+      --of-button-anchor-fg: var(--utrecht-link-hover-color);
+      text-decoration: var(--utrecht-link-text-decoration, underline);
+    }
+  }
```
SilviaAmAm, line 228: Maybe I am doing something wrong, but at the moment when I hover on the "previous page" button, i don't get the right hover colour, I get `--_utrecht-button-hover-color`

SilviaAmAm (changes requested):
I think that the link in the privacy policy checkbox at the end is missing the refactor?

src/scss/components/_button.scss
sergei-maertens: I'll re-check this one! The buttons is definitely the most nasty part.

sergei-maertens: in the SDK this seems to work for me when I specify the `--utrecht-link-hover-color: fuchsia;` design token :thinking:

sergei-maertens: checked via video call and this seems to be okay

PR #344, opened 2023-01-17
[#2539] fix mime type validation

pi-sigma:
Fixes open-formulieren/open-forms#2539 (partly).

Added support for mime type validation on the frontend where the default validation performed by formio doesn't work (e.g. the type of msg files is not recognized by browsers).

### DIFF
--- a/src/formio/components/FileField.js
+++ b/src/formio/components/FileField.js
@@ -189,6 +189,18 @@ class FileField extends Formio.Components.components.file {
     super.upload(files);
   }
 
+  validatePattern(file, val) {
+    // file type not recognized; get file extension and specify mime type manually
+    if (file.type === '') {
+      let ext = file.name.split('.').pop();
+      if (ext === 'msg') {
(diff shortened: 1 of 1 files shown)


src/formio/components/FileField.js
```diff
@@ -189,6 +189,18 @@ class FileField extends Formio.Components.components.file {
     super.upload(files);
   }
 
+  validatePattern(file, val) {
+    // file type not recognized; get file extension and specify mime type manually
+    if (file.type === '') {
+      let ext = file.name.split('.').pop();
+      if (ext === 'msg') {
+        return val.includes('application/vnd.ms-outlook');
+      }
```
sergei-maertens, line 198: I wouldn't shortcut validation here, since wildcards can be supported in formio. Instead, I'd fixup the actual file information:

```
file.type = 'application/vnd.ms-outlook';
```

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #346, opened 2023-01-24
:pencil: Update the contributing guidelines and technical vision

sergei-maertens:


svenvandescheur (approved):
Seems like a good start to me. I only wonder if we should point to this file in the documentation (as I think developer will primary look for documentation there).

sergei-maertens:
> Seems like a good start to me. I only wonder if we should point to this file in the documentation (as I think developer will primary look for documentation there).

Yeah, I'll update the main documentation on RTD - it already points to storybook, but can be made more clear.

PR #347, opened 2023-01-27
appointment step experimental

annashamray:
fixes https://github.com/open-formulieren/open-forms/issues/2471

done:
* added appointment page without submission
* the default form redirects to the appointment page if `appointmentEnabled` is true

Submission would be tackled in the next PR


src/components/ProgressIndicator/ProgressIndicatorDisplay.js
```diff
@@ -173,6 +175,18 @@ const ProgressIndicatorDisplay = ({
               isCurrent={step.isCurrent}
             />
           ))}
+          {showAppointment && (
+            <ProgressItem completed={isAppointment}>
+              <LinkOrSpan
+                to={'/appointment'}
+                useLink={areApplicableStepsCompleted}
```
sergei-maertens, line 182: since there are no applicableSteps at all, I think this should be either always `true` or depend on the state of the "inloggen"/"start" step, however I'm not sure we can tackle this right now due to the rest of the flow not being baked in yet and functional decisions needing to be made about the inloggen/start step.

src/components/ProgressIndicator/ProgressIndicatorDisplay.js
```diff
@@ -173,6 +175,18 @@ const ProgressIndicatorDisplay = ({
...
+                isActive={isSummary}
```
sergei-maertens, line 183: we probably need to refactor the progress indicator component more, since the "appointment" step checkmark is now tied to the summary page which is too late.

It would maybe make sense to use a context of relevant items/steps here and have that state managed by the parent container so that this component is a bit more general-purpose.

src/components/appointments/AppointmentFields.js
```diff
@@ -0,0 +1,189 @@
...
+        <Loader />
+      ) : (
+        <select className={getBEMClassName('select')} value={value} onChange={onChange}>
+          <option value={''}></option>
+          {items.map(item => (
+            <option value={item.identifier} key={item.identifier}>
+              {item.name}
+            </option>
+          ))}
+        </select>
```
sergei-maertens, line 39: there's a lot of low-level HTML repetition for these dropdowns, I'd drop in react-select instead. It can also handle async stuff out of the box.

src/components/appointments/AppointmentFields.js
```diff
@@ -0,0 +1,189 @@
...
+      </Label>
+
+      {loading ? (
+        <Loader />
+      ) : (
+        <select className={getBEMClassName('select')} value={value} onChange={onChange}>
+          <option value={''}></option>
+          {items.map(item => (
+            <option value={item.date} key={item.date}>
+              {getFormattedDateString(intl, item.date)}
```
sergei-maertens, line 96: we prefer using the `react-intl` `FormattedDate` component - the `useIntl` imperative API is an escape hatch for when we're outside of React context.

src/components/appointments/AppointmentFields.js
```diff
@@ -0,0 +1,189 @@
...
+      )}
+    </div>
+  );
+};
+
+const AppointmentFields = () => {
+  const [product, setProduct] = useState('');
+  const [location, setLocation] = useState('');
+  const [date, setDate] = useState('');
+  const [time, setTime] = useState('');
```
sergei-maertens, line 143: are you planning to check out Formik for form state management in a separate PR, or do we want this in here?

src/components/appointments/AppointmentForm.js
```diff
@@ -0,0 +1,183 @@
...
+    description: 'Summary page title',
+    defaultMessage: 'Check and confirm',
+  });
+  useTitle(pageTitle);
+
+  const [privacy, setPrivacy] = useState({
+    requiresPrivacyConsent: true,
+    privacyLabel: '',
+    policyAccepted: false,
+  });
```
sergei-maertens, line 75: seeing the repetition with existing code, this probably means we need a privacy row/component to encapsulate the state/details inside.

sergei-maertens (changes requested):

src/components/ProgressIndicator/ProgressIndicatorDisplay.js
annashamray: Should we make it in the separate PR? With some regression testing?  For now I wanted to add the appointment flow with minimal changes to the existing components. For now only `ProgressIndicatorDisplay` is reused, but the change is quite small.

src/components/appointments/AppointmentFields.js
annashamray: To be honest I didn't managed react-select to work with async requests. But for now all low-level code is encapsulated in `AppointmentField` component and it's more DRYer

src/components/appointments/AppointmentFields.js
annashamray: I tried to use `<FormattedDate/>` and `<FormattedTime/>` in option labels, but the values were not shown in the dropdowns. So I left it as it is

src/components/appointments/AppointmentFields.js
annashamray: refactored with formik

src/components/appointments/AppointmentForm.js
annashamray: Should we make it in the next PRs? It's the same as with Progress indicator - I want to make changes which affect the existing process in separate PRs with additional care

src/components/ProgressIndicator/ProgressIndicatorDisplay.js
sergei-maertens: yeah, we have some existing planned refactoring for the PI component anyway since the markup is... sketchy :grimacing: good to leave it until later.

package.json
```diff
@@ -82,6 +83,16 @@
...
+    "parserOptions": {
+      "babelOptions": {
+        "presets": [
+          [
+            "babel-preset-react-app",
+            false
+          ]
+        ]
+      }
+    },
```
sergei-maertens, line 95: why do we need this?

src/components/appointments/AppointmentStep.js
```diff
@@ -0,0 +1,147 @@
...
+  const [field, meta] = useField(props);
+  // load options from appointments api
+  const {loading, value: items} = useAsync(async () => {
+    if (urlQuery) {
+      const notEmpty = Object.values(urlQuery).every(val => !!val);
+      return await requestOrEmpty(url, urlQuery, notEmpty);
+    } else {
+      return await get(url);
+    }
+  }, Object.values(urlQuery));
```
sergei-maertens, line 42: Objects in Javascript don't have inherent ordering, so relying on `Object.values` is not stable. The exact query params may still be the same, but the order shifted, which will re-trigger network requests.

I also find this makes the code hard to read in terms of which hook dependencies are specified. Maybe an Array of key-value pairs is a better datastructure?

src/components/appointments/AppointmentStep.js
```diff
@@ -0,0 +1,147 @@
...
+const AppointmentStep = ({onSubmit}) => {
+  const {baseUrl} = useContext(ConfigContext);
+  const intl = useIntl();
+
+  return (
+    <Card
+      title={
+        <FormattedMessage
+          description="Experiment appointment step"
+          defaultMessage="Create experiment appointment"
```
sergei-maertens, line 84:
```suggestion
          defaultMessage="Appointment details"
```

src/components/appointments/AppointmentForm.stories.mdx
```diff
@@ -0,0 +1,103 @@
...
+      </LayoutRow>
+    </Layout>
+  );
+};
+
+<Meta
+  title="Composites/Appointment form"
+  component={AppointmentForm}
+  parameters={{
+    mockData: [
```
sergei-maertens, line 36: please rebase on `main` and set up the appropriate msw mocks, you can find some documentation here: https://open-formulieren.github.io/open-forms-sdk/?path=/story/developers-documentation--page#mocking-in-stories and some examples in https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/LanguageSelection/mocks.js and https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/LanguageSelection/LanguageSelection.stories.mdx?plain=1#L61

sergei-maertens (changes requested):
The storybook stories are very broken for me locally :grimacing: 

image

Seems related to the MSW mocks that are missing

sergei-maertens (approved):

Issue #322, opened 2022-12-15
Add focus-styles and design tokens for login-button-logo component


sergei-maertens:
Depends on #313

This will also require some backend changes to include meta-information about the logo (whether it's dark or light).

**Tasks**

- [x] Add design tokens `of.login-button-logo.dark.focus.border-color` and `of.login-button-logo.dark.focus.border-color`
- [ ] Implement CSS classname modifiers `openforms-login-button-logo--{dark|light}` depending on backend information (TODO!)
- [x] Add prop `appearance` with possible values `dark` and `light` to `LoginButtonIcon` component, which configures the modifier of the component
- [x] Add CSS styles that connect the classnames/modifiers to the appropriate design token

PR #349, opened 2023-01-27
:wheelchair: [#322] Add focus-styles for login-button-logo

jiromaykin:
Fixes [#322] ,
in combination with the Design-Tokens branch [open-formulieren/design-tokens#feature/322-login-button-logo-focus-modifiers]](https://github.com/open-formulieren/design-tokens/pull/15)


src/components/LoginButton/LoginButton.stories.mdx
```diff
@@ -22,6 +22,12 @@ import LoginButtonIcon from './LoginButtonIcon';
         type: 'text',
       },
     },
+    appearance: {
+      options: LOGO_APPEARANCES,
+      control: {
+        type: 'radio',
+      },
+    },
```
sergei-maertens, line 30: @jiromaykin I exported the possible values for `appearance` so we can feed that constraint to Storybook - now you can just select the appearance between two radio inputs.

src/components/LoginButton/LoginButtonIcon.js
```diff
@@ -26,6 +29,7 @@ LoginButtonIcon.propTypes = {
     title: PropTypes.string.isRequired,
     imageSrc: PropTypes.string.isRequired,
     href: PropTypes.string,
+    appearance: PropTypes.oneOf(LOGO_APPEARANCES).isRequired,
```
sergei-maertens, line 32: @jiromaykin 
defined the possible values in a constant and used that for a more specific prop type - this matches the possible enum values from the API spec

src/scss/components/_login-button-logo.scss
```diff
@@ -1,17 +1,37 @@
+@use '~microscope-sass/lib/bem';
+
 @import '~microscope-sass/lib/grid';
 @import '~microscope-sass/lib/responsive';
 
 @import '../mixins/prefix';
 
 .#{prefix(login-button-logo)} {
-  display: flex;
+  display: inline-flex;
```
sergei-maertens, line 9: using `inline-flex` removes the need to limit the height/width to an 'arbitrary' pixel value.

src/scss/components/_login-button-logo.scss
```diff
@@ -1,17 +1,37 @@
...
   align-items: center;
 
+  &:focus,
+  &:focus-visible {
+    border-color: var(--_of-login-button-logo-focus-color-border);
+    border-width: 2px;
+    outline: var(--_of-login-button-logo-focus-color-border) solid 2px;
+    outline-offset: 1px;
+    border-radius: var(--of-login-button-focus-border-radius, 1px);
+  }
```
sergei-maertens, line 19: I've moved this into generic component-level styles, using local 'private' design token values. The exact value is filled in using the appropriate modifier.

src/scss/components/_login-button-logo.scss
```diff
@@ -1,17 +1,37 @@
...
   }
 
+  @include bem.modifier('dark') {
+    --_of-login-button-logo-focus-color-border: var(--of-login-button-logo-dark-focus-color-border);
+  }
+  @include bem.modifier('light') {
+    --_of-login-button-logo-focus-color-border: var(
+      --of-login-button-logo-light-focus-color-border
+    );
+  }
```
sergei-maertens, line 33: we prefer using the `bem` library for specific CSS selectors. Also, we specify only the necessary overrides rather than duplicating code.

sergei-maertens (approved):

PR #350, opened 2023-01-28
Use design-token-editor to document design tokens

sergei-maertens:
@Robbert well this seems to work pretty nicely for the first attempt :)


Robbert (approved):
Great stuff!

PR #353, opened 2023-02-01
Document design tokens

sergei-maertens:
This documents the design tokens for the components currently hooked into storybook.


SilviaAmAm (approved):
I notice that the styles of the links in storybook are not right for the other themes than OF? For example, the DH links are not green and in the 'hover' version the line underneath doesn't appear on hover?

sergei-maertens:
> I notice that the styles of the links in storybook are not right for the other themes than OF? For example, the DH links are not green and in the 'hover' version the line underneath doesn't appear on hover?

Could that be because they have their own link component and don't use the `.utrecht-link` component, hence they don't have the design tokens specified for those components?

PR #354, opened 2023-02-06
[OF#2646] Persist privacy policy

SilviaAmAm:
Part of https://github.com/open-formulieren/open-forms/issues/2646

### DIFF
--- a/src/components/Summary/index.js
+++ b/src/components/Summary/index.js
@@ -56,8 +56,10 @@ const loadSummaryData = async submissionUrl => {
   return await get(`${submissionUrl.href}/summary`);
 };
 
-const completeSubmission = async submission => {
-  const response = await post(`${submission.url}/_complete`);
+const completeSubmission = async (submission, privacy) => {
+  const response = await post(`${submission.url}/_complete`, {
+    privacyPolicyAccepted: privacy.policyAccepted,
+  });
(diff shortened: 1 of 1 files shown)


src/components/Summary/index.js
```diff
@@ -27,7 +27,6 @@ const initialState = {
   privacy: {
     requiresPrivacyConsent: true,
     privacyLabel: '',
```
SilviaAmAm, line 29: Could maybe refactor this to no longer use the reducer? There is not much state to manage

src/components/Summary/index.js
```diff
@@ -120,6 +116,20 @@ const Summary = ({
   const onSubmit = async event => {
     event.preventDefault();
     if (refreshedSubmission.submissionAllowed !== SUBMISSION_ALLOWED.yes) return;
+
+    // Update the submission to persist that the privacy policy was accepted
+    try {
+      const response = await patch(submission.url, submission);
```
SilviaAmAm, line 122: This now persists the `privacy_policy_accepted` attribute when the user clicks on 'submit'. It could also happen when the user checks the box, but then we need to block the submit button while that request is happening, make sure that the submission is refreshed in the state, maybe deal with debouncing in case the user clicks many times... So i went for the other option :shrug:

PR #359, opened 2023-02-13
[OF#2686] Add styles for select

SilviaAmAm:
Fixes open-formulieren/open-forms#2686

The style was working in the SDK, but not in embedded forms. This is because this style in the backend was overwriting the style of the select component: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/scss/components/builder/_builder.scss#L78

Related token PR: https://github.com/open-formulieren/design-tokens/pull/16

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   "dependencies": {
     "@formio/protected-eval": "^1.2.1",
     "@fortawesome/fontawesome-free": "^6.1.1",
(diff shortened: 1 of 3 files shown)


SilviaAmAm:
I noticed that if there is a map after the select, it's still overlapping.

sergei-maertens (approved):

PR #360, opened 2023-02-15
[OF#2708] Render zeroes

SilviaAmAm:
Fixes open-formulieren/open-forms#2708
Screenshot from 2023-02-15 13-38-31


src/components/FormStepSummary/ComponentValueDisplay.js
```diff
@@ -207,7 +207,7 @@ const CoSignDisplay = ({component, value}) => {
 const ComponentValueDisplay = ({value, component}) => {
   const {multiple = false, type} = component;
 
-  if (!value) {
+  if (typeof value === 'undefined' || value === null) {
```
sergei-maertens, line 210: `value == null` is more idiomatic and handles `null` and `undefined`

sergei-maertens (approved):

PR #361, opened 2023-02-17
[#2675] make progress indicator sticky]

pi-sigma:
Closes open-formulieren/open-forms#2675

### DIFF
--- a/src/scss/components/_progress-indicator.scss
+++ b/src/scss/components/_progress-indicator.scss
@@ -14,10 +14,10 @@
 .#{prefix('progress-indicator')} {
   @extend .openforms-card; // TODO -> syntax highlighting trips on #{prefix('card')}
 
-  // sticky sidebar
-  position: -webkit-sticky;
-  position: sticky;
-  top: $grid-margin-4;
+  @include bem.modifier('sticky') {
+    position: sticky;
+    top: var(--of-progress-indicator-sticky-spacing);
+  }
 
   @include bem.element('mobile-header') {
     @include body;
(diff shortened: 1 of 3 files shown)


src/scss/components/_progress-indicator.scss
```diff
@@ -14,6 +14,11 @@
 .#{prefix('progress-indicator')} {
   @extend .openforms-card; // TODO -> syntax highlighting trips on #{prefix('card')}
 
+  // sticky sidebar
+  position: -webkit-sticky;
+  position: sticky;
+  top: $grid-margin-4;
+
```
sergei-maertens, line 21: As discussed in the office - this isn't the most appropriate place to write this CSS since it's now always applied and uses hardcoded values.

I took a look at the parent container, but those are also pretty generic, so for this I'd opt to add an additional BEM modifier to opt-in to sticky positioning.

So, add the styles under the modifier `sticky`:

* `position: sticky` (and don't use vendor prefixes, that's handled by the build toolchain)
* define `top: var(--of-progress-indicator-sticky-spacing);` - this sets up using the design token which will define later

Next, ensure this classname is set on the component:

* edit the component `ProgressIndicatorDisplay` in `src/components/ProgressIndicator/ProgressIndicatorDisplay.js` to add a boolean type prop `sticky`, and make the default `true` (this allows other uses to opt-out to sticky behaviour)
* update the modifiers of the `Card` component to include the sticky modifier if the prop asks for it

Finally, let's define the design token for the spacing - this is tied to the theme so we'll set up the value for the default theme.

* In https://github.com/open-formulieren/design-tokens/ create the file `src/components/progress-indicator.tokens.json` (you can inspire the structure from https://github.com/open-formulieren/design-tokens/blob/main/src/components/progress-indicator.mobile.tokens.json)
* Add the token for `of.progress-indicator.sticky.spacing` with a value of `20px`

The design tokens PR will have to be merged first and then we update the dependency in the SDK.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #363, opened 2023-02-22
[#2692] hide input element generated by Formio

pi-sigma:
Fixes open-formulieren/open-forms#2692

* when uploading a file, Formio creates an input element at the bottom of the screen
* the input element should be hidden, but Formio attempts to achieve this with inline styling, which is blocked by CSP
* the patch sets the display property of the generated input element to `none` in the SCSS


src/formio/components/FileField.js
```diff
@@ -189,6 +189,12 @@ class FileField extends Formio.Components.components.file {
     super.upload(files);
   }
 
+  browseFiles(attrs = {}) {
+    // provide class name for hiding element, override style to avoid triggering CSP
+    attrs = {class: 'fileUploadInput', style: ''};
```
sergei-maertens, line 194:
```suggestion
    attrs = {...attrs, class: 'fileUploadInput', style: ''};
```

src/scss/components/_file-upload.scss
```diff
@@ -155,3 +155,8 @@ we don't have strict BEM naming here.
     }
   }
 }
+
+// hide the input element generated by Formio file upload component (GH #2692)
+input[class='fileUploadInput'] {
```
sergei-maertens, line 160:
```suggestion
.fileUploadInput {
```

src/formio/components/FileField.js
sergei-maertens, line 194: please prefix the classname (use the `prefix` utility function) and use dashes in the classname rather than camelCase to be consistent with the rest of the classnames used by OF.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #369, opened 2023-02-24
Document FormStep component in Storybook

### LINKED ISSUE #366: Document FormStep.js component in storybook
Minimal example with mocked API interactions

This will be beneficial to test/develop the alternative renderer.

sergei-maertens:
Closes #366

* Added mocks for form detail, submission detail, submission step detail and submission step logic check endpoints
* Added FormStep.js component to storybook

### DIFF
--- a/src/components/App.stories.mdx
+++ b/src/components/App.stories.mdx
@@ -1,10 +1,11 @@
-import {Meta, Canvas, Story, ArgsTable} from '@storybook/addon-docs';
-import {within, waitFor, waitForElementToBeRemoved} from '@storybook/testing-library';
+import {ArgsTable, Canvas, Meta, Story} from '@storybook/addon-docs';
 import {expect} from '@storybook/jest';
+import {waitFor, waitForElementToBeRemoved, within} from '@storybook/testing-library';
 import {MemoryRouter} from 'react-router-dom';
-import {ConfigContext} from 'Context';
 
-import {BASE_URL, mockLanguageInfoGet, mockLanguageChoicePut} from 'components/LanguageSelection/mocks';
+import {ConfigContext} from 'Context';
+import {mockLanguageChoicePut, mockLanguageInfoGet} from 'components/LanguageSelection/mocks';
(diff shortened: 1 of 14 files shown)


src/components/FormStep/FormStep.stories.mdx
```diff
@@ -0,0 +1,94 @@
...
+          }
+        ]
+      },
+      routerArgs: {
+        route: '/stap/:step',
+        initialEntries: ['/stap/step-1'],
+      },
+      form: getForm(),
+      submission: getSubmission(),
+      showDebug: false,
```
SilviaAmAm, line 85: When interacting with the component in storybook, the controls dont seem to do much if one changes the configuration of the form/the submission. Maybe they shouldn't be in the controls? The only prop that changes something seems to be the DEBUG

src/components/App.stories.mdx
```diff
@@ -133,7 +123,7 @@ If translations are enabled on the form, a `LanguageSelector` is rendered:
     }}
     args={{
       'form.translationEnabled': true,
-      baseUrl: process.env.REACT_APP_BASE_API_URL || BASE_URL,
+      baseUrl: BASE_URL,
```
SilviaAmAm, line 126: If I change the base URL in the controls, it breaks the component. Maybe it shouldn't be a control?
Screenshot from 2023-02-24 15-01-37

SilviaAmAm (changes requested):

src/components/App.stories.mdx
sergei-maertens: yeah this one should be removed, since the mocks don't update dynamically anyway :thinking: 

possibly we find a way in the future, but for now setting the envvar is likely the easiest.

src/components/FormStep/FormStep.stories.mdx
sergei-maertens: `form` and `submission` are now disabled controls and I've added a workaround to get the component to re-render when the formio configuration is changed.

SilviaAmAm (approved):

PR #371, opened 2023-02-28
Use MSW mocks in (Jest) tests

### LINKED ISSUE #368: Replace mocks in tests with MSW mocks
The fragile mocks need to be replaced with actual semantic mocks that capture the behaviour of the API.

Once #367 is merged, we can use MSW for this.

Potential useful tutorial: https://www.wwt.com/article/using-mock-service-worker-to-improve-jest-unit-tests

sergei-maertens:
Closes #368

Rather than mocking the API module, we can use MSW to perform mocking at the network layer.


src/api-mocks/forms.js
```diff
@@ -0,0 +1,68 @@
...
+    formDetail = produce(formDetail, draft => {
+      draft.url = req.url.href;
+      draft.uuid = uuid;
+      for (const step of draft.steps) {
+        step.url = `${draft.url}/steps/${step.uuid}`;
+      }
+    });
+    if (once) {
+      res = res.once;
+    }
```
sergei-maertens, line 66: would be nicer to be able to do this as `rest.get(...).once()` but alas

src/components/FormStep/FormStep.stories.mdx
```diff
@@ -1,13 +1,17 @@
-import {Meta, Canvas, Story, ArgsTable} from '@storybook/addon-docs';
-import {getWorker} from 'msw-storybook-addon';
+import {ArgsTable, Canvas, Meta, Story} from '@storybook/addon-docs';
```
sergei-maertens, line 1: sorry for all the formatting changes because of prettier :(

src/sdk.js
```diff
@@ -104,7 +104,7 @@ class OpenForm {
     this.render();
   }
 
-  async render() {
+  render() {
```
sergei-maertens, line 107: IMO this gave the impression that the rendering is awaitable, which it is not (but it doesn't hurt either)

src/sdk.spec.js
```diff
@@ -1,86 +1,43 @@
...
       languageSelectorTarget: target,
       lang: 'nl',
     });
 
-    try {
-      await act(async () => await form.init());
-    } catch (e) {
-      throw e; // should not error
-    }
+    await act(async () => await form.init());
```
sergei-maertens, line 37: tried out throwing exceptions in various places and they all seem to bubble up and fail/error the test itself, so I don't think this original `try` wrapping was necessary, unless I missed something?

src/sdk.spec.js
```diff
@@ -1,86 +1,43 @@
...
 
+    // wait for the loader to be removed when all network requests have completed
+    await waitForElementToBeRemoved(() => within(formRoot).getByRole('status'));
     expect(target).not.toBeEmptyDOMElement();
```
sergei-maertens, line 41: the new way of mocking actually exposed that our tests weren't handling the async behaviour properly - since we now have actual IO and non-instance promise resolution, the tests were trying to assert while the loader was still visible.

This is resolved by given the loader an accessible role and querying by that role to wait until the loader is gone and the actual form is visible.

Probably this could also be "wait for form title to be visible", I'm not married to this approach :thinking:

src/sdk.spec.js
```diff
@@ -91,70 +48,69 @@ describe('OpenForm', () => {
...
+      await waitForElementToBeRemoved(() => within(formRoot).getByRole('status'));
+    });
+
+    expect(within(formRoot).getAllByText('Nederlandse versie').length).toBeGreaterThan(0);
+
+    await act(async () => {
+      await form.onLanguageChangeDone('en');
+    });
+
+    expect(within(formRoot).getAllByText('English version').length).toBeGreaterThan(0);
```
sergei-maertens, line 114: I've take a different approach here - rather than asserting that X number of API calls were made, the expected result is asserted. First load the form in NL -> Dutch form title visible, then change the language and this should result in the English variants being displayed instead.

src/sdk.spec.js
CharString: Probably copy pasted.

src/sdk.spec.js
CharString: This now waits until all `status` role children are gone from the form element, right? I have no idea whether other status elements are imaginable, nor the desire to deep dive into aria right now.

src/sdk.spec.js
CharString: Nice,  with the new mock the api is abstracted away. How these strings appear is not important.

src/api-mocks/submissions.js
```diff
@@ -0,0 +1,30 @@
...
+};
+
+/**
+ * Return a submission object as if it would be returned from the backend API.
+ * @param  {Object} overrides Key-value mapping with overrides from the defaults. These
+ *                            are deep-assigned via lodash.set to the defaults, so use
+ *                            '.'-joined strings as keys for deep paths.
+ * @return {Object}           A submission object.
+ */
+export const getSubmission = getDefaultFactory(SUBMISSION_DETAILS);
```
CharString, line 30:
```suggestion
export const makeSubmission = defaultFactory(SUBMISSION_DETAILS);
```

Or "build" to mirror FactoryBoy and convey that there's no persistency.  Same for the other factory functions.  Then there's no clash with the GET from the http method.

CharString (changes requested):
Reaching for `immer` and an overal functional style... I think you might like [ReScript](https://rescript-lang.org/).

sergei-maertens:
> Reaching for `immer` and an overal functional style... I think you might like [ReScript](https://rescript-lang.org/).

Probably, but until it gains wide community adoption (and given the popularity of TypeScript I don't think this will happen soon) it's not something to seriously consider.

src/api-mocks/base.js
```diff
@@ -0,0 +1,21 @@
...
+   */
+  const defaultFactory = (overrides = {}) => {
+    const obj = cloneDeep(default_object);
+    for (const [key, value] of Object.entries(overrides)) {
+      set(obj, key, value);
+    }
+    return obj;
+  };
+  return defaultFactory;
+};
```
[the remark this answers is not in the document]
sergei-maertens: ah there was no particular reason other than easy of putting a debugger before the return statement. I don't have any particular strong preferences.

The docbloc does belong to the curried function though, the outside function is a factory that only takes in a `default_object` argument & returns a function. We don't currently generate any documentation from these docblocks so it's more whatever your editor uses. Preferably we switch to TypeScript anyway so the structured typing information is in the code rather than the annotations.

src/api-mocks/submissions.js
sergei-maertens: good suggestion, I'll adapt the code!

PR #375, opened 2023-03-03
Improve password field

SilviaAmAm:


src/formio/components/Password.js
```diff
@@ -14,6 +14,11 @@ class Password extends Formio.Components.components.password {
     );
     return info;
   }
+
+  getView(value, options) {
+    // In Formio they use '--- PROTECTED ---' which then somewhere gets stripped out
+    return '*'.repeat(value.length);
```
KoenBrouwer, line 20: I wouldn't use the length of the actual value.

```suggestion
    return '*'.repeat(12);
```

sergei-maertens:
https://github.com/open-formulieren/open-forms/pull/2886 moves the password field to the "deprecated" group as checking with customers reveals that one uses this in production nor have any concrete examples why/what they would use it for. It will be removed at some point. This also means that we should not waste any resources on this component for could-be's or maybes.

PR #381, opened 2023-03-22
[OF#2912] Fix disappearing drag and drop area

SilviaAmAm:
This was previously fixed by adding the `abortUpload` method. But this gets overwritten during successful upload of a file here: https://github.com/formio/formio.js/blob/v4.13.13/src/components/file/File.js#L748. So this time I tried changing the event listener to directly update the fileDropHidden attribute

Fixes open-formulieren/open-forms#2912


sergei-maertens:
Backports:

* 0a3506275be74ccdb678b0a2b6a112d9b3ffd70b (1.3.x)
* 21003970d88f171030a0e606b200eb00a43b043c (1.2.x)
* 3676998b4db1922d5dd667d1a83718f6d66d2858 (1.1.x)

PR #385, opened 2023-03-29
[#2444] hide non-applicable steps from overview

pi-sigma:
Closes open-formulieren/open-forms#2444

*non-applicable steps were visible in the overview but marked as 'n.v.t' 
*they are now hidden by default
*margin of list items and padding of progress indicator items are adjusted to make the UI consistent


src/components/ProgressIndicator/ProgressItem.js
```diff
@@ -5,9 +5,9 @@ import {getBEMClassName} from 'utils';
 
 import CompletionMark from './CompletionMark';
 
-const ProgressItem = ({completed, children}) => {
+const ProgressItem = ({completed, children, modifiers = []}) => {
```
SilviaAmAm, line 8: Since you added a prop, it is good to also add it to the propTypes on line 20 :slightly_smiling_face:

src/scss/components/_progress-indicator.scss
```diff
@@ -84,8 +84,14 @@
 
 .#{prefix('progress-indicator-item')} {
   display: flex;
+  padding: 10px;
```
SilviaAmAm, line 87: Maybe good to use the `@include margin` that was used in the list css?

Maybe like this?
```suggestion
@include margin($grid-margin-1, $properties: padding);
```

src/scss/components/_progress-indicator.scss
```diff
@@ -84,8 +84,14 @@
...
   justify-content: flex-start;
 
+  @include bem.modifier('not-applicable') {
+    display: none;
+    padding: 0;
+  }
```
SilviaAmAm, line 93: I think that this is how you would specify this property with the design token, maintaining the current default behaviour:

```suggestion
  @include bem.modifier('not-applicable') {
    display: var(--of-progress-indicator-item-not-applicable-display, flex);
  }
```

SilviaAmAm (changes requested):

src/components/ProgressIndicator/ProgressIndicatorDisplay.js
```diff
@@ -60,7 +60,8 @@ const SidebarStepStatus = ({
   completed = false,
 }) => {
   return (
-    <ProgressItem completed={completed}>
+    <ProgressItem completed={completed} modifiers={isApplicable ? [] : ['not-applicable']}>
+      {' '}
```
SilviaAmAm, line 64:
```suggestion
```
This can be removed!

SilviaAmAm (changes requested):
Looks good! Just one more minor comment :slightly_smiling_face:

src/components/List.js
```diff
@@ -8,6 +8,7 @@ const List = ({children, modifiers = [], ordered = false, component = ''}) => {
   const className = getBEMClassName('list', modifiers);
   // remove empty children
   const filtered = children.filter(child => !!child);
+
```
sergei-maertens, line 11: this is an unnecessary change, no?

src/components/ProgressIndicator/ProgressItem.js
```diff
@@ -20,6 +20,7 @@ const ProgressItem = ({completed, children}) => {
 ProgressItem.propTypes = {
   completed: PropTypes.bool.isRequired,
   children: PropTypes.node,
+  modifiers: PropTypes.array,
```
sergei-maertens, line 23: make this explicit with all the supported modifiers please: `PropTypes.oneOf(['opt1', 'opt2'])`.

This is what the storybook documentation uses for information about possible radio choices iirc.

src/components/ProgressIndicator/ProgressIndicatorDisplay.js
```diff
@@ -60,7 +60,7 @@ const SidebarStepStatus = ({
   completed = false,
 }) => {
   return (
-    <ProgressItem completed={completed}>
+    <ProgressItem completed={completed} modifiers={isApplicable ? [] : ['not-applicable']}>
```
sergei-maertens, line 63: imo it would be a nicer API to have a prop `isApplicable` and let the component `ProgressItem` translate that into the CSS modifiers. In this component, you want to reason in props instead of CSS classes. That also encapsulates the specific CSS class names to the `ProgressItem` component instead of leaking into the `ProgressIndicatorDisplay` component.

src/components/ProgressIndicator/ProgressIndicatorDisplay.js
```diff
@@ -158,7 +158,7 @@ const ProgressIndicatorDisplay = ({
 
         <Caption component="h2">{formTitle}</Caption>
 
-        <List ordered>
+        <List ordered modifiers={['extra-compact']}>
```
sergei-maertens, line 161: I don't see in the ticket where this is requested?

src/scss/components/_progress-indicator.scss
```diff
@@ -84,8 +84,13 @@
 
 .#{prefix('progress-indicator-item')} {
   display: flex;
+  @include margin($grid-margin-1, $properties: padding);
   justify-content: flex-start;
 
+  @include bem.modifier('not-applicable') {
+    display: var(--of-progress-indicator-item-not-applicable-display, flex);
+  }
```
sergei-maertens, line 92: This CSS-only fix is an accesibility problem, as there's nothing that's now marking the progress item as visually hidden. Screen readers will still read it out, while it's not visible.

I don't think a design token is the right approach here - rather the progress item should be `return null` if it's configured to be hidden for non-applicable steps OR have an `aria-hidden="true"` attribute for accessibility. Both require the backend to expose some information whether progress items should be displayed or not.

sergei-maertens (changes requested):

src/scss/components/_progress-indicator.scss
joeribekker: > You can also use CSS to set display: none or visibility: hidden to hide an element from screen readers and visually.

- https://www.pluralsight.com/guides/how-to-hide-text-from-screen-readers

sergei-maertens: Ah cool, I didn't realize this! Resolving the conversation

src/components/ProgressIndicator/ProgressIndicatorDisplay.js
sergei-maertens, line 1: Can you please add the new `hideNonApplicableSteps` proptype to the proptypes for the component?

src/components/List.js
sergei-maertens: can you please actually remove this line again so the diff is not cluttered?

src/components/Form.js
```diff
@@ -285,6 +285,7 @@ const Form = ({form}) => {
     <ProgressIndicator
       title={form.name}
       steps={form.steps}
+      hideNonApplicableSteps={form.hideNonApplicableSteps}
```
sergei-maertens, line 288: Can you please add the new `hideNonApplicableSteps` prop to the mocks in `src/api-mocks/forms.js`?

src/components/ProgressIndicator/index.js
```diff
@@ -16,6 +16,7 @@ const ProgressIndicator = ({
   steps,
   submissionAllowed,
   completed = false,
+  hideNonApplicableSteps = false,
```
sergei-maertens, line 19: please add this prop type to the `ProgressIndicator` component proptypes

src/components/ProgressIndicator/index.js
```diff
@@ -24,6 +25,7 @@ const ProgressIndicator = ({
   const confirmationMatch = !!useRouteMatch('/bevestiging');
   const isStartPage = !summaryMatch && stepMatch == null && !confirmationMatch;
   const [expanded, setExpanded] = useState(false);
+  // const hideSteps = config.hideSteps;
```
sergei-maertens, line 28:
```suggestion
```

sergei-maertens (changes requested):

PR #390, opened 2023-04-05
[#2240] Added hash routing option

joeribekker:
Fixes open-formulieren/open-forms#2240

Allows you to enable hash routing by adding `useHashRouting: true` in the Javascript snippet.
It seemed odd to me to influence this from the backend since it's really a frontend/integration setting.


src/sdk.js
```diff
@@ -122,9 +124,15 @@ class OpenForm {
...
-              </Router>
+              {this.useHashRouting ? (
+                <HashRouter basename={this.basePath}>
+                  <App form={this.formObject} />
+                </HashRouter>
+              ) : (
+                <BrowserRouter basename={this.basePath}>
+                  <App form={this.formObject} />
+                </BrowserRouter>
+              )}
```
sergei-maertens, line 135: you can also make the component dynamic:

```jsx
const Router = this.useHashRouting ? HashRouter : BrowserRouter;
...
<Router basename={this.basePath><App ... /></Router>
```
so then the actual router wrapper code is not modified, just the component alias is.

sergei-maertens (changes requested):

src/sdk.js
joeribekker: Heh, I had this but changed it to make it more clear/specific. I'll change to this.

joeribekker:
After your review I made 2 more changes:

1. I changed `...hash.substring(2)` to `...hash.substring(1)`. This basically only removes the `#` now and can lead to double slashes: `example.com/foo/#/stap/1` -> `example.com/foo/#/stap/1` but in case the page does not end in a `/` at least you can distinguish the path elements: `example.com/foo#/stap/1` -> `example.com/foo/stap/1`
2. I changed the dev logging to not use substring. Removing data before the developer sees it is harder to figure out than to remove the data in your head.

sergei-maertens:
Nice, LGTM!

sergei-maertens (approved):

PR #391, opened 2023-04-07
[OF#2788] Change startpage URL and name in progress indicator

SilviaAmAm:
Changes:
* Now the URL for the initial landing page has 'startpagina' in the URL
* The step in the progress indicator is called 'start page' instead of 'inloggen'

Fixes open-formulieren/open-forms#2788


sergei-maertens (approved):
Cleanest solution so that old URLs being redirect to from the backend still work :+1:

PR #392, opened 2023-04-11
[OF#2921] Add form title to form step

SilviaAmAm:
Screenshot from 2023-04-11 12-58-49
Screenshot from 2023-04-11 12-59-06
Fixes open-formulieren/open-forms#2921


src/scss/components/_card.scss
sergei-maertens, line 1: If we can do this without nested cards, these style changes don't seem necessary to me.

sergei-maertens (approved):

PR #399, opened 2023-04-18
:sparkles: - Hide suspend button if suspending form is not allowed.

svenvandescheur:
Depends on: https://github.com/open-formulieren/open-forms/pull/3006

### DIFF
--- a/src/components/ButtonsToolbar/index.js
+++ b/src/components/ButtonsToolbar/index.js
@@ -12,7 +12,7 @@ const ButtonsToolbar = ({
   literals,
   canSubmitStep,
   canSubmitForm,
-  loginRequired,
+  canSuspendForm,
   isAuthenticated,
   isLastStep,
   isCheckingLogic,
@@ -32,9 +32,11 @@ const ButtonsToolbar = ({
             </Button>
           </ToolbarList>
           <ToolbarList>
-            <Button type="button" variant="secondary" name="save" onClick={onFormSave}>
-              <Literal name="saveText" />
-            </Button>
+            {canSuspendForm && (
+              <Button type="button" variant="secondary" name="save" onClick={onFormSave}>
+                <Literal name="saveText" />
+              </Button>
+            )}
             {showSubmitButton && (
               <Button type="submit" variant="primary" name="next" disabled={!canSubmitStep}>
                 {isCheckingLogic ? (
@@ -56,8 +58,8 @@ ButtonsToolbar.propTypes = {
   literals: PropTypes.object,
   canSubmitStep: PropTypes.bool.isRequired,
   canSubmitForm: PropTypes.string.isRequired,
+  canSuspendForm: PropTypes.string.isRequired,
   isLastStep: PropTypes.bool.isRequired,
-  loginRequired: PropTypes.bool.isRequired,
   isAuthenticated: PropTypes.bool.isRequired,
   isCheckingLogic: PropTypes.bool.isRequired,
   onNavigatePrevPage: PropTypes.func.isRequired,
--- a/src/components/ButtonsToolbar/index.js
+++ b/src/components/ButtonsToolbar/index.js
@@ -58,7 +58,7 @@ ButtonsToolbar.propTypes = {
   literals: PropTypes.object,
   canSubmitStep: PropTypes.bool.isRequired,
   canSubmitForm: PropTypes.string.isRequired,
-  canSuspendForm: PropTypes.string.isRequired,
+  canSuspendForm: PropTypes.bool.isRequired,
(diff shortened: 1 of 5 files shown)


src/components/ButtonsToolbar/index.js
```diff
@@ -12,7 +12,7 @@ const ButtonsToolbar = ({
   literals,
   canSubmitStep,
   canSubmitForm,
-  loginRequired,
```
sergei-maertens, line 15: can you remove the prop type too?

src/components/ButtonsToolbar/test.spec.js
```diff
@@ -141,3 +145,32 @@ it('Non-last step of non-submittable form without overview, button IS present',
...
+    render(
+      <ButtonsToolbar
+        literals={LITERALS}
+        canSubmitStep={true}
+        canSubmitForm={SUBMISSION_ALLOWED.yes}
+        canSuspendForm={false}
+        isAuthenticated={false}
+        isLastStep={true}
+        isCheckingLogic={false}
+        loginRequired={false}
```
sergei-maertens, line 162: prop is not used anymore

src/components/ButtonsToolbar/test.spec.js
```diff
@@ -141,3 +145,32 @@ it('Non-last step of non-submittable form without overview, button IS present',
...
+      />,
+      container
+    );
+  });
+
+  const buttons = container.getElementsByClassName('openforms-toolbar__list-item');
+
+  expect(buttons.length).toEqual(2);
+  expect(buttons[0].textContent).toEqual('Previous step');
+  expect(buttons[1].textContent).toEqual('Next step');
```
sergei-maertens, line 175: please make the assertion that the 'Save step' literal is not present, as that is what the test/prop is about. If we add or tweak another unrelated button, this test will break while the `canSuspendForm` behaviour has not changed.

I'm really expecting something like `expect(await screen.queryByRole('button', {name: 'Save step'}).toBeNull()`

src/components/FormStep/index.js
```diff
@@ -818,6 +818,7 @@ const FormStep = ({
                 literals={formStep.literals}
                 canSubmitStep={canSubmit}
                 canSubmitForm={submission.submissionAllowed}
+                canSuspendForm={form.suspendAllowed}
```
sergei-maertens, line 821: can you please update the mocks in https://github.com/open-formulieren/open-forms-sdk/blob/main/src/api-mocks/forms.js#L6

sergei-maertens (changes requested):

sergei-maertens (changes requested):
Can you also update the relevant story that is causing this visual regression? Might be fixed by updating the mocks.

https://www.chromatic.com/pullrequest?appId=634043ae879dc17405b0e567&number=399&activeElementId=comparison-test-key-63f8a7518c4a45bf32e29812-1200&view=changes

src/components/ButtonsToolbar/test.spec.js
svenvandescheur: The nicest way to do this would be to introduce something like testing library here, but that would either differentiate from the other tests in the same file (like testing that the submit button is hidden).  I'd rather have a consistent testing approach for all the tests for a component. Do we want to refactor the other tests as well?

sergei-maertens: I thought testing-library is already set up?

```
➜  open-forms-sdk git:(main) yarn list --pattern=testing-library
yarn list v1.22.19
├─ @storybook/jest@0.0.10
│  └─ @testing-library/jest-dom@5.16.5
├─ @storybook/testing-library@0.0.13
│  └─ @testing-library/user-event@13.5.0
├─ @testing-library/dom@8.16.0
├─ @testing-library/jest-dom@5.11.10
├─ @testing-library/react@11.2.6
│  └─ @testing-library/dom@7.30.3
├─ @testing-library/user-event@12.8.3
├─ @types/testing-library__jest-dom@5.9.5
└─ eslint-plugin-testing-library@5.4.0
```

svenvandescheur: Yes testing library is already available so can easily be added and I agree that making sure something doesn't exist is more accurate (and descriptive).

I find different patterns in similar tests confusing but as discussed: I will create an issue for this.

sergei-maertens (changes requested):
* Can you add a FormStep story with suspension disallowed?
* Can you clean up the commit history and make sure ticket number references are included in the format `open-formulieren/open-forms#issueid`

sergei-maertens (approved):

PR #403, opened 2023-04-25
Added padding under form step header

SilviaAmAm:
Leftover from open-formulieren/open-forms#2921

I didnt like how close the header was to the label of the first field:
Screenshot from 2023-04-25 16-25-22

### DIFF
--- a/src/components/Card.js
+++ b/src/components/Card.js
@@ -5,7 +5,13 @@ import React from 'react';
 import Caption from 'components/Caption';
 import {getBEMClassName} from 'utils';
 
-const CardTitle = ({title, component = 'h1', headingType = 'title', blockClassName = 'card'}) => {
+const CardTitle = ({
+  title,
+  component = 'h1',
+  headingType = 'title',
+  blockClassName = 'card',
+  modifiers = [],
+}) => {
   const headingLevel = {
     h1: 1,
     h2: 2,
@@ -16,7 +22,7 @@ const CardTitle = ({title, component = 'h1', headingType = 'title', blockClassNa
   };
 
   return (
-    <header className={getBEMClassName(`${blockClassName}__header`)}>
+    <header className={getBEMClassName(`${blockClassName}__header`, modifiers)}>
       <Heading level={headingLevel[component] || 1} className={getBEMClassName(headingType)}>
         {title}
       </Heading>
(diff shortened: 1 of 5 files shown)


src/components/Card.js
```diff
@@ -29,6 +35,7 @@ CardTitle.propTypes = {
   component: PropTypes.string,
   blockClassName: PropTypes.string,
   headingType: PropTypes.oneOf(['title', 'subtitle']),
+  modifiers: PropTypes.array,
```
sergei-maertens, line 38: Try to make prop types as explicit as possible - these modifiers are a known, finite list of possible values:

```suggestion
  modifiers: PropTypes.arrayOf(
    PropTypes.oneOf(['padded'])
  ),
```

src/scss/components/_card.scss
```diff
@@ -10,12 +10,18 @@
   padding: $grid-margin-8;
   box-sizing: border-box;
 
+  @include bem.element('header') {
+    &--padded {
+      @include rows(1, padding-bottom);
+    }
```
sergei-maertens, line 16:
```suggestion
    @include bem.modifier('padded') {
      @include rows(1, padding-bottom);
    }
```

The `&--` syntax results in a selector `.openforms-card .openforms-card__header--padded`, while the mixin syntax results in the stricter `.openforms-card .openforms-card__header.openforms-card__header--padded`

In BEM, you should *not* be able to use a modifier without also specifying the base that was modified.

Do:

`<span class="block__element block__element--modifier">`

Don't:

`<span class="block__element--modifier">`

Making the selector specific enforces correct classnames in the markup.

src/scss/components/_card.scss
```diff
@@ -10,12 +10,18 @@
...
+  }
+
   @include mobile-only {
     @include container-margin($properties: padding);
 
     @include bem.element('header') {
-      &:not(:last-child) {
-        margin-bottom: $grid-margin-2;
+      &--padded {
```
sergei-maertens, line 23:
```suggestion
      @include bem.modifier('padded') {
```

sergei-maertens (changes requested):

src/scss/components/_card.scss
sergei-maertens: I've seen this before because of some string utilities being missing :thinking: I'll see if I can look into it tomorrow

sergei-maertens (approved):

sergei-maertens (approved):
Should now actually be fixed... pff :)

PR #404, opened 2023-04-25
[OF#2986] Make existing submission obvious

SilviaAmAm:
Fixes open-formulieren/open-forms#2986

Screenshot from 2023-04-26 15-27-40
Screenshot from 2023-04-26 15-27-32

### DIFF
--- a/src/components/FormStart/index.js
+++ b/src/components/FormStart/index.js
@@ -40,7 +40,7 @@ const FormStartMessage = ({form}) => {
  * This is shown when the form is initially loaded and provides the explicit user
  * action to start the form, or present the login button (DigiD, eHerkenning...)
  */
-const FormStart = ({form, hasActiveSubmission, onFormStart}) => {
+const FormStart = ({form, hasActiveSubmission, onFormStart, onFormAbort}) => {
   const doStart = useStartSubmission();
   const outagePluginId = useDetectAuthenticationOutage();
   const authErrors = useDetectAuthErrorMessages();
@@ -126,7 +126,7 @@ const FormStart = ({form, hasActiveSubmission, onFormStart}) => {
         <FormStartMessage form={form} />
 
         {hasActiveSubmission ? (
-          <ExistingSubmissionOptions form={form} />
+          <ExistingSubmissionOptions form={form} onFormAbort={onFormAbort} />
         ) : (
           <LoginOptions form={form} onFormStart={onFormStart} />
         )}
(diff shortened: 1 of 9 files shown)


src/components/FormStart/tests.spec.js
```diff
@@ -75,3 +87,29 @@ it('Form start does not start if there are auth errors', () => {
...
+      </IntlProvider>,
+      container
+    );
+  });
+
+  const buttons = container.getElementsByTagName('button');
+
+  expect(buttons.length).toEqual(2);
+  expect(buttons[0].textContent).toEqual('Continue existing submission');
+  expect(buttons[1].textContent).toEqual('Abort existing submission');
```
sergei-maertens, line 114: can you please use testing-library/react patterns for the tests rather than this low-level DOM stuff?

You can check https://github.com/open-formulieren/open-forms-sdk/pull/399 for an example

Additionally, going forward we should focus on writing storybook stories and implementing the `play` function for interaction tests, this automatically takes care of mocking callbacks etc. and is generally a bit more manageable than writing barebones jest tests.

You can find many examples in the https://github.com/open-formulieren/formio-builder repository, e.g. https://github.com/open-formulieren/formio-builder/blob/main/src/components/builder/key.stories.tsx#L96

src/components/FormStart/index.js
```diff
@@ -125,15 +125,21 @@ const FormStart = ({form, onFormStart}) => {
...
+          <LoginOptions form={form} onFormStart={onFormStart} />
+        )}
       </Card>
     </LiteralsProvider>
   );
 };
 
 FormStart.propTypes = {
   form: Types.Form.isRequired,
+  hasActiveSubmission: PropTypes.bool,
```
sergei-maertens, line 140: if you don't provide a default, please mark the prop as required

```suggestion
  hasActiveSubmission: PropTypes.bool.isRequired,
```

sergei-maertens (changes requested):

src/components/FormStart/tests.spec.js
```diff
@@ -75,3 +88,24 @@ it('Form start does not start if there are auth errors', () => {
...
+        form={testForm}
+        onFormStart={onFormStart}
+        onFormAbort={onFormAbort}
+        hasActiveSubmission={true}
+      />
+    </IntlProvider>,
+    container
+  );
+
+  expect(screen.queryByText('Continue existing submission')).toBeInTheDocument();
```
sergei-maertens, line 109:
```suggestion
  expect(screen.queryByRole('button', {name: 'Continue existing submission'})).toBeInTheDocument();
```

See https://testing-library.com/docs/queries/about#priority - if you can't query by accessible role, this is often an indication that there is an a11y issue.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #405, opened 2023-04-26
[OF#2929] Restarting form after pause

SilviaAmAm:
Fixes open-formulieren/open-forms#2929

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -104,6 +104,10 @@ const reducer = (draft, action) => {
         config: draft.config,
       };
     }
+    case 'RESET': {
+      const initialState = action.payload;
+      return initialState;
+    }
     default: {
       throw new Error(`Unknown action ${action.type}`);
     }
@@ -351,6 +355,11 @@ const Form = ({form}) => {
                 }
                 onStepSubmitted={onStepSubmitted}
                 onLogout={onLogout}
+                onSessionDestroyed={() => {
+                  resetSession();
+                  history.push('/');
+                  dispatch({type: 'RESET', payload: initialStateFromProps});
+                }}
                 component={FormStep}
               />
             </RequireSession>
--- a/src/components/FormStep/index.js
+++ b/src/components/FormStep/index.js
@@ -834,6 +834,9 @@ const FormStep = ({
         isOpen={isFormSaveModalOpen}
         closeModal={closeFormStepSaveModal}
         onSaveConfirm={onSaveConfirm}
+        onSessionDestroyed={() => {
+          localStorage.removeItem(form.uuid);
+        }}
         suspendFormUrl={`${submission.url}/_suspend`}
         suspendFormUrlLifetime={form.resumeLinkLifetime}
         submissionId={submission.id}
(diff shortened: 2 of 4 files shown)


src/components/FormStep/index.js
```diff
@@ -834,6 +834,9 @@ const FormStep = ({
         isOpen={isFormSaveModalOpen}
         closeModal={closeFormStepSaveModal}
         onSaveConfirm={onSaveConfirm}
+        onSessionDestroyed={() => {
+          localStorage.removeItem(form.uuid);
+        }}
```
sergei-maertens, line 839: we should not use the global localStorage here, but instead use the `removeSubmissionId` callback from the `useRecycleSubmission` hook (which uses `useLocalStorage` under the hood).

src/components/modals/FormStepSaveModal.js
```diff
@@ -114,9 +115,12 @@ const FormStepSaveModal = ({
       return;
     }
 
+    onSessionDestroyed();
     // redirect back to start page
     dispatch({type: 'SAVE_SUCCEEDED'});
     history.push('/');
+    // TODO: replace with a proper reset of the state instead of a page reload.
+    window.location.reload();
```
sergei-maertens, line 123: can't we do this as part of this hook then? should just be a reducer action type so you can `dispatch({type: 'RESET'})` in the form step, part of the `onSessionDestroyed`, no?

sergei-maertens (changes requested):

src/components/Form.js
```diff
@@ -104,6 +104,10 @@ const reducer = (draft, action) => {
         config: draft.config,
       };
     }
+    case 'RESET': {
+      window.location.reload();
+      break;
+    }
```
sergei-maertens, line 110: Sorry, I didn't make myself clear enough.

There should *not* be a `window.location.reload()` anywhere, ideally, but especially not in a reducer because a reducer is responsible for producing the next state based on the dispatched action and the previous state.

A reset here should instead do something along the lines of `return initialState`, so that the "next" state is effectively a clean/empty state as if the component was just mounted.

The component calls the `useImmerReducer` hook here:

```js
  // load the state management/reducer
  const initialStateFromProps = {...initialState, config, step: steps[0]};
  const [state, dispatch] = useImmerReducer(reducer, initialStateFromProps);
```

So, to reset the state, we basically want to do something along the lines of: `dispatch({type: 'reset', payload: initialStateFromProps})`

and the reducer bit should then look something along the lines of:

```suggestion
    case 'RESET': {
      const initialStateFromProps = action.payload;
      return initialStateFromProps; 
    }
```

The reducer works in a couple of possible modes:

1. You return something. That return value is used as the next state (and immer makes it immutable).
2. You mutate the draft and do or don't `break`. If you don't `break`, the other `case`s below will also evaluate. Then immer takes the mutated draft and essentially makes that the next state.

So, in most cases a reset action is essentially `return initialState`, here it's a bit more complex because the actual initial state has a dynamic aspect from passed in props, but we can capture all that in the callback closures.

sergei-maertens (changes requested):

sergei-maertens (approved):

SilviaAmAm:
Backport to 1.3.x: https://github.com/open-formulieren/open-forms-sdk/pull/408
Backport to 1.2.x:

PR #406, opened 2023-05-01
Issue/of/2863 tab elements order

pi-sigma:
Closes open-formulieren/open-forms#2863


src/hooks/useTitle.js
```diff
@@ -4,8 +4,8 @@ import {useTitle as reactUseTitle} from 'react-use';
 import {ConfigContext} from 'Context';
 
 const useTitle = localTitle => {
-  const {titlePrefix} = useContext(ConfigContext);
-  return reactUseTitle(titlePrefix ? `${titlePrefix} | ${localTitle}` : localTitle);
+  const {documentTitle} = useContext(ConfigContext);
```
sergei-maertens, line 7: IMO a name like `titleSuffix` or `baseTitle` is more appropriate, because the point of this hook is setting the document title (`document.title`) so this could be confusing?

Looking at the backend code I see there could be leading/trailing whitespace too, so perhaps we should use:

```js
const baseTitle = useContext(ConfigContext).baseTitle.trim();
```

sergei-maertens (changes requested):
Can you also add the `baseTitle` default value (empty string) to the context here: https://github.com/open-formulieren/open-forms-sdk/blob/main/src/Context.js#L3 I noticed it was missing.

sergei-maertens (approved):

PR #407, opened 2023-05-01
[OF#3040] Improve error messages for file components

SilviaAmAm:
Fixes open-formulieren/open-forms#3040

### DIFF
--- a/src/formio/components/FileField.js
+++ b/src/formio/components/FileField.js
@@ -1,4 +1,5 @@
 import {default as formioUrlStorage} from 'formiojs/providers/storage/url';
+import FormioUtils from 'formiojs/utils';
 import {Formio} from 'react-formio';
 
 import {CSRFToken} from 'headers';
@@ -212,7 +213,62 @@ class FileField extends Formio.Components.components.file {
       this.loading = false;
     });
 
+    let hasTypeErrors = false;
+    let statuses = [];
+
+    // Issue #3040 - Overwriting code from the super, in order to update the error message for wrong
+    // file types, so that it doesn't show the (escaped) MIME type to the user.
+    Array.prototype.forEach.call(files, file => {
+      const fileUpload = {
+        originalName: file.name,
+        name: FormioUtils.uniqueName(
+          file.name,
+          this.component.fileNameTemplate,
+          this.evalContext()
+        ),
+        size: file.size,
+        status: 'info',
+        message: this.t('Processing file. Please wait...'),
+      };
+
+      if (this.component.filePattern && !this.validatePattern(file, this.component.filePattern)) {
+        fileUpload.status = 'error';
+        fileUpload.message = this.t(
+          'The uploaded file is not of an allowed type. It must be: {{ pattern }}.',
+          {
+            pattern: this.formatAllowedTypesLabels(
+              this.component.file.allowedTypesLabels,
+              this.component.filePattern
+            ),
+          }
+        );
+        hasTypeErrors = true;
+      }
+
+      statuses.push(fileUpload);
+    });
+
+    if (hasTypeErrors) {
+      this.statuses = this.statuses.concat(statuses);
+      this.redraw();
+      return;
+    }
+
     super.upload(files);
+  }
+
+  formatAllowedTypesLabels(allowedTypesLabels, filePattern) {
+    if (!allowedTypesLabels) return filePattern;
+
+    const numberOfAllowedTypes = allowedTypesLabels.length;
+    if (numberOfAllowedTypes > 1) {
+      return this.t('{{ labels }} or {{ lastLabel }}', {
+        labels: allowedTypesLabels.slice(0, numberOfAllowedTypes - 1).join(', '),
+        lastLabel: allowedTypesLabels[numberOfAllowedTypes - 1],
+      });
+    }
+
+    return allowedTypesLabels[0];
   }
 
   browseFiles(attrs = {}) {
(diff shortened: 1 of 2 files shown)


src/formio/components/FileField.js
```diff
@@ -212,9 +213,64 @@ class FileField extends Formio.Components.components.file {
...
+          }
+        );
+        hasTypeErrors = true;
+      }
+
+      statuses.push(fileUpload);
+    });
+
+    if (hasTypeErrors) {
+      this.statuses = this.statuses.concat(statuses);
```
CharString, line 252: We did push some objects with 'info' status on `statuses` But we only append them on `this.statuses` if an error has happened?

If that is correct, can we short circuit before the forEach if `this.component.filePattern` is falsy and `return super.upload(files)`?
And remove that invariant check from the callback function?

CharString (approved):
We discussed that it feels iffy to me to repeat a stable mapping on de components, but it's simplest.

src/formio/components/FileField.js
sergei-maertens: I like the suggestion, but it's unclear to me how much of this code is ours and how much is copied from Form.io and modified. If it's mostly copied code from Form.io, I'd keep it as is so that it's easier to see the diff between ours and theirs so that we can eventually get rid of this again when better options present themselves (either a formio upgrade or our own formio-renderer).

src/formio/components/FileField.js
```diff
@@ -212,9 +213,64 @@ class FileField extends Formio.Components.components.file {
...
+  formatAllowedTypesLabels(allowedTypesLabels, filePattern) {
+    if (!allowedTypesLabels) return filePattern;
+
+    const numberOfAllowedTypes = allowedTypesLabels.length;
+    if (numberOfAllowedTypes > 1) {
+      return this.t('{{ labels }} or {{ lastLabel }}', {
+        labels: allowedTypesLabels.slice(0, numberOfAllowedTypes - 1).join(', '),
+        lastLabel: allowedTypesLabels[numberOfAllowedTypes - 1],
+      });
+    }
```
sergei-maertens, line 269: I think you can also so something like:

```js
const [last, ...otherLabels] = allowedTypesLabels.reverse();
if (!otherLabels.length) return last;
return this.t(
  '{{ labels }} or {{ lastLabel }}', 
  {
    labels: otherLabels.join(', '),
    lastLabel: last,
  }
);
```
So basically use array destructuring to grab the tail of the array, and it should be able to handle arrays of len >= 1

src/formio/components/FileField.stories.mdx
```diff
@@ -49,6 +61,9 @@ export const ERRORS = ['File is too big', 'File is of the wrong type'];
 **Note** - file uploads on the Docs page will always error due to the way
 the network mocking library works. Use the canvas instead.
 
+By selecting a `filePattern` and uploading a file of different type, it is possible to see the
+error raised when a user uploads a file of wrong type.
```
sergei-maertens, line 65: Can you please implement a `play` function/interaction test that asserts the necessary error messages and formats are displayed?

sergei-maertens (changes requested):
I agree with Chris's sentiment about the elegance of it all, but I don't see a more elegant and pragmatic way to deal with this.

src/formio/components/FileField.stories.mdx
```diff
@@ -101,6 +119,75 @@ export const Template = ({
...
+      label: 'Attachment',
+      description: 'Klik om een bestand te selecteren en in te zenden. Alleen PDFs zijn toegestaan.',
+      required: false,
+      multiple: false,
+      filePattern: ['application/pdf']
+    }}
+    play={async ({args, canvasElement}) => {
+      const canvas = within(canvasElement);
+      // Click on browse to make the input node injected in the dom
+      const browseLink = canvas.getByText('browse');
```
sergei-maertens, line 144: I think `canvas.getByRole('link', {name: 'browse'})` is better?

src/formio/components/FileField.stories.mdx
```diff
@@ -101,6 +119,75 @@ export const Template = ({
...
+      label: 'Attachment',
+      description: 'Klik om een bestand te selecteren en in te zenden. Alleen PDFs en PNGs zijn toegestaan.',
+      required: false,
+      multiple: false,
+      filePattern: ['application/pdf', 'image/png']
+    }}
+    play={async ({args, canvasElement}) => {
+      const canvas = within(canvasElement);
+      // Click on browse to make the input node injected in the dom
+      const browseLink = canvas.getByText('browse');
```
sergei-maertens, line 176: see above

src/formio/components/FileField.stories.mdx
```diff
@@ -101,6 +119,75 @@ export const Template = ({
...
+    }}
+    args={{
+      id: 'fileUpload2',
+      label: 'Attachment',
+      description: 'Klik om een bestand te selecteren en in te zenden. Alleen PDFs zijn toegestaan.',
+      required: false,
+      multiple: false,
+      filePattern: ['application/pdf']
+    }}
+    play={async ({args, canvasElement}) => {
```
sergei-maertens, line 141: `args` is unused

src/formio/components/FileField.stories.mdx
```diff
@@ -101,6 +119,75 @@ export const Template = ({
...
+    }}
+    args={{
+      id: 'fileUpload2',
+      label: 'Attachment',
+      description: 'Klik om een bestand te selecteren en in te zenden. Alleen PDFs en PNGs zijn toegestaan.',
+      required: false,
+      multiple: false,
+      filePattern: ['application/pdf', 'image/png']
+    }}
+    play={async ({args, canvasElement}) => {
```
sergei-maertens, line 173: `args` is unused

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #412, opened 2023-05-10
Feature/text field

Xaohs:
Closes open-formulieren/open-forms#3057

### DIFF
--- a/src/components/forms/TextField/TextField.js
+++ b/src/components/forms/TextField/TextField.js
@@ -1,0 +1,54 @@
+import {
+  FormFieldDescription,
+  FormLabel,
+  Paragraph,
+  Textbox,
+  FormField as UtrechtFormField,
+} from '@utrecht/component-library-react';
+import {Field} from 'formik';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import {getBEMClassName} from 'utils';
+
+export const TextField = ({
+  name,
+  label = '',
+  isRequired = false,
+  description = '',
+  id = '',
+  disabled = false,
+  invalid = false,
+}) => {
+  const labelClassName = getBEMClassName('label', [isRequired && 'required'].filter(Boolean));
+
+  const inputProps = {
+    id,
+    disabled,
+    invalid,
+  };
+
+  return (
+    <UtrechtFormField type={'text'} invalid={invalid}>
+      <Paragraph className={labelClassName}>
+        <FormLabel htmlFor={id}>{label}</FormLabel>
+      </Paragraph>
+      <Paragraph>
+        <Field name={name} as={Textbox} {...inputProps} />
+      </Paragraph>
+      {description && <FormFieldDescription invalid={invalid}>{description}</FormFieldDescription>}
+    </UtrechtFormField>
+  );
+};
+
+TextField.propTypes = {
+  name: PropTypes.string.isRequired,
+  label: PropTypes.string,
+  isRequired: PropTypes.bool,
+  description: PropTypes.string,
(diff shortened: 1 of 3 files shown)


src/components/forms/TextField/TextField.js
```diff
@@ -0,0 +1,47 @@
...
+
+export const TextField = ({
+  label = '',
+  isRequired = false,
+  description = '',
+  id = '',
+  disabled,
+  invalid,
+}) => {
+  const labelClassName = classNames(getBEMClassName('label'), {'field-required': isRequired});
```
sergei-maertens, line 21: This `field-required` is because of the existing/legacy styles, right? I think I'd rather normalize that to be a modifier:

```suggestion
  const labelClassName = getBEMClassName(
    'label',
    [isRequired && 'required'].filter(Boolean),
  ));
```
and then update the `_label.scss` file to have:

```scss

// ... other imports

.#{prefix(label)} {
  @include body;
  @include body--big;
  @include anchor.extend-utrecht-link;
  display: block;
  font-weight: var(--of-label-font-weight);

  // .field-required is deprecated, use the modifier instead.
  &.field-required,
  @include bem.modifier('required') {
    &:after {
      content: '*';
      color: var(--of-color-danger);
      z-index: 0; // otherwise this is above the progress indicator in some browsers
    }  
  }
}
```

src/components/forms/TextField/TextField.js
```diff
@@ -0,0 +1,47 @@
...
+    <UtrechtFormField type={'text'} invalid={invalid}>
+      <Paragraph className={labelClassName} onClick={handleLabelClick}>
+        <FormLabel htmlFor={id}>{label}</FormLabel>
+      </Paragraph>
+      <Paragraph>
+        <Textbox {...inputProps} />
+      </Paragraph>
+      {description && <FormFieldDescription invalid={invalid}>{description}</FormFieldDescription>}
+    </UtrechtFormField>
+  );
```
sergei-maertens, line 46: We do want to wire this up with Formik directly, and I think something along these lines should be possible:

```jsx
import {Field} from 'formik';

<Paragraph>
  <Field name={name} as={Textbox} {...inputProps} />
</Paragraph>
```

-> the `name` prop should be added, it's the only required thing for formik

the `inputProps` can override a bunch of defaults or add additional things that should end up on the underlying `<input>` element and formik will make sure to hook up the `onChange`, `onFocus` and `onBlur` events and pass the `value` prop from its own internal form state.

This should also alleviate the need for the `ref`

src/components/forms/TextField/TextField.stories.mdx
```diff
@@ -0,0 +1,32 @@
...
+    name="Default"
+    decorators={[FormikDecorator]}
+    args={{
+      id: 'test',
+      label: "test",
+      description: "This is a custom description",
+      disabled: false,
+      invalid: false,
+      isRequired: true,
+    }}
```
sergei-maertens, line 28: this will need the `name` arg/prop as well, matching with the key in the `initialValues` data.

src/components/forms/TextField/test.spec.js
sergei-maertens, line 1: we're phasing out jest based tests in favour of adding them in storybook stories `play` function.

Chromatic renders the stories and executes the interaction tests for us in the CI pipeline, so please convert these to interaction tests in storybook.

sergei-maertens (changes requested):

sergei-maertens (approved):
Did some final tweaks and rebased the PR into a single commit.

PR #413, opened 2023-05-10
:sparkles: Features #3058 Implemented Number Field, Story, and Test.

Xaohs:
Closes open-formulieren/open-forms#3058

### DIFF
--- a/src/components/forms/NumberField/NumberField.js
+++ b/src/components/forms/NumberField/NumberField.js
@@ -1,0 +1,83 @@
+import {
+  FormFieldDescription,
+  FormLabel,
+  Paragraph,
+  Textbox,
+  FormField as UtrechtFormField,
+} from '@utrecht/component-library-react';
+import {useField} from 'formik';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import {getBEMClassName} from 'utils';
+
+const NumberField = ({
+  name,
+  min,
+  step,
+  label = '',
+  isRequired = false,
+  description = '',
+  id = '',
+  disabled = false,
+  invalid = false,
+}) => {
+  // We have to use the hook here instead of the Field component since we need to intercept the onChange event
+  const [field] = useField(name);
+
+  const labelClassName = getBEMClassName('label', [isRequired && 'required'].filter(Boolean));
+
+  const onChange = e => {
+    const inputValue = e.target.value;
+    const numericValue = Number(inputValue);
+
+    // Check if the input value is not a number and not a minus sign
+    if (isNaN(numericValue) && inputValue !== '-') {
+      return;
+    }
+    // Check if the numeric value is below the minimum value
+    if (min && numericValue < min) {
+      return;
+    }
+    // Check if the numeric value is not divisible by the step value
+    if (step && (numericValue % step !== 0 || inputValue.includes('.'))) {
+      return;
+    }
+    field.onChange(e);
+  };
+
+  const inputProps = {
+    ...field,
+    id,
+    disabled,
+    invalid,
+    min,
+    step,
+    type: 'text',
+    onChange,
+  };
+  return (
+    <UtrechtFormField type="text" invalid={invalid.toString()}>
+      <Paragraph className={labelClassName}>
+        <FormLabel htmlFor={id}>{label}</FormLabel>
+      </Paragraph>
+      <Paragraph>
+        <Textbox {...inputProps} />
+      </Paragraph>
+      {description && <FormFieldDescription invalid={invalid}>{description}</FormFieldDescription>}
+    </UtrechtFormField>
+  );
+};
+NumberField.propTypes = {
+  name: PropTypes.string.isRequired,
+  min: PropTypes.number,
+  step: PropTypes.number,
+  label: PropTypes.string,
+  isRequired: PropTypes.bool,
+  description: PropTypes.string,
+  id: PropTypes.string,
+  disabled: PropTypes.bool,
+  invalid: PropTypes.bool,
+};
+
+export default NumberField;
--- a/src/components/forms/NumberField/NumberField.js
+++ b/src/components/forms/NumberField/NumberField.js
@@ -36,14 +36,16 @@ const NumberField = ({
   const decimalSeparator = numberFormat
     .formatToParts(1.1)
     .find(part => part.type === 'decimal').value;
-  const thousandSeparator = decimalSeparator === ',' ? '.' : ',';
-  const isAllowedToTypeDecimals = !step;
+  const thousandSeparator = numberFormat
+    .formatToParts(1000)
+    .find(part => part.type === 'group').value;
+  const isAllowedToTypeDecimals = !step || !Number.isInteger(step);
 
   // This function makes sure that the user can't type a number that is lower than the min value
   const isAllowed = values => {
     const {formattedValue, floatValue} = values;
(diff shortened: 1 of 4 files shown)


src/components/forms/NumberField/NumberField.js
```diff
@@ -0,0 +1,56 @@
...
+  const labelClassName = getBEMClassName('label', [isRequired && 'required'].filter(Boolean));
+
+  const inputProps = {
+    name,
+    as: Textbox,
+    id,
+    disabled,
+    invalid,
+    min: 0,
+    step: 1,
```
sergei-maertens, line 33: these should not be hardcoded, but be props of the `NumberField` itself. You can then add two stories - one without `min` / `step`  specified -> assert that you can enter negative decimal values in the input and a second story with `min: 0` and `step: 1` specified -> assert that you can not enter negative or decimal values

src/components/forms/NumberField/NumberField.js
```diff
@@ -0,0 +1,56 @@
...
+      </Paragraph>
+      <Paragraph>
+        <Field {...inputProps} />
+      </Paragraph>
+      {description && <FormFieldDescription invalid={invalid}>{description}</FormFieldDescription>}
+    </UtrechtFormField>
+  );
+};
+NumberField.propTypes = {
+  name: PropTypes.string,
```
sergei-maertens, line 49:
```suggestion
  name: PropTypes.string.isRequired,
```

src/components/forms/NumberField/NumberField.js
```diff
@@ -0,0 +1,56 @@
...
+  FormFieldDescription,
+  FormLabel,
+  Paragraph,
+  Textbox,
+  FormField as UtrechtFormField,
+} from '@utrecht/component-library-react';
+import classNames from 'classnames';
+import {Field} from 'formik';
+import PropTypes from 'prop-types';
+import React, {useRef} from 'react';
```
sergei-maertens, line 11:
```suggestion
import React from 'react';
```
unused import

src/components/forms/NumberField/NumberField.stories.mdx
```diff
@@ -0,0 +1,50 @@
...
+
+<Canvas>
+  <Story
+    name="Default"
+    decorators={[FormikDecorator]}
+    play={async ({canvasElement}) => {
+      const canvas = within(canvasElement);
+      await expect(canvas.getByRole('textbox')).toBeInTheDocument();
+      await expect(canvas.getByText('Amount')).toBeInTheDocument();
+      await expect(canvas.getByText('This is a custom description for the amount field')).toBeInTheDocument();
```
sergei-maertens, line 32: I personally like `expect(canvas.getByText(...)).toBeVisible()` [a better assertion](https://github.com/testing-library/jest-dom#tobevisible)

sergei-maertens (changes requested):

src/components/forms/NumberField/NumberField.js
```diff
@@ -0,0 +1,80 @@
...
+    const numericValue = Number(inputValue);
+
+    // Check if the input value is a valid number or a hyphen
+    if (!isNaN(numericValue) || inputValue === '-') {
+      // Check if the numeric value is below the minimum value
+      if (min && numericValue < min) {
+        return;
+      }
+      // Check if the numeric value is not divisible by the step value
+      if (step && (numericValue % step !== 0 || inputValue.includes('.'))) {
```
sergei-maertens, line 38: additional thought - the input value should be interpreted according to the browser locale and be normalized as such :grimacing: 

e.g. NL/Dutch browser should be able to enter `1,5` while EN should enter `1.5` and this should both normalize to the float `1.5` in the actual form data.

sergei-maertens (changes requested):

src/components/forms/NumberField/NumberField.js
```diff
@@ -0,0 +1,99 @@
...
+    .formatToParts(1.1)
+    .find(part => part.type === 'decimal').value;
+  const thousandSeparator = decimalSeparator === ',' ? '.' : ',';
+  const isAllowedToTypeDecimals = !step;
+
+  // This function makes sure that the user can't type a number that is lower than the min value
+  const isAllowed = values => {
+    const {formattedValue, floatValue} = values;
+    const isEmpty = formattedValue === '';
+    const isBelowMin = min ? floatValue >= min : true;
```
sergei-maertens, line 46: should check if `min` is a number instead of truthy, because with `min=0` you end up in the `true` result instead of checking if the value is greater than or equal than 0

src/components/forms/NumberField/NumberField.stories.mdx
```diff
@@ -0,0 +1,93 @@
...
+      const canvas = within(canvasElement);
+      // Assert that you're able to provide a negative number
+      const input = canvas.getByRole('textbox');
+      userEvent.clear(input);
+      userEvent.type(input, '-1');
+      await expect(input).toHaveValue('-1');
+      // Assert that you're able to provide a decimal number
+      userEvent.clear(input);
+      userEvent.type(input, '1.5');
+      await expect(input).toHaveValue('1,5');
```
sergei-maertens, line 53:
```suggestion
      await expect(input).toHaveDisplayValue('1,5');
```
I think pretty much all occurrences should be replaced with `toHaveDisplayValue`, see https://github.com/testing-library/jest-dom#tohavedisplayvalue

src/components/forms/NumberField/NumberField.stories.mdx
```diff
@@ -0,0 +1,93 @@
...
+    name="Negative/Decimals Disabled"
+    decorators={[FormikDecorator]}
+    play={async ({canvasElement}) => {
+      await generalTest(canvasElement);
+      //Assert that you're not able to provide a negative number
+      const canvas = within(canvasElement);
+      const input = canvas.getByRole('textbox');
+      userEvent.clear(input);
+      userEvent.type(input, '-1');
+      await expect(input).toHaveValue('');
```
sergei-maertens, line 77: should use `toHaveDisplayValue` here, but also the test is currently failing. In storybook it discards the minus sign but accepts the `1` character, so it displays `1` instead of nothing.

That behaviour is okay, I think.

```suggestion
      userEvent.clear(input);
      userEvent.type(input, '-1');
      await expect(input).toHaveDisplayValue('1');
```

src/components/forms/NumberField/NumberField.js
```diff
@@ -0,0 +1,99 @@
...
+  const allowNegative = min && min < 0;
+  // We get the decimal separator by formatting a arbitrary number, and then extracting the decimal separator
+  const decimalSeparator = numberFormat
+    .formatToParts(1.1)
+    .find(part => part.type === 'decimal').value;
+  const thousandSeparator = decimalSeparator === ',' ? '.' : ',';
+  const isAllowedToTypeDecimals = !step;
+
+  // This function makes sure that the user can't type a number that is lower than the min value
+  const isAllowed = values => {
```
sergei-maertens, line 43: this also doesn't really work yet how it should, try:

* setting the `min` prop to 3
* typing the value `12` in the input

so we'll have to search for the appropriate balance between preventing invalid input and validation :thinking: I don't have clear ideas about this at the moment, I'll get back to it on Friday.

sergei-maertens (changes requested):

PR #415, opened 2023-05-15
Take the basename/basepath into account for analytics

sergei-maertens:
Closes open-formulieren/open-forms#3082

react-router does not take this into account by default, but that results in the form slug being cut off otherwise.

### DIFF
--- a/src/hooks/usePageViews.js
+++ b/src/hooks/usePageViews.js
@@ -1,5 +1,7 @@
-import {useEffect, useRef} from 'react';
+import {useContext, useEffect, useRef} from 'react';
 import {useLocation} from 'react-router-dom';
+
+import {ConfigContext} from 'Context';
 
 const isDev = process.env.NODE_ENV === 'development';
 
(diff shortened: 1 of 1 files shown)


src/hooks/usePageViews.js
```diff
@@ -102,8 +105,13 @@ const usePageViews = async () => {
...
       return;
-    trackPageView(location, previousLocation);
-  }, [location, previousLocation]);
+    const fullPath = `${basePath}${location.pathname}`;
+    const fullPreviousPath = previousLocation && `${basePath}${previousLocation.pathname}`;
+    trackPageView(
+      {...location, pathname: fullPath},
+      previousLocation && {...previousLocation, pathname: fullPreviousPath}
+    );
+  }, [basePath, location, previousLocation]);
```
SilviaAmAm, line 114: Note from discussion: here the path is constructed manually instead of using `window.location` to avoid using globals and because the react router location can have additional useful state.

SilviaAmAm (approved):

sergei-maertens:
Backports: 9c1f3c241973ff25290c719be581f453196d181e and 75951e4873f9c379b3ad90c67bdc9c736f49025e

PR #416, opened 2023-05-17
[#2808] fix overflowing filenames

pi-sigma:
Closes open-formulieren/open-forms#2808

- Removed ellipsis and let filename wrap around
- Moved trashcan icon and filesize to separate line

### DIFF
--- a/src/formio/templates/file.ejs
+++ b/src/formio/templates/file.ejs
@@ -106,10 +106,12 @@
 {% ctx.statuses.forEach(function(status) { %}
   <div class="file {{status.status === 'error' ? ' has-error' : ''}}">
     <div class="row">
-      <div class="fileName col-form-label col-sm-10">
+      <div class="fileName col-form-label">
         <span class="fileName__file">{{{status.originalName}}}</span>
-        <i class="{{ctx.iconClass('trash-can')}}" ref="fileStatusRemove" title="{{ctx.t('remove')}}"></i>
       </div>
+    </div>
+    <div class="row">
+			<i class="{{ctx.iconClass('trash-can')}}" ref="fileStatusRemove" title="{{ctx.t('remove')}}"></i>
       <div class="fileSize col-form-label col-sm-2 text-right">{{ctx.fileSize(status.size)}}</div>
     </div>
     {% if (status.status === 'progress') { %}
--- a/src/scss/components/_file-upload.scss
+++ b/src/scss/components/_file-upload.scss
@@ -69,7 +69,6 @@ we don't have strict BEM naming here.
       }
       overflow-wrap: break-word;
     }
-
     .col-sm-2 {
       @include bootstrap-span(width, 2);
     }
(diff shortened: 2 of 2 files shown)


src/scss/components/_file-upload.scss
```diff
@@ -74,7 +74,7 @@ we don't have strict BEM naming here.
       @include bootstrap-span(width, 2);
     }
     .col-sm-10 {
-      @include bootstrap-span(width, 10);
+      max-width: 100%;
```
SilviaAmAm, line 77: I'm not sure if it's a good idea to change this: i think these CSS rules are here because the default formio uses bootstrap, so then the templates use these `col-sm-XX` to specify how many columns an element spans. 

I think if we don't want this styling, it might be better to remove the class from the HTML instead of changing its css (and if needed add a new class with other styling).

src/scss/components/_file-upload.scss
```diff
@@ -107,9 +107,9 @@ we don't have strict BEM naming here.
       align-items: center;
 
       &__file {
-        @include ellipsis;
+        overflow-wrap: break-word;
```
SilviaAmAm, line 110: fyi, `break-word` is deprecated: https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#syntax

You can do the same by using:
```
word-break: normal;
overflow-wrap: anywhere;
```

SilviaAmAm (changes requested):

src/scss/components/_file-upload.scss
```diff
@@ -107,21 +108,28 @@ we don't have strict BEM naming here.
...
+        word-break: normal;
+        overflow-wrap: anywhere;
         display: inline-block;
-        max-width: calc(100% - 2em);
+        width: 100%;
       }
+    }
+
+    .fileSize {
+      width: 50%;
```
SilviaAmAm, line 119: Why not 100%?

src/formio/templates/file.ejs
```diff
@@ -106,8 +106,12 @@
 {% ctx.statuses.forEach(function(status) { %}
   <div class="file {{status.status === 'error' ? ' has-error' : ''}}">
     <div class="row">
-      <div class="fileName col-form-label col-sm-10">
+      <div class="fileName col-form-label">
         <span class="fileName__file">{{{status.originalName}}}</span>
+      </div>
+    </div>
+    <div class="row">
+      <div>
```
SilviaAmAm, line 114: Is this extra div needed?

SilviaAmAm:
(The comments above are just questions, but the status of the review hasn't refreshed to show the 'comment' image instead of the 'requested changes' image :shrug: )

CharString (changes requested):
I'd like to see a story added to the storybook that shows this behaviour of handling large filenames. That way regressions can be monitored with chromatic.

sergei-maertens (approved):

PR #418, opened 2023-05-17
[open-formulieren/open-forms#3061] Feature: Select Field

Xaohs:
Closes open-formulieren/open-forms#3061

### DIFF
--- a/src/components/forms/SelectField/AsyncSelectField.js
+++ b/src/components/forms/SelectField/AsyncSelectField.js
@@ -1,29 +1,21 @@
+import PropTypes from 'prop-types';
 import React from 'react';
 import {useAsync} from 'react-use';
 
-import {SelectField} from './SelectField';
+import SelectField, {SelectFieldPropTypes} from './SelectField';
 
-const withOptions = WrappedComponent => {
-  return ({optionsRetriever, ...props}) => {
-    const {loading, error, value: options} = useAsync(optionsRetriever);
-
-    if (loading) {
-      return (
-        <WrappedComponent
-          options={[{label: 'Loading...', value: null}]} // Display loading message
-          {...props}
-        />
-      );
-    }
-
-    if (error) {
-      return <div>Error loading options: {error.message}</div>; // Display error message
-    }
-
-    return <WrappedComponent options={options} {...props} />;
-  };
+const AsyncSelectField = ({getOptions, ...props}) => {
+  const {loading, error, value: options} = useAsync(getOptions, [getOptions]);
+  // re-throw any error, it's up to the error boundary to handle it
+  if (error) {
+    throw error;
+  }
+  return <SelectField options={options || []} isLoading={loading} {...props} />;
 };
 
-const AsyncSelectField = withOptions(SelectField);
+AsyncSelectField.propTypes = {
+  ...SelectFieldPropTypes,
+  getOptions: PropTypes.func.isRequired,
+};
 
 export default AsyncSelectField;
--- a/src/components/forms/SelectField/AsyncSelectField.js
+++ b/src/components/forms/SelectField/AsyncSelectField.js
@@ -1,0 +1,29 @@
+import React from 'react';
+import {useAsync} from 'react-use';
+
+import {SelectField} from './SelectField';
+
+const withOptions = WrappedComponent => {
+  return ({optionsRetriever, ...props}) => {
+    const {loading, error, value: options} = useAsync(optionsRetriever);
+
+    if (loading) {
+      return (
+        <WrappedComponent
+          options={[{label: 'Loading...', value: null}]} // Display loading message
+          {...props}
+        />
+      );
+    }
+
+    if (error) {
+      return <div>Error loading options: {error.message}</div>; // Display error message
+    }
+
+    return <WrappedComponent options={options} {...props} />;
+  };
+};
+
+const AsyncSelectField = withOptions(SelectField);
+
+export default AsyncSelectField;
--- a/src/components/forms/SelectField/SelectField.js
+++ b/src/components/forms/SelectField/SelectField.js
@@ -1,0 +1,49 @@
+import {FormFieldDescription, FormLabel, Paragraph} from '@utrecht/component-library-react';
+import SelectAsync from 'react-select';
+
+import {getBEMClassName} from '../../../utils';
+
+export const SelectField = ({
+  label = '',
+  id = '',
+  isRequired = false,
+  description = '',
+  disabled = false,
+  invalid = false,
+  options = [],
+}) => {
+  const labelClassName = getBEMClassName('label', [isRequired && 'required'].filter(Boolean));
+
+  return (
+    <>
+      <Paragraph className={labelClassName}>
+        <FormLabel htmlFor={id}>{label}</FormLabel>
+      </Paragraph>
+      <SelectAsync
+        inputId={id}
+        name={id}
+        components={{DropdownIndicator: () => null, IndicatorSeparator: () => null}}
+        classNames={{
+          control: state => {
+            return getBEMClassName('rs-select', [
+              state.isFocused && 'focus',
+              disabled && 'disabled',
+              invalid && 'invalid',
+              isRequired && 'required',
+            ]);
+          },
+          menu: () => {
+            return getBEMClassName('rs-select-menu');
+          },
+          option: state => {
+            return getBEMClassName('rs-select-option', [state.isFocused && 'focus']);
+          },
+        }}
+        options={options}
+        unstyled
+        isDisabled={disabled}
+      ></SelectAsync>
+      {description && <FormFieldDescription invalid={invalid}>{description}</FormFieldDescription>}
+    </>
+  );
+};
--- a/src/components/forms/SelectField/SelectField.js
+++ b/src/components/forms/SelectField/SelectField.js
@@ -1,9 +1,12 @@
 import {FormFieldDescription, FormLabel, Paragraph} from '@utrecht/component-library-react';
-import SelectAsync from 'react-select';
+import {Field} from 'formik';
+import PropTypes from 'prop-types';
+import Select from 'react-select';
 
 import {getBEMClassName} from '../../../utils';
 
 export const SelectField = ({
+  name = '',
   label = '',
   id = '',
   isRequired = false,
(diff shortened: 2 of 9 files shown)


src/components/forms/SelectField/AsyncSelectField.js
```diff
@@ -0,0 +1,29 @@
...
+
+    if (error) {
+      return <div>Error loading options: {error.message}</div>; // Display error message
+    }
+
+    return <WrappedComponent options={options} {...props} />;
+  };
+};
+
+const AsyncSelectField = withOptions(SelectField);
```
sergei-maertens, line 27: this should specify the additional proptypes for the wrapper: `optionsRetriever: PropTypes.func.isRequired`

src/components/forms/SelectField/AsyncSelectField.js
```diff
@@ -0,0 +1,29 @@
...
+import {SelectField} from './SelectField';
+
+const withOptions = WrappedComponent => {
+  return ({optionsRetriever, ...props}) => {
+    const {loading, error, value: options} = useAsync(optionsRetriever);
+
+    if (loading) {
+      return (
+        <WrappedComponent
+          options={[{label: 'Loading...', value: null}]} // Display loading message
```
sergei-maertens, line 13: we don't need to handle this ourselves, instead we can pass `isLoading={loading}` to the underlying react-select, see: https://github.com/open-formulieren/formio-builder/blob/main/src/components/builder/prefill/plugin.tsx#L54

Similarly, all user-facing texts should be properly localized using react-intl, currently we support NL/EN but more languages will follow. You can use react-select's [`loadingMessage`](https://react-select.com/props) prop for this.

src/components/forms/SelectField/AsyncSelectField.js
```diff
@@ -0,0 +1,29 @@
...
+          {...props}
+        />
+      );
+    }
+
+    if (error) {
+      return <div>Error loading options: {error.message}</div>; // Display error message
+    }
```
sergei-maertens, line 21: "nice" in a way that it doesn't just crash, but ultimately unhelpful for the end-user because they have no way of correcting the problem.

So with that in mind, I'd favour just throwing the exception like [here](https://github.com/open-formulieren/formio-builder/blob/main/src/components/builder/prefill/plugin.tsx#L36) and let the error boundary deal with it in a generic way.

In the future, we may support an `onError` prop so that the calling component can define retry behaviour or what to do when loading fails. (so don't add this now).

src/components/forms/SelectField/SelectField.js
```diff
@@ -0,0 +1,69 @@
+import {FormFieldDescription, FormLabel, Paragraph} from '@utrecht/component-library-react';
+import {Field} from 'formik';
+import PropTypes from 'prop-types';
+import Select from 'react-select';
+
+import {getBEMClassName} from '../../../utils';
```
sergei-maertens, line 6:
```suggestion
import {getBEMClassName} from 'utils';
```

src/components/forms/SelectField/SelectField.js
```diff
@@ -0,0 +1,69 @@
...
+
+export const SelectField = ({
+  name = '',
```
sergei-maertens, line 9:
```suggestion
  name,
```
it's a required prop (due to how we use formik), so no point in providing a default.

src/components/forms/SelectField/SelectField.js
sergei-maertens, line 1: The styling is very different from our own theme: https://open-formulieren.github.io/open-forms-sdk/?path=/story/form-io-components-select--default-story&globals=locale:nl;locales.nl:Nederlands;locales.en:English

So let's break that out in a separate PR to tackle - we should set the values of design tokens of Utrecht component to get as close to our own theme as possible. Let's leave this PR for the functional side so you're unblocked for the product unit, but there is more to do here.

src/components/forms/SelectField/SelectField.js
```diff
@@ -0,0 +1,69 @@
...
+  isRequired: PropTypes.bool,
+  description: PropTypes.string,
+  disabled: PropTypes.bool,
+  invalid: PropTypes.bool,
+  options: PropTypes.arrayOf(
+    PropTypes.shape({
+      label: PropTypes.string,
+      value: PropTypes.string,
+    })
+  ).isRequired,
```
sergei-maertens, line 68: this is a bit too constrained, see https://github.com/open-formulieren/formio-builder/blob/main/src/components/formio/select.tsx#LL40C3-L40C16

-> you can pass the `getOptionValue` prop to react-select so that it supports any option shape, really. I think you can only really demand that `options: PropTypes.array,` and there isn't much more to say about that.

src/components/forms/SelectField/SelectField.stories.mdx
sergei-maertens, line 1: Please add some proze to this documentation as well - explain when to use the `SelectField`, give examples on how to use the `AsyncSelectField` etc.

src/components/forms/SelectField/SelectField.stories.mdx
```diff
@@ -0,0 +1,80 @@
+import {Canvas, Meta, Story} from '@storybook/addon-docs';
+import {expect} from '@storybook/jest';
+import {userEvent, within} from '@storybook/testing-library';
+
+import {FormikDecorator} from '../../../story-utils/decorators';
```
sergei-maertens, line 5:
```suggestion
import {FormikDecorator} from 'story-utils/decorators';
```

src/components/forms/SelectField/SelectField.stories.mdx
```diff
@@ -0,0 +1,80 @@
...
+import AsyncSelectField from './AsyncSelectField';
+import {SelectField} from './SelectField';
```
sergei-maertens, line 7:
```suggestion
import SelectField from './SelectField';
```

src/scss/components/_rs-select.scss
sergei-maertens, line 1: There are too many hardcoded values/styles here. Open Forms is a white label product that can be styled using design tokens. For example, the form.io select component has the following CSS rules:

image

Note the use of the `var(--of-color-border)`, which is configurable in the admin environment. To be picked up in the separate, styling-related PR.

sergei-maertens (changes requested):
Can you please create a new issue (in this repo) for the select-styling related remarks?

src/components/forms/SelectField/AsyncSelectField.js
```diff
@@ -1,29 +1,25 @@
...
   };
 };
 
 const AsyncSelectField = withOptions(SelectField);
-
+AsyncSelectField.propTypes =
+  SelectField.propTypes &
+  {
+    optionsRetriever: PropTypes.func.isRequired,
+  };
```
sergei-maertens, line 24: I had to fix this in main for the Email component: d50e2f6ad2155e9bf0d77b24c08eab44e3233737 - I suspect that the `Foo.propTypes` get stripped out in production builds and it causes those storybook stories to crash.

You can define the common/shared props in a variable and use that to reduce the duplication, but getting them from `SelectField.propTypes` is not reliable.

src/components/forms/SelectField/AsyncSelectField.js
```diff
@@ -0,0 +1,25 @@
...
+
+const withOptions = WrappedComponent => {
+  return ({optionsRetriever, ...props}) => {
+    const {loading, error, value: options} = useAsync(optionsRetriever);
+
+    if (error) {
+      throw Error;
+    }
+
+    return <WrappedComponent options={options} isLoading={loading} {...props} />;
```
sergei-maertens, line 15: I'm still missing the localized `loadingMesage` prop for react-select

src/components/forms/SelectField/SelectField.stories.mdx
```diff
@@ -0,0 +1,96 @@
...
+      test: 'test',
+    },
+  }}
+/>
+
+# Select Field
+
+We should make use of the SelectField component whenever we have `options` that are statically
+available, and we don't need to retrieve them from an API endpoint. Whenever we **do** need to
+retrieve the options from an API endpoint, we should instead use the AsyncSelectField component.
```
sergei-maertens, line 44:
```suggestion
Use the `SelectField` component whenever you have `options` that are known upfront. If your
options are retrieved from some external data source (like an API endpoint) instead, use the 
`AsyncSelectField` variant instead.
```

- address the reader
- imperative form makes it easier to extract what to do when, as a reader

src/components/forms/SelectField/SelectField.stories.mdx
```diff
@@ -0,0 +1,96 @@
...
+    }}
+  >
+    {Template.bind({})}
+  </Story>
+</Canvas>
+
+## AsyncSelectField
+
+When using this component, we provide a `optionsRetriever` callback that will be used to retrieve
+the options. This callback should return a promise that resolves to an array of options.
```
sergei-maertens, line 75:
```suggestion
When using this component, provide an `optionsRetriever` callback that will be used to retrieve
the options. This callback should be an async function returning an array of options.
```

async/await is the easier-to-understand abstraction, so while not incorrect, it's advisable to steer people towards using the modern syntax instead of letting them think about promises.

sergei-maertens (changes requested):

src/components/forms/SelectField/SelectField.js
```diff
@@ -0,0 +1,87 @@
...
+const SelectField = ({
+  name,
+  label = '',
+  id = '',
+  isRequired = false,
+  description = '',
+  disabled = false,
+  invalid = false,
+  options = [],
+  valueProperty = 'value',
```
sergei-maertens, line 18: I copied the implementation from the form-builder repository, providing a higher-level API to the consumers of these components instead of having to define callback functions everywhere.

src/components/forms/SelectField/SelectField.js
```diff
@@ -0,0 +1,87 @@
...
+            description="(Async) select options loading message"
+            defaultMessage="Loading..."
+          />
+        )}
+        noOptionsMessage={() => (
+          <FormattedMessage
+            description="Select 'no options' message"
+            defaultMessage="No results found"
+          />
+        )}
```
sergei-maertens, line 63: Added the localized options for these messages

src/components/forms/SelectField/SelectField.js
```diff
@@ -0,0 +1,87 @@
...
+        {...props}
```
sergei-maertens, line 64: react-select is a bit of a leaky abstraction, but it's nice that you can easily pass a `isClearable` prop from top-level component and have them forwarded, so I opted to just forward any additional props.

PR #419, opened 2023-05-17
Feature/product unit Implemented.

Xaohs:
closes open-formulieren/open-forms#3062

### DIFF
--- a/src/components/appointments/AppointmentProduct.stories.mdx
+++ b/src/components/appointments/AppointmentProduct.stories.mdx
@@ -1,0 +1,14 @@
+import {Canvas, Meta, Story} from '@storybook/addon-docs';
+
+import {FormikDecorator} from '../../story-utils/decorators';
+import AppointmentProduct from './AppointmentProduct';
+
+export const Template = args => <AppointmentProduct {...args} namePrefix={'test'} />;
+
+<Meta title="Private API / Appointments / Appointment Product" component={AppointmentProduct} />
+
+<Canvas>
+  <Story name="Static" decorators={[FormikDecorator]}>
+    {Template.bind({})}
+  </Story>
+</Canvas>
--- a/src/components/appointments/AppointmentProduct.stories.mdx
+++ b/src/components/appointments/AppointmentProduct.stories.mdx
@@ -1,4 +1,6 @@
 import {Canvas, Meta, Story} from '@storybook/addon-docs';
+import {expect} from '@storybook/jest';
+import {within} from '@storybook/testing-library';
 
 import {FormikDecorator} from '../../story-utils/decorators';
 import AppointmentProduct from './AppointmentProduct';
@@ -8,7 +10,17 @@ export const Template = args => <AppointmentProduct {...args} namePrefix={'test'
 <Meta title="Private API / Appointments / Appointment Product" component={AppointmentProduct} />
 
 <Canvas>
-  <Story name="Static" decorators={[FormikDecorator]}>
+  <Story
+    name="Static"
+    decorators={[FormikDecorator]}
+    play={async ({canvasElement}) => {
+      const canvas = within(canvasElement);
+      //Ensure the Select Field is rendered
+      await expect(canvas.getByRole('combobox')).toBeVisible();
+      // Ensure the Text Field is rendered
+      await expect(canvas.getByRole('textbox')).toBeVisible();
+    }}
+  >
     {Template.bind({})}
   </Story>
 </Canvas>
--- a/src/components/forms/NumberField/NumberField.js
+++ b/src/components/forms/NumberField/NumberField.js
@@ -1,0 +1,83 @@
+import {
+  FormFieldDescription,
+  FormLabel,
+  Paragraph,
+  Textbox,
+  FormField as UtrechtFormField,
+} from '@utrecht/component-library-react';
+import {useField} from 'formik';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import {getBEMClassName} from 'utils';
+
+const NumberField = ({
+  name,
+  min,
+  step,
+  label = '',
+  isRequired = false,
+  description = '',
+  id = '',
+  disabled = false,
+  invalid = false,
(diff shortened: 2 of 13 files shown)


src/components/appointments/AppointmentProduct.stories.mdx
```diff
@@ -0,0 +1,31 @@
...
+import AppointmentProduct from './AppointmentProduct';
+
+export const Template = args => <AppointmentProduct {...args} />;
+
+<Meta title="Private API / Appointments / Appointment Product" component={AppointmentProduct} />
+
+<Canvas>
+  <Story
+    name="Static"
+    decorators={[FormikDecorator]}
```
sergei-maertens, line 15: I prefer the decorator being specified on the `<Meta>` component so it automatically applies to all stories.

src/components/forms/ProductSelect/ProductSelect.js
sergei-maertens, line 1: This entire file/component belongs in `src/components/appointments` - the component is not intended to be generic/re-usable like TextField/SelectField/NumberField etc.

src/components/forms/ProductSelect/ProductSelect.js
```diff
@@ -0,0 +1,51 @@
+import PropTypes from 'prop-types';
+
+import AsyncSelectField from '../SelectField/AsyncSelectField';
```
sergei-maertens, line 3:
```suggestion
import {AsyncSelectField} from 'components/forms';
```
this is possible with the version in `main` now :)

src/components/forms/ProductSelect/ProductSelect.js
```diff
@@ -0,0 +1,51 @@
...
+            name: 'Product 2',
+          },
+          {
+            code: 'P3',
+            identifier: 'product3',
+            name: 'Product 3',
+          },
+        ]);
+      }, 1000);
+    });
```
sergei-maertens, line 28: The mocking should be done in storybook, not in the component. For the component, you need to implement the real API call.

Please review the [developer documentation](https://open-formulieren.github.io/open-forms-sdk/?path=/story/developers-documentation--page#mocking-in-stories) and check an [example](https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/LanguageSelection/LanguageSelection.stories.mdx?plain=1#L61)

And an example of the API call being made: https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/LanguageSelection/LanguageSelection.js#L39

src/components/forms/ProductSelect/ProductSelect.js
```diff
@@ -0,0 +1,51 @@
...
+    return response.map(({code, identifier, name}) => ({
+      label: `${code} - ${name}`,
+      value: identifier,
+    }));
```
sergei-maertens, line 32: We don't need to remap this, you can instead specify `valueProperty="identifier"` on the `<AsyncSelect>` component.

For the label, I think only `name` is relevant, looking at the existing dropdown:

image

src/components/forms/SelectField/SelectField.js
```diff
@@ -68,6 +85,14 @@ const SelectField = ({
           />
         )}
         {...props}
+        onChange={newValue => {
+          const isSingle = !Array.isArray(newValue);
+          const normalized = isSingle ? [newValue] : newValue;
+          const rawValues = normalized.map(val => val?.[valueProperty] ?? null);
```
CharString, line 91: The `?? null` turns `undefined` into `null`?

src/components/forms/SelectField/SelectField.js
```diff
@@ -68,6 +85,14 @@ const SelectField = ({
           />
         )}
         {...props}
+        onChange={newValue => {
```
CharString, line 88: Isn't `newValue` and event?

src/components/appointments/Product.stories.mdx
```diff
@@ -0,0 +1,72 @@
...
+      await userEvent.type(canvas.getByLabelText('Product'), 'Rijbewijs');
+      await userEvent.click(await canvas.findByText('Rijbewijs aanvraag (Drivers license)'));
+      await expect(canvas.queryByText('Paspoort aanvraag')).toBeNull();
+      await expect(await canvas.findByText('Rijbewijs aanvraag (Drivers license)')).toBeVisible();
+      const amountInput = canvas.getByLabelText('Amount');
+      await userEvent.clear(amountInput);
+      await userEvent.type(amountInput, '1');
+      await expect(amountInput).toHaveDisplayValue('1');
+    }}
+  >
```
CharString, line 69: Maybe it's me, but I was confused by the visual state these interactions leave the component in. It wasn't clear to me that it wasn't untouched.

src/components/forms/NumberField/NumberField.js
```diff
@@ -81,7 +83,7 @@ NumberField.propTypes = {
   useNumberType: PropTypes.bool,
   min: PropTypes.number,
   step: PropTypes.number,
-  label: PropTypes.string,
+  label: PropTypes.node,
```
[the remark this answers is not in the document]
sergei-maertens: no, `node` covers strings, null, undefined and react elements such as `<FormattedMessage ... />` and fragments `<>...</>`

src/components/forms/SelectField/SelectField.js
sergei-maertens: Yep, while leaving other falsy values like `''` and `0` untouched.

src/components/forms/SelectField/SelectField.js
sergei-maertens: it's not because of the API of react-select. it doesn't dispatch any events, but instead handles its internal state that you can tap into via certain callbacks.

image

src/components/appointments/Product.stories.mdx
sergei-maertens: I struggle with that too - perhaps a `Default` story without any tests makes it more obvious?

PR #421, opened 2023-05-19
Number field improvements

sergei-maertens:
I went ahead and explored some ideas - I'm relatively satisfied with these though I still have some concerns. See the comments/annotations on the PR itself.

### DIFF
--- a/src/components/forms/NumberField/NumberField.js
+++ b/src/components/forms/NumberField/NumberField.js
@@ -13,10 +13,19 @@ import {NumericFormat} from 'react-number-format';
 
 import {getBEMClassName} from 'utils';
 
+const getSeparators = locale => {
+  const numberFormat = new Intl.NumberFormat(locale);
+  const parts = numberFormat.formatToParts(1000.1);
+  const decimalSeparator = parts.find(part => part.type === 'decimal').value;
+  const thousandSeparator = parts.find(part => part.type === 'group').value;
+  return {decimalSeparator, thousandSeparator};
+};
+
 const NumberField = ({
   name,
   min,
   step,
+  useNumberType = false, // number type can be convenient for small numbers with the +- spinner buttons
   label = '',
   isRequired = false,
   description = '',
@@ -27,27 +36,12 @@ const NumberField = ({
   const [field, meta, helpers] = useField(name);
   const {locale} = useIntl();
 
-  const numberFormat = new Intl.NumberFormat(locale);
   const labelClassName = getBEMClassName('label', [isRequired && 'required'].filter(Boolean));
   const inputClassName = getBEMClassName('input', [invalid && 'invalid'].filter(Boolean));
 
-  const allowNegative = min && min < 0;
   // We get the decimal separator by formatting a arbitrary number, and then extracting the decimal separator
-  const decimalSeparator = numberFormat
-    .formatToParts(1.1)
-    .find(part => part.type === 'decimal').value;
-  const thousandSeparator = numberFormat
-    .formatToParts(1000)
-    .find(part => part.type === 'group').value;
-  const isAllowedToTypeDecimals = !step || !Number.isInteger(step);
-
-  // This function makes sure that the user can't type a number that is lower than the min value
-  const isAllowed = values => {
-    const {formattedValue, floatValue} = values;
-    const isEmpty = formattedValue === '';
-    const isBelowMin = !isNaN(min) ? floatValue >= min : true;
-    return isEmpty || isBelowMin;
-  };
+  const {decimalSeparator, thousandSeparator} = getSeparators(locale);
+  const isInteger = step != null && Number.isInteger(step);
 
   const inputProps = {
     // To handle the State in formik
@@ -62,15 +56,18 @@ const NumberField = ({
     invalid,
     min,
     step,
-    type: 'text',
+    type: useNumberType ? 'number' : 'text',
 
     // These are for the NumericFormat component
     customInput: Textbox,
-    isAllowed,
-    allowNegative,
-    decimalSeparator,
-    thousandSeparator,
-    decimalScale: isAllowedToTypeDecimals ? 2 : 0,
+    // only pass the localized separators when we're not using type="number" input
+    ...(!useNumberType
+      ? {
+          decimalSeparator,
+          thousandSeparator,
+          decimalScale: isInteger ? undefined : 2,
+        }
+      : {}),
   };
 
   return (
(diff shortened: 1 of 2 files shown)


src/components/forms/NumberField/NumberField.js
```diff
@@ -13,10 +13,19 @@ import {NumericFormat} from 'react-number-format';
 
 import {getBEMClassName} from 'utils';
 
+const getSeparators = locale => {
```
sergei-maertens, line 16: Broke this out into a standalone function to de-clutter a bit.

src/components/forms/NumberField/NumberField.js
```diff
@@ -13,10 +13,19 @@ import {NumericFormat} from 'react-number-format';
...
+  const decimalSeparator = parts.find(part => part.type === 'decimal').value;
+  const thousandSeparator = parts.find(part => part.type === 'group').value;
+  return {decimalSeparator, thousandSeparator};
+};
+
 const NumberField = ({
   name,
   min,
   step,
+  useNumberType = false, // number type can be convenient for small numbers with the +- spinner buttons
```
sergei-maertens, line 28: Basically the idea is: the place where the `NumberField` would be used knows best whether `type="number"` or free-form input is best, so let's provide a prop instead of guessing that from other props.

E.g. the product unit would definitely used `<NumberField name="amount" useNumberType min={1} step={1} />` because it's about a discrete number of products you're ordering.

src/components/forms/NumberField/NumberField.js
```diff
@@ -62,15 +56,18 @@ const NumberField = ({
...
-    thousandSeparator,
-    decimalScale: isAllowedToTypeDecimals ? 2 : 0,
+    // only pass the localized separators when we're not using type="number" input
+    ...(!useNumberType
+      ? {
+          decimalSeparator,
+          thousandSeparator,
+          decimalScale: isInteger ? undefined : 2,
+        }
+      : {}),
```
sergei-maertens, line 70: I noticed I couldn't enter `1000` when using a number input, the localized formatting was probably breaking stuff, so the idea is to only leverage that functionality when we're using a text-based input component.

src/components/forms/NumberField/NumberField.stories.mdx
```diff
@@ -26,7 +27,14 @@ const generalTest = async canvasElement => {
 <Meta
   title="Pure React Components / Forms / NumberField"
   component={NumberField}
+  decorators={[FormikDecorator]}
```
sergei-maertens, line 30: we can specify the decorators at the meta level instead of repeating it for every component, bit more DRY as it applies to every story anyway.

src/components/forms/NumberField/NumberField.stories.mdx
```diff
@@ -26,7 +27,14 @@ const generalTest = async canvasElement => {
...
   parameters={{
+    docs: {
+      source: {
+        type: 'dynamic',
+        excludeDecorators: true,
+      },
+    },
```
sergei-maertens, line 37: This leads to useful code previews in the "Docs" tab:

image

src/components/forms/NumberField/NumberField.stories.mdx
```diff
@@ -35,22 +43,118 @@ const generalTest = async canvasElement => {
   }}
 />
 
+# Number inputs
```
sergei-maertens, line 46: I've tried to summarize our back-and-forth thoughts/discussions in some proper documentation with recommendations on when to use what.

src/components/forms/NumberField/NumberField.stories.mdx
```diff
@@ -35,22 +43,118 @@ const generalTest = async canvasElement => {
...
+      const input = canvas.getByLabelText('Amount');
+      // assert that we can still type 'large' numbers
+      await userEvent.clear(input);
+      await userEvent.type(input, '1000');
+      await expect(input).toHaveDisplayValue('1000');
+      // assert that we can type floats without localisation
+      await userEvent.clear(input);
+      await userEvent.type(input, '1.5');
+      await expect(input).toHaveDisplayValue('1.5');
+    }}
```
sergei-maertens, line 112: Note that I didn't apply the `generalTest` thing here.

I am personally of the philosophy that tests should be 100% self-contained and calling out to subtests/functions etc. makes it harder to debug/check tests on what is going on.

Additionally, I typically rely on the `canvas.getByLabelText` to ensure that label + input are correctly linked, which then also implicitly proves that clicking the label focuses the input. So, for me, those `expect(...).toBeFocused()` tests do not need to written out, but it doesn't hurt to do so either.

src/components/forms/NumberField/NumberField.stories.mdx
```diff
@@ -35,22 +43,118 @@ const generalTest = async canvasElement => {
...
+## Free form number fields
+
+Free form number inputs allow entering arbitrary numbers - decimals, negative, large, small... They
+take the current locale into account and present formatted numbers. E.g. in the NL locale, the `.`
+decimal separator will be converted to a comma, while thousand groups are automatically formatted
+with a period.
+
+Only numerical keyboard entries are allowed.
+
+**TODO**: check accessibility impact of this!
```
sergei-maertens, line 127: Form.io does also block non-numerical inputs, but it also doesn't have a reputation of being the most accessible.

So, while the UX here is "good" for (able) keyboard users that you can't enter non-number characters, I do still have concerns about how this is experienced by people with assistive technology. @Robbert maybe you have some thoughts about this? You can check out the deployed storybook from the PR to try things out.

src/components/forms/NumberField/NumberField.stories.mdx
```diff
@@ -35,22 +43,118 @@ const generalTest = async canvasElement => {
...
+
+The developer/form designers are responsible for defining validation rules for the values to express
+minimum/maximum values.
```
sergei-maertens, line 130: Here I am anticipating that we'll add `zod` for the appointment form validation schema.

src/components/forms/NumberField/NumberField.stories.mdx
```diff
@@ -35,22 +43,118 @@ const generalTest = async canvasElement => {
...
+
+**Notes**
+
+- Props like `min` and `step` are ignored as they only apply to the `type="number"` inputs.
+
+- The interaction tests rely on the NL locale. TODO: figure out how to change the locale in play
+  functions.
```
sergei-maertens, line 137: I spent too much time trying to figure out if I can force the locale to EN within a play function, but it just doesn't seem possible... :(

PR #423, opened 2023-05-19
[OF#1530] Part 6 - Add cosign button on form start page

SilviaAmAm:
Fixes open-formulieren/open-forms#1530

Screenshot from 2023-05-31 11-56-31

### DIFF
--- a/src/components/LoginOptions/LoginOptions.stories.mdx
+++ b/src/components/LoginOptions/LoginOptions.stories.mdx
@@ -132,13 +132,14 @@ export const LiteralDecorator = (Story, {args}) => (
   </LiteralsProvider>
 );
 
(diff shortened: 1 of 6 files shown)


SilviaAmAm:
> Is there ever an array of co-sign login options?

Now no, the cosign component supports specifying only one login method. I can imagine that in the future we will be asked to extend it, but it should be straightforward then.

> And this is the button that, when pressed asks for the code that was e-mailed. Which makes me think about the story, when we, as signer and co-signer are together filling in the form and take turns on the same computer

This is the button that when clicked first redirects you to the auth plugin and then to the view where you can retrieve the submission. If both users are at the same computer, then maybe the confirmation email to the 'signer' and/or the template at the end of the form submission could mention that the reference can be used to retrieve the submission for cosign (so that the co-signer doesn't need to look at their email).

PR #424, opened 2023-05-22
[#2809] make report download link title configurable

pi-sigma:
Part of solution to open-formulieren/open-forms#2809
Complement to open-formulieren/open-forms#3108

- Changed `components.SubmissionConfirmation.js` to use dynamic title for download link

### DIFF
--- a/src/components/SubmissionConfirmation.js
+++ b/src/components/SubmissionConfirmation.js
@@ -15,6 +15,7 @@ import PaymentForm from 'components/PaymentForm';
 import {Toolbar, ToolbarList} from 'components/Toolbar';
 import usePoll from 'hooks/usePoll';
 import useTitle from 'hooks/useTitle';
+import Types from 'types';
 
 const RESULT_FAILED = 'failed';
 const RESULT_SUCCESS = 'success';
@@ -67,7 +68,7 @@ StartPayment.propTypes = {
  * @param {Function} onFailure Callback to invoke if the background processing result is failure.
  * @param {Function} onConfirmed Callback to invoke if the background processing result is success.
  */
-const SubmissionConfirmation = ({statusUrl, onFailure, onConfirmed}) => {
+const SubmissionConfirmation = ({statusUrl, onFailure, onConfirmed, form}) => {
   const intl = useIntl();
   const pageTitle = intl.formatMessage({
     description: 'Confirmation page title',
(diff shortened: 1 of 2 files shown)


CharString (commented):
Hard to judge if this is correct with the API not fleshed out. open-formulieren/open-forms#3108
This change assumes there's always a translated link title in the api response.

pi-sigma:
@CharString The global config has a default for the title ([link](https://github.com/open-formulieren/open-forms/blob/f4adc9a4d21d18fdce554eb2851628bbb856ce09/src/openforms/config/models.py#L71))

src/components/SubmissionConfirmation.js
```diff
@@ -160,10 +160,7 @@ const SubmissionConfirmation = ({statusUrl, onFailure, onConfirmed}) => {
         <Body>
           <FAIcon icon="download" aria-hidden="true" modifiers={['inline']} />
           <Anchor href={reportDownloadUrl} target="_blank" rel="noopener noreferrer">
-            <FormattedMessage
-              description="Download report PDF link title"
-              defaultMessage="Download PDF"
-            />
+            {form.submissionReportDownloadLinkTitle}
```
sergei-maertens, line 163: Let's take a conservative and backwards compatible approach here - try to get the value from the API endpoint, but if it's falsy, fall back to the existing `<FormattedMessage>` approach. I don't think this change warrants an SDK major version bump.

sergei-maertens (changes requested):

src/components/SubmissionConfirmation.js
```diff
@@ -67,7 +67,7 @@ StartPayment.propTypes = {
  * @param {Function} onFailure Callback to invoke if the background processing result is failure.
  * @param {Function} onConfirmed Callback to invoke if the background processing result is success.
  */
-const SubmissionConfirmation = ({statusUrl, onFailure, onConfirmed}) => {
+const SubmissionConfirmation = ({statusUrl, onFailure, onConfirmed, submission, form}) => {
```
sergei-maertens, line 70: `submission` prop is unused, and `form` prop should be added to the component prop types.

src/components/SubmissionConfirmation.js
```diff
@@ -140,6 +140,20 @@ const SubmissionConfirmation = ({statusUrl, onFailure, onConfirmed}) => {
...
+  try {
+    linkTitle = form.submissionReportDownloadLinkTitle;
+  } catch (error) {
+    linkTitle = (
+      <FormattedMessage
+        description="Download report PDF link title"
+        defaultMessage="Download PDF"
+      />
+    );
+  }
```
sergei-maertens, line 155: I don't think this will throw an error, since `form` is an object, `linkTitle` will just resolve to `undefined` if the attribute is not present.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #425, opened 2023-05-23
[open-formulieren/open-forms-sdk/issues/#417] Added Stylin…

### LINKED ISSUE #417: Styling inconsistencies between own & form.io components
Primarily looking at the e-mail component for this, but it applies to TextField/NumberField too.

- [ ] Font-weight of input box: in the form.io forms these are bolder

  **form.io component**
  image

  **NLDS component**
  image

- [ ] invalid styles are missing background color / border on the left
  
  **form.io component** 
  image

  **NLDS component**
  image

Xaohs:
Closes #417

styling for the `invalid` prop, across all fields. (NumberField, EmailField, TextField)

### DIFF
--- a/src/components/forms/NumberField/NumberField.js
+++ b/src/components/forms/NumberField/NumberField.js
@@ -8,7 +8,7 @@ import {
 import {useField} from 'formik';
 import PropTypes from 'prop-types';
 import React from 'react';
-import {useIntl} from 'react-intl';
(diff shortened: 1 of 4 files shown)


sergei-maertens:
I took a look at this and decided to start from scratch instead in #426 - it's a complex situation with the styling overrides and design tokens mechanism from NL DS, but we can get pretty far using only that!

PR #426, opened 2023-05-23
Display validation errors for component and style form field appropriately

### LINKED ISSUE #417: Styling inconsistencies between own & form.io components

sergei-maertens:
Closes #417
Closes #425 

Depends on https://github.com/open-formulieren/design-tokens/pull/20
Depends on https://github.com/open-formulieren/open-forms-sdk/pull/428

Changes

* Updated to latest version of Utrecht's react component library - this contains a number of fixes and features we need for correct behaviour
* Refactor: don't make invalid a boolean prop, but instead extract that information from the formik state/meta
* Theme with design tokens as much as possible (see PR in design tokens repo)
* Added own modifiers for Utrecht components for targetted CSS overrides


src/scss/nl-design-system-community.scss
```diff
@@ -1,5 +1,4 @@
 @import '@utrecht/components/dist/document/css/index.css';
-@import '@utrecht/components/dist/form-field/css/index.css';
 @import '@utrecht/components/dist/form-field-description/css/index.css';
```
sergei-maertens, line 2: this will probably need to be moved so that it's included in the prod bundle :warning:

src/components/forms/SelectField/SelectField.js
```diff
@@ -63,8 +69,9 @@ const SelectField = ({
         )}
         {...props}
       />
-      {description && <FormFieldDescription invalid={invalid}>{description}</FormFieldDescription>}
-    </>
+      {description && <FormFieldDescription>{description}</FormFieldDescription>}
```
SilviaAmAm, line 72: For the open-forms theme, the padding above the description of the select is not quite consistent with the other components:
Screenshot from 2023-05-26 11-32-07

src/scss/components/_form-field.scss
```diff
@@ -0,0 +1,20 @@
+@use 'microscope-sass/lib/bem';
+
+@import '@utrecht/components/dist/form-field/css/index.css';
+
+// Overrides of the utrecht form field styles for our own theme
+.openforms-theme {
+  .utrecht-form-field {
+    @include bem.modifier('openforms') {
+      background-color: var(--of-utrecht-form-field-background-color, unset);
```
SilviaAmAm, line 9: Why have both "of" and "utrecht" in the token?

src/components/forms/SelectField/SelectField.js
sergei-maertens: yeah the `Description` styling itself is also not consistent yet, I was planning to tackle that in another issue/PR and wanted to focus this on the validation errors (styling).

src/scss/components/_form-field.scss
sergei-maertens: `of` for our own namespace, as it's a custom design token that doesn't exist for the utrecht component, and the `utrecht` to relate it back to the `utrecht-form-field` component

PR #428, opened 2023-05-24
Upgrade to React 18 in dependencies

sergei-maertens:
The React component libraries for NL DS make use of `React.useId` which is React 18+.

**Analysis whether to upgrade or not**

Alternatives explored:

* get the component libraries to distribute individual components instead of a bundle ( :red_circle: lot of work in build tooling)
* (temporarily) use something else for the `React.useId` usage (:red_circle: not desirable - what is the policy on supporting old versions etc.)

Risks:

* Unsure how this plays with potential breaking changes and old API usage in react-formio
* React.StrictMode + our `useEffect` usage may reveal a number of bugs
* Effect on consumers of our NPM package -> should be minimal, as we only use the React API's internally and otherwise there are no breaking changes

-> upgrade to React 18 turns out to be okay-ish, bigger upgrade was react-router from v5 -> v6 but that all seems to work as expected.

**Manual tests performed**

* Configured a form with two steps, started it from a clean session, filled out step 1, step 2, submitted + verified I could download the confirmation PDF
* Added an appointment step to the form (step 3) using the demo plugin, filled out the appoinment details, submitted. In the admin, clicked the "cancel appointment" link and verified it redirects back to the frontend correctly (see also open-formulieren/open-forms#3115 for that). Cancelled appointment & correctly got confirmation screen.
* Clicked through storybook - there are some router-related stories that are broken at the moment (Private API / Form Step).

I did not test the "new" appointment stuff since that's very much in Flux and known to be broken-ish right now, but this is not in production so it's acceptable.

**TODO**

- [x] Fix `Private API / FormStep` stories


src/components/Link.js
```diff
@@ -0,0 +1,35 @@
...
+    replace,
+    state,
+    target,
+  });
+  return (
+    <Anchor
+      {...rest}
+      href={href}
+      onClick={event => {
+        onClick?.(event);
```
SilviaAmAm, line 21: I didn't know this syntax :exploding_head:

src/formio/components/Map.js
SilviaAmAm, line 1: I tried testing a form with a map component and it seems to give an error when going to the summary page.

In the same form I had a cosign component and I noticed that if I select a place on the map and then I cosign, upon returning to the form the map is not pointing to the right place, even though the data is in the submission data. I haven't tried on main if this is a problem too!

SilviaAmAm (commented):
I manually tried the following things:
- Login / logout with demo BSN
- Ooievaarspas (files, files inside repeating groups, conditions, ...)
- Continue existing submission / abort submission buttons
- Pausing form

I didn't test resuming form because it seems that the email with the link is broken? So we should look into that separately :sweat_smile:  I'll create an issue for it

src/components/Link.js
sergei-maertens: me neither - copy pasted it straight from the react-router documentation!

src/components/appointments/AppointmentForm.js
```diff
@@ -175,31 +175,20 @@ const AppointmentForm = ({form}) => {
...
+      <Route path="bevestiging" element={<AppointmentConfirmation />} />
+      <Route path="" element={<AppointmentStep form={form} onSubmit={onAppointmentSubmit} />} />
+    </Routes>
   );
 
   const FormDisplayComponent = config?.displayComponents?.form ?? FormDisplay;
   return (
     <FormDisplayComponent
-      router={router}
+      router={<ErrorBoundary useCard>{router}</ErrorBoundary>}
```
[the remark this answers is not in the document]
sergei-maertens: no particular reason, I just moved things around without thinking to much. This `AppointmentForm` code is not used in prod right now anyway, and will change in the next few weeks so I'll tackle that as part of appointments work.

In other situations sometimes the error boundary should not be wrapped in a card, and sometimes it does, depending on the route, so for those you'd specify it per route instead of wrapping the entire component.

That said - there's also an argument to put your error boundary only around your own code/components and not around the third party library, as it could mask bugs in the router library.

src/components/Link.js
```diff
@@ -0,0 +1,35 @@
...
+        }
+      }}
+      ref={ref}
+      target={target}
+    />
+  );
+});
+
+// Prop types deliberately unspecified, please use the typescript definitions of
+// react-router-dom instead.
```
CharString, line 33: :ok:  :slightly_smiling_face:

src/formio/components/Map.js
sergei-maertens: I think map components are in a weird state right now, see also https://github.com/open-formulieren/open-forms/issues/2605 and the comment there:

> Maps will be revamped with upcomming user stories. Leaving this issue to be solved with those.

So I think we can ignore this?

src/components/FormStep/FormStep.stories.mdx
```diff
@@ -56,6 +56,7 @@ export const Template = args => {
       onLogicChecked={() => {}}
       onStepSubmitted={() => alert('step submitted')}
       onLogout={() => alert('logged out')}
+      onSessionDestroyed={() => alert('session destroyed')}
```
CharString, line 59: Did your TS compiler/some linter point out this was missing?

src/components/PaymentOverview/__snapshots__/test.spec.js.snap
```diff
@@ -75,8 +75,12 @@ exports[`on valid redirect renders an error message on failure 1`] = `
         </div>
       </div>
     </div>
-    <a>
-      some link
+    <a
+      className="utrecht-link utrecht-link--openforms"
+      href="/"
+      onClick={[Function]}
```
CharString, line 81: I'll try not to try to understand snapshots. I'm just happy this gets recorded, I would not have been able to come up with this assertion.

src/setupTests.js
```diff
@@ -6,6 +6,8 @@ import '@testing-library/jest-dom';
 
 import mswServer from 'api-mocks/msw-server';
 
+globalThis.IS_REACT_ACT_ENVIRONMENT = true;
```
CharString, line 9: Where is this referenced? One of the testing libraries?

Edit: yes https://react.dev/blog/2022/03/08/react-18-upgrade-guide

src/sdk.spec.js
```diff
@@ -100,11 +100,10 @@ describe('OpenForm', () => {
...
     });
 
-    await act(async () => {
-      await form.init();
-      await waitForElementToBeRemoved(() => within(formRoot).getByRole('status'));
-    });
+    await act(async () => await form.init());
 
+    // wait for the loader to be removed when all network requests have completed
+    await waitForElementToBeRemoved(() => within(formRoot).getByRole('status'));
```
CharString, line 106: I don't understand this. Does the `init` promise resolve before api calls are completed?

CharString (approved):
Looks fine. Just my regular :dizzy_face: puzzled question when dealing with JS.

package.json
```diff
@@ -123,9 +123,9 @@
     "@storybook/testing-library": "^0.0.13",
     "@svgr/webpack": "^5.5.0",
     "@testing-library/dom": ">=7.21.4",
-    "@testing-library/jest-dom": "^5.11.4",
-    "@testing-library/react": "^11.1.0",
-    "@testing-library/user-event": "^12.1.10",
+    "@testing-library/jest-dom": "^5.14.1",
+    "@testing-library/react": "^13.0.0",
+    "@testing-library/user-event": "^13.2.1",
```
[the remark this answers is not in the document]
sergei-maertens: Yep, I checked with the template repo of create-react-app. I don't like that this is modifying the ejected version without tackling all the build tooling, but I'm hopeful that after the formio-renderer work we can start migrating to a TypeScript toolchain for the SDK and then we should be able to shed a bunch of CRA stuff.

src/components/App.js
```diff
@@ -40,18 +40,13 @@ const App = ({...props}) => {
...
   const router = (
-    <Switch>
+    <Routes>
       {/* Anything dealing with appointments gets routed to it's own sub-router */}
-      <Route path="/afspraak*" component={ManageAppointment} />
-      <Route path="/appointment">
-        <AppointmentForm {...props} />
-      </Route>
-
+      <Route path="afspraak-annuleren/*" element={<ManageAppointment />} />
```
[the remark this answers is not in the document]
sergei-maertens: The actual URLs have not changed, however react-router v6 now properly supports nesting routes. The wildcard must follow a trailing slash, so I had to refactor the routes a bit.

src/components/PaymentOverview/__snapshots__/test.spec.js.snap
sergei-maertens: yeah I check the output, verified that it's correct and the snapshot just needs updating, which is done via `jest` flags/interactive session, so neither do I.

src/setupTests.js
sergei-maertens: yep, was part of the official upgrade docs. I find it odd that I can't find this in their CRA template, but yeah :shrug:

src/sdk.spec.js
sergei-maertens: I think something changes with the `createRoot(...)` and then the `render` after of React 18.

In general, dealing with the `useAsync` effects is nasty in this upgrade, there are many warnings about state updates happening outside of the `act` calls, but I spent 2 hours trying to fix/understand them and I just don't see how it can reasonably be done. I think the proper path forward here is to get away from the `react-dom/test-utils` and instead use `testing-library/react` which is much more interaction focused anyway.

In general I find the `act` usage confusing as **** and it's one of the reasons I prefer writing storybook interaction tests :grimacing:

SilviaAmAm (approved):

src/sdk.spec.js
CharString: Could it be the way we setup and teardown?  I don't know.. Ship it. :wink:

PR #438, opened 2023-06-07
Implement routing for appointment creation form

sergei-maertens:
(Partially) closes open-formulieren/open-forms#3067

I'm wrapping up this PR to not make it bigger than it already is, but more will follow when the remaining steps are to be implemented.


src/components/appointments/ChooseProductStep.js
```diff
@@ -1,77 +1,134 @@
...
-              // to make the key guaranteed unique
-              <div key={`${productId}-${index}`} className={getBEMClassName('editgrid__group')}>
-                <div className={getBEMClassName('editgrid__group-label')}>
-                  <FormattedMessage
-                    description="Appointments: single product label/header"
-                    defaultMessage="Product {number}/{total}"
-                    values={{number: index + 1, total: numProducts}}
-                  />
-                </div>
+    <Card title={formName} titleComponent="h1" modifiers={['mobile-header-hidden']}>
```
sergei-maertens, line 39: TODO: check if this can be moved into the parent component instead, as it applies to all steps.

src/components/appointments/ChooseProductStep.js
```diff
@@ -1,77 +1,134 @@
...
+        canSubmitStep={isStepValid(values)}
+        canSubmitForm={SUBMISSION_ALLOWED.yes}
+        canSuspendForm={false}
+        isAuthenticated={false} // TODO -> if authenticated (for prefill), logout must be shown
+        isLastStep={false}
+        isCheckingLogic={false}
+        loginRequired={false}
+        onFormSave={() => {}}
+        onLogout={() => {}}
+      />
```
sergei-maertens, line 125: *probably* better to use the primitives here instead of passing in a lot of irrelevant props. All we need is simple back links and submit buttons.

src/components/appointments/CreateAppointment.js
```diff
@@ -0,0 +1,202 @@
...
+  const currentStepIndex = APPOINTMENT_STEP_PATHS.indexOf(currentStep);
+  const steps = APPOINTMENT_STEPS.map(({path, name, validCheck}) => {
+    const index = APPOINTMENT_STEP_PATHS.indexOf(path);
+    const previousStepIndex = Math.max(index - 1, 0);
+    const previousStepCompleted = APPOINTMENT_STEPS[previousStepIndex].validCheck(values);
+    const isValid = validCheck(values);
+    return {
+      uuid: `appointments-${path}`,
+      href: path,
+      isCompleted: previousStepCompleted && isValid,
```
SilviaAmAm, line 95: I noticed in storybook that now when you select the product, the step in the progress indicator automatically gets checked (marked as complete). This is a bit inconsistent with how it is in the form?

src/components/appointments/LocationAndTimeStep.js
```diff
@@ -24,15 +48,77 @@ const LocationAndTimeStep = () => {
...
+  return (
+    <>
+      <Heading3 className="utrecht-heading-3--distanced">
+        <FormattedMessage
+          description="Product summary on appointments location and time step heading"
+          defaultMessage="Your products"
+        />
+      </Heading3>
+      <UnorderedList className="utrecht-unordered-list--distanced">
+        {products.map(({product, amount}, index) => (
```
SilviaAmAm, line 105:
```suggestion
        {products.map(({productId, amount}, index) => (
```
Maybe clearer?

src/components/appointments/CreateAppointment.js
sergei-maertens: I tried tweaking this around but it was odd either way - we don't really have an explicit submit event because all the state is just kept in memory and not synced to the backend (yet).

Let's discuss this after the 2.2 release

src/components/appointments/LocationAndTimeStep.js
sergei-maertens: the data structure is `products: [{product: 'some-id', amount: 42}]`, so deconstructing this would require `({product: productId, amount}, index)` in that case and I didn't find that particularly more readable, but either is fine.

I probably could also change the field name to be `productId` rather than `product` to clear things up?

sergei-maertens: `product` is indeed the ID - note that the arrow function destructures the data already:

```js
(entry, index) => {...}
```
with `entry` an item of the `products` array.

so this destructures into:

```js
({product, amount}, index) => {...}
```

SilviaAmAm (approved):

PR #439, opened 2023-06-07
[OF#3096] Blocked form because of open editgrid row

SilviaAmAm:
Fixes https://github.com/open-formulieren/open-forms/issues/3096

### The problem

- The function `checkComponentValidity` is called to validate the editGrid (https://github.com/formio/formio.js/blob/v4.13.13/src/components/editgrid/EditGrid.js#L969). This makes a call to super.
- The `super.checkComponentValidity` checks the editGrid 'as a whole': it checks the 'required' validator, the max length validator... If the component as a whole is valid, then it schedules a callback to `setComponentValidity`, which will clear errors on the component.
- The `checkComponentValidity` of editGrid now checks if each row is valid and if there are any open/unsaved rows. Since it is the case, it sets the component validity (note: it does it syncronously, it doesn't schedule a callback like the super.
- The callback of the super executes. This clears the error on the component (but it must miss something somewhere because the form submission is blocked, but no error is visible).

### Solution

Moved the super call AFTER the row checks. If the rows are invalid/unsaved, then the function returns and no super call is done.

The repeating group doesn't deal with async validation, so this could be a problem at some point :shrug:

### DIFF
--- a/src/formio/components/EditGrid.js
+++ b/src/formio/components/EditGrid.js
@@ -128,6 +128,55 @@ class EditGrid extends FormioEditGrid {
     return !!valid;
   }
 
+  /**
+   * Copied from Formio https://github.com/formio/formio.js/blob/v4.13.13/src/components/editgrid/EditGrid.js#L969
+   * The call to super was moved AFTER the checks of whether the rows are invalid or if there are open (unsaved rows).
+   *
+   * See: open-formulieren/open-forms#3096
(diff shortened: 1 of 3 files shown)


SilviaAmAm:
backport to 1.2.x: https://github.com/open-formulieren/open-forms-sdk/pull/440
backport to 1.3.x:

PR #446, opened 2023-06-09
Styling of form field description

### LINKED ISSUE #442: Style/theme form field description properly
Currently, descriptions/help texts are lacking the lightish blue background and the vertical spacing is a bit off.

**Tasks**

- [x] Fix color appearances to be consistent with form.io components
- [x] Check spacing of datefield for consistent padding/margins

sergei-maertens:
Closes #442
~~Depends on open-formulieren/design-tokens#24~~


sergei-maertens:
@Robbert I've asigned you for the refactor part where we are dropping our own class names in favour of the utrecht component ones. When the new design tokens version is published, Chromatic can do the visual regression testing.

src/components/forms/Label.js
```diff
@@ -1,19 +1,21 @@
...
-    </Paragraph>
-  );
-};
+const Label = ({id, isRequired = false, disabled = false, children}) => (
+  <Paragraph className="utrecht-form-field__label">
+    <FormLabel
+      htmlFor={id}
+      disabled={disabled}
+      className={classNames({
+        'utrecht-form-label--openforms-required': isRequired,
```
SilviaAmAm, line 12: Does this take into account the special DH style for required fields (plain for required fields and followed by 'optioneel' for optional fields)?

sergei-maertens: I've lifted up this configuration into the `ConfigContext` which is initialized/set in the `sdk.js`. This exposes it to the `Label` and new `Wrapper` components. I've also updated the CSS to apply this style to the utrecht components instead of only our own.

PR #443, opened 2023-06-08
Location and time step for appointments

sergei-maertens:
Closes open-formulieren/open-forms#3065

~~Depends on #452~~

PR #450, opened 2023-06-12
Appointment: location and date fields

sergei-maertens:
(Partially) closes open-formulieren/open-forms#3065

I've broken these components with their stories and tests out from #443 to make the PR a bit more manageable.

**Changes**

* `LocationSelect` field for appointment, based on the selected product(s)
* `DateSelect` field using the calendar, based on the selected product(s) and chosen location


src/components/appointments/LocationSelect.js
```diff
@@ -0,0 +1,47 @@
+import {useFormikContext} from 'formik';
+import React, {useCallback, useContext} from 'react';
+import {FormattedMessage} from 'react-intl';
+
+import {ConfigContext} from 'Context';
+import {get} from 'api';
+import {AsyncSelectField} from 'components/forms';
+
+// TODO: use a nicer widget/form field than select
```
sergei-maertens, line 9: Probably that'll be in v2.3 of OF and we'll ship with the dropdown in v2.2

src/components/appointments/Product.js
```diff
@@ -6,7 +6,7 @@ import {NumberField} from 'components/forms';
 import ProductSelect from './ProductSelect';
 
 const Product = ({namePrefix, index}) => (
-  <>
+  <div>
```
sergei-maertens, line 9: needed because the CSS rules target some `:first-child` so wrapping in a div makes the vertical spacing more digestible

src/components/appointments/mocks.js
```diff
@@ -7,9 +8,37 @@ const DEFAULT_PRODUCTS = [
...
+  },
+];
+
+const _getDate = (numDaysToAdd = 0) => {
+  return () => {
+    const now = new Date();
+    const newDate = numDaysToAdd ? addDays(now, numDaysToAdd) : now;
+    return formatISO(newDate, {representation: 'date'});
+  };
+};
```
sergei-maertens, line 28: this is a closure so that the `new Date()` is at run time and not at module-load time, to be able to use `jest.useFakeTimer({now: ...})` in tests, which gives stable dates to make assertions about, while still providing dynamic dates in storybook so that next year you don't need to click the "left" icon 10 times to see the available dates.

src/components/appointments/DateSelect.stories.mdx
```diff
@@ -0,0 +1,80 @@
...
+
+<Canvas>
+  <Story name="Calendar with available date range">{Template.bind({})}</Story>
+</Canvas>
+
+## Disable missing dates within range
+
+<Canvas>
+  <Story
+    name="Calendar with disabled gaps in date range"
```
SilviaAmAm, line 40: I think the difference between these two stories is not clear without looking at the code?
Screenshot from 2023-06-13 12-34-42

src/components/appointments/DateSelect.js
```diff
@@ -0,0 +1,70 @@
...
+    [baseUrl, JSON.stringify(productIds), locationId]
+  );
+
+  if (error) {
+    throw error;
+  }
+
+  const today = new Date();
+  const minDate = availableDates.length ? parseISO(availableDates[0]) : today;
+  const maxDate = availableDates.length ? parseISO(availableDates.at(-1)) : today;
```
CharString, line 48: The name availableDates suggest a set of dates to me, not a 2-tuple (start, end). I think the interval `[start, end>` is sanest, but we depend on 3rd parties; I know…

If no dates are available, today isn't either, right? Shouldn't that render some other suggestion? Raise an error maybe?

I love https://pypi.org/project/portion/ for these sort of problems.

CharString (changes requested):

src/components/appointments/DateSelect.js
```diff
@@ -0,0 +1,70 @@
+import {eachDayOfInterval, formatISO, parseISO} from 'date-fns';
+import {useFormikContext} from 'formik';
+import React, {useCallback, useContext} from 'react';
```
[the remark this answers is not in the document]
sergei-maertens: yeah, I noticed in my follow up PR, I'll clean it up since the production build likely will crash linting

src/components/appointments/DateSelect.js
```diff
@@ -0,0 +1,70 @@
...
+
+  // get the available dates from the API
+  const {
+    loading,
+    value: availableDates = [],
+    error,
+  } = useAsync(
+    async () => await getDates(baseUrl, productIds, locationId),
+    // about JSON.stringify: https://github.com/facebook/react/issues/14476#issuecomment-471199055
+    [baseUrl, JSON.stringify(productIds), locationId]
```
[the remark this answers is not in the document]
sergei-maertens: I'll change it to `JSON.stringify(values)` instead, as I noticed this is an option in a derived PR :)

Formik returns the same array I think, as it's in state, but it's iffy to rely on those behaviours.

> I hate JS where equality on a sequence of integers is too hard of a problem.

You and I both

src/components/appointments/DateSelect.js
sergei-maertens: availableDates is a `str[]` with dates in ISO 8601 format, it has arbitrary length. It's not a two-tuple. min and max are set to the first and last possible date, respectively. This makes up the interval of possible available dates, anything before or after is marked as disabled.

The 'gaps' in that array are then explicitly turned into disabled events for the calendar. If no dates are available, min and max are set to the same value and this results in all dates being disabled, which can indeed be improved by just rendering a message "sorry, no available dates, try again later" or something like instead.

I think there's also a bug when only one available date is returned, I'll add a story for that too.

CharString: > availableDates is a `str[]` with dates in ISO 8601 format, it has arbitrary length. It's not a two-tuple. min and max are set to the first and last possible date, respectively. This makes up the interval of possible available dates, anything before or after is marked as disabled.

Oh right. I remember the api for the date picker.

PR #451, opened 2023-06-12
Refactor formio inputs to be pure NL DS components

sergei-maertens:


sergei-maertens:
@Robbert you can take a look already, but this is some prep-work for next week!

sergei-maertens:
The height differences is because our own CSS used to pin the height via `height: 46px !important`, while the NL DS components derive the height via line-height, padding and border (ending up at 50px). The latter is more correct.

PR #452, opened 2023-06-13
Implement appointment time field

sergei-maertens:
(Partially) closes open-formulieren/open-forms#3065

Depends on #450 

Implements the time field for the appointment time, based on select products, location and date. Times are displayed in the browser/OS timezone.

I've reached out to Chromatic support to ask if we can set the timezone for the play functions, but in the worst case we can disable this particular story completely.

### DIFF
--- a/src/components/appointments/TimeSelect.js
+++ b/src/components/appointments/TimeSelect.js
@@ -1,0 +1,75 @@
+import {format, parseISO} from 'date-fns';
+import {useFormikContext} from 'formik';
+import React, {useCallback, useContext} from 'react';
+import {FormattedMessage} from 'react-intl';
+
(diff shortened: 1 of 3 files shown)


CharString (approved):
I don't know how my mom would interpret the "Times are in your local time" message.
But from experience from overhearing support questions on calendar systems, this is hard to get right without confusing anyone.

Implementation looks fine. My linter does warn, but we have #453 for that:

Diagnostics:
1. React Hook useCallback has missing dependencies: 'values.date', 'values.location', and 'values.products'. Either include them or remove the dependency array.
2. React Hook useCallback has a complex expression in the dependency array. Extract it to a separate variable so it can be statically checked.

PR #455, opened 2023-06-14
:art: Replace bem mixin with sass-module usage from microscope-sass

sergei-maertens:


CharString (commented):
:fire: Running out of Chromatic snapshots and the news of smarter snapshots has me confused. Does this mean it compared 129 snapshots and they were all pixel perfect?
image

PR #458, opened 2023-06-19
:globe_with_meridians: Update translations for 1.4.0 release

sergei-maertens:


src/components/FormStart/tests.spec.js
```diff
@@ -33,7 +33,7 @@ afterEach(() => {
 });
 
 const Wrap = ({children}) => (
-  <IntlProvider locale="nl" messages={messagesNL}>
+  <IntlProvider locale="nl" messages={messagesEN}>
```
SilviaAmAm, line 36: Should the locale also be `en`?

SilviaAmAm (approved):

src/components/FormStart/tests.spec.js
sergei-maertens: uh yeah, good catch!

PR #480, opened 2023-07-04
Ensure submission exists

sergei-maertens:
Partly closes open-formulieren/open-forms#3067

This hook ensures that a submission is created in the backend (and when done so, stores it in the browser session storage).

This is required for the backend permission checks, which require having a valid submission in the session.

The loader UI is displayed when the submission is being created, which prevents any other components from retrieving data from
the backend until a submission is available.

### DIFF
--- a/src/components/appointments/CreateAppointment.stories.js
+++ b/src/components/appointments/CreateAppointment.stories.js
@@ -5,7 +5,6 @@ import {RouterProvider, createMemoryRouter} from 'react-router-dom';
 
 import {buildForm} from 'api-mocks';
 import {mockSubmissionPost} from 'api-mocks/submissions';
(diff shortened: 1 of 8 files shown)


src/components/appointments/CreateAppointment.stories.js
```diff
@@ -15,13 +17,19 @@ import {
   mockAppointmentTimesGet,
 } from './mocks';
 
+const withClearSessionStorage = Story => {
+  window.sessionStorage.removeItem(SESSION_STORAGE_KEY);
+  return <Story />;
+};
+
```
SilviaAmAm, line 24: Discussed: Possibly make it into a global decorator and add a comment on when it runs

Issue #453, opened 2023-06-13
Avoid unnecessary fetches


sergei-maertens:
Using `JSON.stringify` in the appointments components as effect dependencies implies that the effect will run again when any form field value changes, e.g. when you select the date -> locations are going to be re-fetched. This does not make any sense.

sergei-maertens:
Solution -> spread only the relevant properties and pass those to `JSON.stringify`, and see if we can get stable ordering in there too

PR #485, opened 2023-07-05
Add client-side validation for appointment form

### LINKED ISSUE #435: Appointments: implement client-side validation
Epic: open-formulieren/open-forms#2471

Now that we're using Formik, we can/should also implement client-side validation (using zod) for the appointment information. Certain aspects (like at least one product required, minimum amount of product is 1) cannot be properly validated using the form inputs and we should catch those errors before even attempting to submit data to the backend.

- [x] Add Zod
- [x] Set up (partial) validation schema's depending on the current form step
    - [x] Set up product step schema
    - [x] Set up location step schema
    - [x] Set up customer fields schema
    - [x] Configure correct schema on `<Formik>` when navigating back
- [ ] #492
- [x] Fix Storybook play function tests
- [ ] #493

sergei-maertens:
Partly closes #435

~~Depends on #491~~

Client-side validation for a better user-experience. Note that this does *not* absolve the backend from doing any validation :wink:

### DIFF
--- a/src/components/appointments/ContactDetailsStep.js
+++ b/src/components/appointments/ContactDetailsStep.js
@@ -1,34 +1,28 @@
-import {useFormikContext} from 'formik';
-import {default as lodashGet} from 'lodash/get';
-import React, {useContext} from 'react';
+import {Form, Formik} from 'formik';
+import PropTypes from 'prop-types';
+import React, {useContext, useMemo} from 'react';
+import {flushSync} from 'react-dom';
 import {FormattedMessage, useIntl} from 'react-intl';
+import {useNavigate} from 'react-router-dom';
 import {useAsync} from 'react-use';
+import {z} from 'zod';
+import {toFormikValidationSchema} from 'zod-formik-adapter';
 
 import {ConfigContext} from 'Context';
 import {get} from 'api';
 import {CardTitle} from 'components/Card';
 import Loader from 'components/Loader';
-import {FormioComponent} from 'components/formio';
+import {FormioComponent, getEmptyValue, getSchema} from 'components/formio';
 import useTitle from 'hooks/useTitle';
 
-import {isStepValid as isProductStepValid} from './ChooseProductStep';
-import {isStepValid as isLocationStepValid} from './LocationAndTimeStep';
+import {useCreateAppointmentContext} from './CreateAppointment/CreateAppointmentState';
 import SubmitRow from './SubmitRow';
 
-// TODO: replace with ZOD validation, see #435
-export const isStepValid = (data, components = []) => {
-  if (!components.length) return false;
-  const hasMissingInput = components.some(component => {
-    const value = lodashGet(data, component.key);
-    return !value;
-  });
-  return isProductStepValid(data) && isLocationStepValid(data) && !hasMissingInput;
-};
-
-const ContactDetailsStep = () => {
+const ContactDetailsStep = ({navigateTo = null}) => {
   const intl = useIntl();
   const {baseUrl} = useContext(ConfigContext);
-  const {values} = useFormikContext();
+  const {submitStep, appointmentData, stepData} = useCreateAppointmentContext();
+  const navigate = useNavigate();
   useTitle(
     intl.formatMessage({
       description: 'Appointments: contact details step page title',
@@ -36,7 +30,9 @@ const ContactDetailsStep = () => {
     })
   );
 
-  const productIds = values.products.map(p => p.productId);
+  const products = appointmentData?.products || [];
+  const productIds = products.map(p => p.productId).sort();
+
   const {
     loading,
     value: components,
@@ -47,7 +43,17 @@ const ContactDetailsStep = () => {
   }, [baseUrl, JSON.stringify(productIds)]);
   if (error) throw error;
 
-  // TODO: hook up component.validate properties into zod schema, #435
+  const emptyValues =
+    !loading &&
+    Object.fromEntries(components.map(component => [component.key, getEmptyValue(component)]));
+
+  const validationSchema = useMemo(() => {
+    if (loading) return null;
+    const fieldSchemas = Object.fromEntries(
+      components.map(component => [component.key, getSchema(component)])
+    );
+    return z.object(fieldSchemas);
+  }, [loading, components]);
 
   return (
     <>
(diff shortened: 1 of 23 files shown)


src/components/appointments/LocationAndTimeStep.spec.js
```diff
@@ -60,27 +69,29 @@ beforeEach(() => {
...
+          location: '',
+          date: '',
+          datetime: '',
+        }
+      );
+      // No location should be selected, as there are multiple options
+      await waitFor(() => {
+        expect(screen.queryByText('Open Gem')).not.toBeInTheDocument();
+        expect(screen.queryByText('Bahamas')).not.toBeInTheDocument();
+      });
```
CharString, line 94:
```suggestion
    await act(async () => {
      render(
        <LocationAndTimeStep />,
        // product with multiple locations, see ./mocks.js
        {
          products: [{productId: 'e8e045ab', amount: 1}],
          location: '',
          date: '',
          datetime: '',
        }
      );
      // No location should be selected, as there are multiple options
    });
    await waitFor(() => {
      expect(screen.queryByText('Open Gem')).not.toBeInTheDocument();
    });
    expect(screen.queryByText('Bahamas')).not.toBeInTheDocument();

```
https://github.com/testing-library/eslint-plugin-testing-library/blob/ba14becb6b61d80cc029ddf9e249ba5e19020ffe/docs/rules/no-wait-for-multiple-assertions.md#rule-details

CharString: And there's another one of those in:
`autoselects location when there is only one option and enables the date field`

src/components/appointments/LocationSelect.js
```diff
@@ -15,10 +15,9 @@ const getLocations = async (baseUrl, productIds) => {
   return locationList;
 };
 
-const LocationSelect = () => {
+const LocationSelect = ({products}) => {
   const {baseUrl} = useContext(ConfigContext);
-  const {values} = useFormikContext();
-  const productIds = (values.products || []).map(prod => prod.productId);
+  const productIds = products.map(prod => prod.productId);
```
CharString, line 20:
```suggestion
  const productIds = products.map(prod => prod.productId).sort(); // sorting for stable identity
```

src/components/appointments/TimeSelect.js
```diff
@@ -20,14 +21,14 @@ const getDatetimes = async (baseUrl, productIds, locationId, date) => {
...
-const TimeSelect = () => {
+const TimeSelect = ({products}) => {
   const {baseUrl} = useContext(ConfigContext);
   const {values} = useFormikContext();
   const calendarLocale = useCalendarLocale();
 
   const getOptions = useCallback(
     async () => {
-      const productIds = (values.products || []).map(prod => prod.productId);
+      const productIds = products.map(prod => prod.productId);
```
CharString, line 31: Should we splitup the stringify in the dependency list of this callback to somehting like
```js
    [baseUrl, calendarLocale, JSON.stringify(values.location), values.date]
```
or whatever makes sense for the shape of these...

I'm a bit concerned this pattern of depending on the complete form state will get copied to bigger form states.

CharString (changes requested):

src/components/appointments/ContactDetailsStep.js
```diff
@@ -65,29 +71,52 @@ const ContactDetailsStep = () => {
...
+          )}
+        </Formik>
+      )}
     </>
   );
 };
 
-ContactDetailsStep.propTypes = {};
+ContactDetailsStep.propTypes = {
+  navigateTo: PropTypes.string,
```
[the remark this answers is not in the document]
sergei-maertens: I tried that but then we end up in cyclic imports which results in `undefined` for some component imports so I gave up on getting that to work. Another illusion shattered - JS modules can in fact *not* always guarantee that circular imports work as expected.

src/components/appointments/DateSelect.js
```diff
@@ -22,36 +23,40 @@ const getDates = async (baseUrl, productIds, locationId) => {
...
+    [baseUrl, JSON.stringify(productIds), location]
   );
 
   if (error) {
     throw error;
   }
 
-  if (!loading && values.location && availableDates && !availableDates.length) {
+  if (!loading && location && availableDates && !availableDates.length) {
+    // TODO: add label? make this a polite warning/error/alert?
```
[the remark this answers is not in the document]
sergei-maertens: *coughs*

src/components/appointments/LocationAndTimeStep.js
```diff
@@ -116,6 +153,13 @@ const ProductSummary = () => {
...
 
-ProductSummary.propTypes = {};
+ProductSummary.propTypes = {
+  products: PropTypes.arrayOf(
+    PropTypes.shape({
+      productId: PropTypes.string.isRequired,
+      amount: PropTypes.number.isRequired,
+    })
+  ).isRequired,
+};
```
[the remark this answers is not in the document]
sergei-maertens: yeah, was being repeated, I'll abstract it into a `types` module somewhere :+1:

src/components/appointments/ProductSelect.js
```diff
@@ -6,12 +6,36 @@ import {ConfigContext} from 'Context';
...
+
+const CACHED_PRODUCTS_KEY = 'appointment|products';
+const CACHED_PRODUCTS_MAX_AGE_MS = 15 * 60 * 1000; // 15 minutes
+
+const getProductsFromCache = () => {
+  const cachedValue = window.sessionStorage.getItem(CACHED_PRODUCTS_KEY);
+  if (cachedValue === null) return null;
+  const {value, timestamp} = JSON.parse(cachedValue);
+  // check if it's expired
+  const now = new Date().getTime();
```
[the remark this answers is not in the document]
sergei-maertens: and it's in milliseconds rather than a float of seconds :)

src/components/appointments/LocationAndTimeStep.spec.js
sergei-maertens: I thought I tried this variant and wasn't actually able to get rid of the warning about act, did that work for you?

src/components/appointments/TimeSelect.js
sergei-maertens: I had a ticket #453 where I intended to clean those up yeah!

PR #486, opened 2023-07-06
:sparkles: added searchbar in map component

bart-maykin:
Fixes open-formulieren/open-forms#2174

- added searchbar and button
- retrieved bag data

### DIFF
--- a/src/components/Map/Map.stories.js
+++ b/src/components/Map/Map.stories.js
@@ -3,7 +3,7 @@ import {userEvent, within} from '@storybook/testing-library';
 import {ConfigDecorator} from 'story-utils/decorators';
 
 import LeafletMap from '.';
-import {mockPdokApiResponseGet} from './mocks';
+import {mockAddressSearchGet} from './mocks';
 
 const withMapLayout = Story => (
   <div className="openforms-leaflet-map" style={{maxWidth: '600px'}}>
@@ -23,7 +23,7 @@ export default {
   },
   parameters: {
     msw: {
-      handlers: [mockPdokApiResponseGet],
+      handlers: [mockAddressSearchGet],
     },
   },
 };
--- a/src/components/Map/Map.stories.js
+++ b/src/components/Map/Map.stories.js
@@ -1,4 +1,9 @@
+import {userEvent, within} from '@storybook/testing-library';
+
+import {ConfigDecorator} from 'story-utils/decorators';
+
 import LeafletMap from '.';
+import {mockPdokApiResponseGet} from './mocks';
 
 const withMapLayout = Story => (
   <div className="openforms-leaflet-map" style={{maxWidth: '600px'}}>
@@ -9,16 +14,32 @@ const withMapLayout = Story => (
 export default {
   title: 'Private API / Map',
   component: LeafletMap,
-  decorators: [withMapLayout],
+  decorators: [withMapLayout, ConfigDecorator],
   args: {
     markerCoordinates: [52.1326332, 5.291266],
     defaultCenter: [52.1326332, 5.291266],
     defaultZoomLevel: 12,
-    disabled: true, // TODO: ideally this would be false but firefox has an infinite loop with onMarkerSet.
+    disabled: false,
   },
   parameters: {
-    chromatic: {disableSnapshot: true},
+    msw: {
+      handlers: [mockPdokApiResponseGet],
+    },
   },
 };
 
 export const Map = {};
+
+export const MapWithAddressSearch = {
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+    const button = await canvas.findByLabelText('Map component search button');
+    await userEvent.click(button);
+
+    const searchField = await canvas.findByPlaceholderText('Enter address, please');
+    await userEvent.type(searchField, 'Gemeente Utrecht');
+    const searchResult = await canvas.findByText('Utrecht, Utrecht, Utrecht');
+
+    await userEvent.click(searchResult);
+  },
+};
--- a/src/components/Map/index.js
+++ b/src/components/Map/index.js
@@ -176,6 +176,7 @@ const SearchControl = ({onMarkerSet, options}) => {
   }, [
     map,
     setMarker,
+    baseUrl,
     showMarker,
     showPopup,
     retainZoomLevel,
--- a/src/components/Map/index.js
+++ b/src/components/Map/index.js
@@ -1,11 +1,31 @@
+import {GeoSearchControl} from 'leaflet-geosearch';
 import isEqual from 'lodash/isEqual';
 import PropTypes from 'prop-types';
-import React, {useEffect} from 'react';
+import React, {useCallback, useContext, useEffect} from 'react';
+import {defineMessages, useIntl} from 'react-intl';
 import {MapContainer, Marker, TileLayer, useMap, useMapEvent} from 'react-leaflet';
 import {useGeolocation} from 'react-use';
 
+import {ConfigContext} from 'Context';
 import {DEFAULT_LAT_LNG, DEFAULT_ZOOM, MAP_DEFAULTS, TILE_LAYERS} from 'map/constants';
 import {getBEMClassName} from 'utils';
+
+import OpenFormsProvider from './provider';
+
+const searchControlMessages = defineMessages({
+  buttonLabel: {
+    description: "The leaflet map's search button areaLabel text.",
+    defaultMessage: 'Map component search button',
+  },
+  searchLabel: {
+    description: "The leaflet map's input fields placeholder message.",
+    defaultMessage: 'Enter address, please',
+  },
+  notFound: {
+    description: "The leaflet map's location not found message.",
+    defaultMessage: 'Sorry, that address could not be found.',
+  },
+});
 
 const useDefaultCoordinates = () => {
   // FIXME: can't call hooks conditionally
@@ -31,6 +51,7 @@ const LeaftletMap = ({
   defaultZoomLevel = DEFAULT_ZOOM,
   disabled = false,
 }) => {
+  const intl = useIntl();
   const defaultCoordinates = useDefaultCoordinates();
   const coordinates = markerCoordinates || defaultCoordinates;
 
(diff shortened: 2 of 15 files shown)


src/components/Map/index.js
```diff
@@ -82,6 +102,32 @@ const MapView = ({coordinates = null, zoomLevel = DEFAULT_ZOOM}) => {
...
+      onMarkerSet([result.location.y, result.location.x]);
+    }
+  };
+
+  useEffect(() => {
+    const searchControl = new GeoSearchControl({
+      provider: props.provider,
+      style: 'button',
+      ...props,
+    });
```
sergei-maertens, line 120: Can we investigate if we can modify this to add aria-label attributes where required?

Looking at the [source code](https://github.com/smeijer/leaflet-geosearch/blob/develop/src/SearchControl.ts#L487), it should be possible to do something like:

```js
searchControl.button.setAttribute("aria-label", intl.formatMessage(...)); // set accessible label
```
we need to make sure that `initialize` has run though, that might only be the case after the `map.addControl` call?

CharString: And I think a components `props`  are always implicitly a dependency, so don't have to be added to the dependency list of hooks.

sergei-maertens: no, that's not right - but it would be better to destructure the props and make explicit what is being passed along.

While the individual props may stay the same, this is not guaranteed for the object exposing them, and Javascript/hook dependency comparison does not work with deep equality but with object identity. So this would trigger the hook every time the component re-renders _if React doesn't memoize the props itself_.

.storybook/preview.js
```diff
@@ -11,6 +11,7 @@ import {Formio, Templates} from 'react-formio';
 import {fixIconUrls as fixLeafletIconUrls} from 'map';
 import 'scss/dte-theme.scss';
 import 'leaflet/dist/leaflet.css';
+import 'leaflet-geosearch/dist/geosearch.css';
```
sergei-maertens, line 14: this should not be needed anymore because of https://github.com/open-formulieren/open-forms-sdk/pull/486/files#diff-8b021783ff171ffef1c5228dfc0ff05cc75ba3a05883fc133f76cf25fda22d03R2

src/components/Map/Map.stories.js
```diff
@@ -18,6 +21,9 @@ export default {
   },
   parameters: {
     chromatic: {disableSnapshot: true},
```
sergei-maertens, line 23:
```suggestion
```
so, things appear to work properly in chromatic with the tile loading, let's give it at try for these stories too?

src/components/Map/index.js
```diff
@@ -59,6 +72,19 @@ const LeaftletMap = ({
...
         </>
       ) : null}
+      <SearchControl
+        provider={provider}
+        showMarker={false}
+        showPopup={false}
+        retainZoomLevel={false}
+        animateZoom={true}
+        autoClose={false}
+        searchLabel={'Enter address, please'}
```
sergei-maertens, line 82: this needs to be translated via `react-intl` and the `useIntl` hook (or maybe a `FormattedMessage` will also work)

src/components/Map/index.js
```diff
@@ -59,6 +72,19 @@ const LeaftletMap = ({
...
+        keepResult={true}
+        updateMap={true}
+        notFoundMessage={'Sorry, that address could not be found.'}
```
sergei-maertens, line 85: needs to be marked translatable

src/components/Map/index.js
```diff
@@ -77,11 +103,38 @@ const MapView = ({coordinates = null, zoomLevel = DEFAULT_ZOOM}) => {
...
+      style: 'button',
+      ...props,
+    });
+    searchControl.button.setAttribute('aria-label', intl.formatMessage(searchButtonAreaLabel));
+
+    map.addControl(searchControl);
+    return () => map.removeControl(searchControl);
+  }, []);
+
+  map.on('geosearch/showlocation', setMarker);
```
sergei-maertens, line 134: shouldn't this go in the `useEffect` to bind the event, and consequently in the cleanup function -> unbind the event?

src/components/Map/mocks.js
```diff
@@ -0,0 +1,119 @@
+import {rest} from 'msw';
+
+import {BASE_URL} from 'api-mocks';
+
+export const mockPdokApiResponse = rest.get(`${BASE_URL}formio/mapsearch`, (req, res, ctx) => {
```
sergei-maertens, line 5:
```suggestion
export const mockPdokApiResponseGet = rest.get(`${BASE_URL}formio/mapsearch`, (req, res, ctx) => {
```
please follow the convention of the mock names

src/components/Map/provider.js
```diff
@@ -0,0 +1,26 @@
...
+
+  parse(results) {
+    return results.map(location => ({
+      x: location.latLng.lng,
+      y: location.latLng.lat,
+    }));
+  }
+
+  async search({query}) {
+    const searchUrl = `${this.baseUrl}formio/mapsearch`;
```
sergei-maertens, line 16: you can lift this up into the `constructor`, now you're interpolating the API endpoint for every request, while this is only needed once.

sergei-maertens (changes requested):

src/components/Map/index.js
```diff
@@ -1,12 +1,32 @@
...
 import {MapContainer, Marker, TileLayer, useMap, useMapEvent} from 'react-leaflet';
 import {useGeolocation} from 'react-use';
 
+import {ConfigContext} from 'Context';
 import {DEFAULT_LAT_LNG, DEFAULT_ZOOM, MAP_DEFAULTS, TILE_LAYERS} from 'map/constants';
 import {getBEMClassName} from 'utils';
 
+import OpenFormsProvider from './provider';
+
+export const mapSearchControlButtonAreaLabel = defineMessage({
```
sergei-maertens, line 15: 1. why is this being exported?
2. Typo: areaLabel -> ar**i**aLabel
3. you can also use the `defineMessages` to created an object with your own keys, so you can refer to them in the code with `messages.searchControlLabel`, for example

src/components/Map/index.js
```diff
@@ -77,11 +114,43 @@ const MapView = ({coordinates = null, zoomLevel = DEFAULT_ZOOM}) => {
     if (!coordinates || coordinates.length !== 2) return;
     if (!coordinates.filter(value => isFinite(value)).length === 2) return;
     map.setView(coordinates, zoomLevel);
-  });
+    // eslint-disable-next-line
+  }, [coordinates]);
```
sergei-maertens, line 118: I don't think that disabling the linter is the way to go here, the `zoomLevel` prop is used so if it changes, the map view should be updated accordingly.

Additionally, the `map` itself is also a dependency, if that changes (in the parent component for example), then this component *also* has to update or you may even get memory leaks.

src/components/Map/index.js
```diff
@@ -77,11 +114,43 @@ const MapView = ({coordinates = null, zoomLevel = DEFAULT_ZOOM}) => {
...
   // rendering is done by leaflet, so just return null
   return null;
 };
 
+const SearchControl = props => {
```
sergei-maertens, line 123: we prefer spreading the props already in the component definition:

```suggestion
const SearchControl = ({ onMarkerSet, ... }) => {
```
since this component is wrapping the `GeoSearchControl` object, I'd package those options as a separate prop instead of polluting the component prop namespace.

So do this instead:

```jsx
<SearchControl onMarkerSet={onMarkerSet} options={{'showMarker': false, ...}} />
```
which makes it easier to pass everything down.

Additionally, I see that you pass along the provider instance, but can't you just create that instance inside of this component? That way, everything related to the geo search is self-contained and you have less props to pass along.

src/components/Map/index.js
```diff
@@ -77,11 +114,43 @@ const MapView = ({coordinates = null, zoomLevel = DEFAULT_ZOOM}) => {
...
+    searchControl.button.setAttribute(
+      'aria-label',
+      intl.formatMessage(mapSearchControlButtonAreaLabel)
+    );
+
+    map.addControl(searchControl);
+    map.on('geosearch/showlocation', setMarker);
+    return () => map.removeControl(searchControl);
+    // eslint-disable-next-line
+  }, [intl]);
```
sergei-maertens, line 149: please don't just shut up the linter without a good explanation why.

src/components/Map/provider.js
sergei-maertens, line 1: I mentioned that testing of the leaflet component itself was not necessary, but you can definitely add some unit tests for this provider implementation, as it's just pure Javascript.

src/components/Map/provider.js
```diff
@@ -0,0 +1,27 @@
...
+    return results.map(location => ({
+      label: location.label,
+      x: location.latLng.lng,
+      y: location.latLng.lat,
+    }));
+  }
+
+  async search({query}) {
+    const searchUrl = this.baseUrl;
+    const searchResults = await get(searchUrl, {
```
sergei-maertens, line 18:
```suggestion
    const searchResults = await get(this.endpoint, {
```
and set the `this.endpoint` in the constructor.

There's no benefit to assigning an intermediate variable.

sergei-maertens (changes requested):

src/components/Map/index.js
```diff
@@ -59,6 +80,20 @@ const LeaftletMap = ({
           <MarkerWrapper position={coordinates} onMarkerSet={onWrapperMarkerSet} />
         </>
       ) : null}
+      <SearchControl
+        onMarkerSet={onMarkerSet}
+        options={{
+          showMarker: 'false',
```
sergei-maertens, line 86:
```suggestion
          showMarker: false,
```

sergei-maertens (commented):
So, to wrap up this PR two more things:

* fix the string/boolean prop that was left over from our testing yesterday
* add some unit tests for the map provider

And then meta - when I'm looking at the individual commits, the commit messages don't reflect what's in the code changes and you're not using the correct gitmoji (see https://gitmoji.dev/). Can you please apply a bit more discipline on that aspect? The commit history is frequently used to figure out why certain changes were made. For this PR, I'd expect 3 commits:

* :heavy_plus_sign: Add leaflet-geosearch depenency (changed files: package.json and yarn.lock)
* :white_check_mark: Add interaction tests in map stories for geo search
* :sparkles: Add geo search control to leaflet map using our own search provider

I'm sure you can split them apart even further to account for the refactors, but let's get started with the simple stuff and focus on entire files for now.

sergei-maertens (approved):

PR #488, opened 2023-07-11
:sparkles: [#2175] added configureable zoom and start location

bart-maykin:
Fixes open-formulieren/open-forms#2175

- retrieved default zoom and location from configuration
- set starting location as configured latitude and longitude 
- set starting zoomlevel to be either map default or the configured zoomlevel

### DIFF
--- a/src/components/Map/Map.stories.js
+++ b/src/components/Map/Map.stories.js
@@ -12,6 +12,8 @@ export default {
   decorators: [withMapLayout],
   args: {
     markerCoordinates: [52.1326332, 5.291266],
+    defaultCenter: [52.1326332, 5.291266],
+    defaultZoomLevel: 12,
     disabled: true, // TODO: ideally this would be false but firefox has an infinite loop with onMarkerSet.
   },
   parameters: {
@@ -19,4 +21,12 @@ export default {
   },
 };
 
-export const Map = {};
+export const Map = {
+  args: {
+    component: {
+      lat: 52.1326332,
+      lng: 5.291266,
+      defaultZoom: 12,
+    },
+  },
+};
--- a/src/components/Map/Map.stories.js
+++ b/src/components/Map/Map.stories.js
@@ -21,12 +21,4 @@ export default {
   },
 };
 
-export const Map = {
-  args: {
-    component: {
-      lat: 52.1326332,
-      lng: 5.291266,
-      defaultZoom: 12,
-    },
-  },
-};
+export const Map = {};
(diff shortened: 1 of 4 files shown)


src/formio/components/Map.js
```diff
@@ -90,16 +91,26 @@ export default class Map extends Field {
...
     if (!container) return;
 
     this.reactRoot.render(
       <LeafletMap
-        markerCoordinates={markerCoordinates || null}
+        markerCoordinates={markerCoordinates}
         onMarkerSet={this.onMarkerSet.bind(this)}
+        zoomLevel={zoom || MAP_DEFAULTS.zoom}
       />
     );
```
sergei-maertens, line 115: this isn't correct - the marker implies the select location and when no default value is provided, there just initially are no coordinates selected. Therefor, `markerCoordinates` can still be `null`.

It looks like you're trying to use `markerCoordinates` to set the center of the map view, but that's an entirely separate concept/prop that should be added.

sergei-maertens (changes requested):

src/components/Map/index.js
```diff
@@ -49,7 +55,7 @@ const LeaftletMap = ({markerCoordinates, onMarkerSet, disabled = false}) => {
       <TileLayer url={TILE_LAYERS.url} {...TILE_LAYERS.options} />
       {coordinates ? (
         <>
-          <MapView coordinates={coordinates} />
+          <MapView coordinates={defaultCenter} zoomLevel={defaultZoomLevel} />
```
bart-maykin, line 58: Double check if this doesn't reset the center view and zoom level when user places a marker.

src/formio/components/Map.js
```diff
@@ -91,15 +92,23 @@ export default class Map extends Field {
   }
 
   renderReact() {
+    const defaultCenter =
+      this.component.initialCenter.lat && this.component.initialCenter.lng
+        ? [this.component.initialCenter.lat, this.component.initialCenter.lng]
+        : DEFAULT_LAT_LON;
```
bart-maykin, line 98: maybe

```js
const [defaultLat, defaultLon] = DEFAULT_LAT_LON;
const {lat = defaultLat, lng = defaultLon} = this.component?.initalCenter || {};
const defaultCenter = [lat, lng];
```

src/components/Map/Map.stories.js
```diff
@@ -12,11 +12,21 @@ export default {
   decorators: [withMapLayout],
   args: {
     markerCoordinates: [52.1326332, 5.291266],
+    defaultCenter: [52.1326332, 5.291266],
```
bart-maykin, line 15: changing this does not appear to have an effect in storybook -> check if it's also in the component itself but either way this should work as expected

sergei-maertens (approved):

PR #503, opened 2023-07-19
Refactor form 'prop' and (nested) routes

sergei-maertens:
Partly closes open-formulieren/open-forms#3067

This refactors the existing application to pass the `form` via context so we can make the router routes statically defined and avoid some prop-drilling, moving more towards the intended usage of react-router's data routers.

### DIFF
--- a/src/Context.js
+++ b/src/Context.js
@@ -1,5 +1,25 @@
 import React from 'react';
 
+const FormContext = React.createContext({
+  uuid: '',
+  name: '',
+  slug: '',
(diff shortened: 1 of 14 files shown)


src/components/appointments/CreateAppointment/Confirmation.js
sergei-maertens, line 1: WIP and will fill this out in a different PR, you can ignore this.

SilviaAmAm (approved):

Issue #505, opened 2023-07-20
Appointment form - ensure session expiration is handled correctly


sergei-maertens:
When the session expires, any and all session-related data should be removed from the react state and session storage in the browser.

Desired flow:

1. Load the SDK and form
2. A submission is created and the user is redirected to the products selection step
3. User idles on any step, including the summary page, sufficiently long for the session to expire
4. Session expires
    - The submission must be removed from the state and sessionStorage so that a navigate/refresh does not try to re-use it
    - All appointment data must be removed from the state and sessionStorage
5. When the user navigates back to the start of the form
    - GOTO 2

PR #510, opened 2023-07-20
Appointment confirmation page

sergei-maertens:
Closes open-formulieren/open-forms#3067

Depends on #507

### DIFF
--- a/src/components/appointments/CreateAppointment/CreateAppointment.stories.js
+++ b/src/components/appointments/CreateAppointment/CreateAppointment.stories.js
@@ -6,6 +6,7 @@ import {RouterProvider, createMemoryRouter} from 'react-router-dom';
 import {FormContext} from 'Context';
 import {buildForm} from 'api-mocks';
 import {mockSubmissionPost} from 'api-mocks/submissions';
+import {mockSubmissionProcessingStatusGet} from 'api-mocks/submissions';
 import {mockPrivacyPolicyConfigGet} from 'components/Summary/mocks';
 import {loadCalendarLocale} from 'components/forms/DateField/DatePickerCalendar';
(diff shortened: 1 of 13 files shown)


src/components/appointments/CreateAppointment/CreateAppointment.stories.js
SilviaAmAm, line 1: Maybe it would be nice to add a story for when the session expires?

src/components/appointments/CreateAppointment/Summary.js
```diff
@@ -38,7 +38,8 @@ const createAppointment = async (baseUrl, submission, appointmentData, privacyPo
     contactDetails,
     privacyPolicyAccepted,
   };
-  return await post(`${baseUrl}appointments/appointments`, body);
+  const response = await post(`${baseUrl}appointments/appointments`, body);
+  return response.data;
```
SilviaAmAm, line 42: I think that if the POST returns a 400, `data` contains the errors. Should we maybe check the status code and raise an error in that case? :thinking:

src/components/appointments/CreateAppointment/CreateAppointment.stories.js
sergei-maertens: #505 will handle this!

src/components/appointments/CreateAppointment/Summary.js
sergei-maertens: that already happens - the API client essentially throws for non-success status codes, see: https://github.com/open-formulieren/open-forms-sdk/blob/92fe7b7a840087d1b1aa20cc399c5229862efcf1/src/api.js#L126

in short, the call order: `post -> _unsafe -> apiCall -> throwForStatus`

SilviaAmAm: Ohh indeed!

PR #511, opened 2023-07-20
:sparkles: [#3045] implemented suffix for number field

bart-maykin:
Fixes https://github.com/open-formulieren/open-forms/issues/3045

- changed text.ejs file to use OF suffix
- added group styling to support suffix (and prefix if wanted later)
- added storybook story for number field with suffix

Screenshot from 2023-07-21 00-22-34

### DIFF
--- a/src/formio/templates/text.ejs
+++ b/src/formio/templates/text.ejs
@@ -2,7 +2,7 @@
 <div class="{{ctx.ofPrefix}}input-container">
 {% } %}
 {% if (ctx.prefix) { %}
-<span class="{{ctx.ofPrefix}}input-container__prefix" ref="prefix">
+<span class="{{ctx.ofPrefix}}input-container__affix {{ctx.ofPrefix}}input-container__affix--prefix" ref="prefix">
     {% if(ctx.prefix instanceof HTMLElement){ %}
       {{ ctx.t(ctx.prefix.outerHTML) }}
     {% } else{ %}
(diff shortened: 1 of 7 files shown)


src/formio/templates/text.ejs
```diff
@@ -1,14 +1,14 @@
 {% if (ctx.prefix || ctx.suffix) { %}
-<div class="group">
+<div class="{{ctx.ofPrefix}}group">
```
sergei-maertens, line 2:
```suggestion
<div class="{{ctx.ofPrefix}}input-group">
```

src/formio/templates/text.ejs
```diff
@@ -1,14 +1,14 @@
...
 {% } %}
 {% if (ctx.prefix) { %}
-<div class="prefix" ref="prefix">
+<span class="{{ctx.ofPrefix}}group {{ctx.ofPrefix}}group--prefix" ref="prefix">
```
sergei-maertens, line 5:
```suggestion
<span class="{{ctx.ofPrefix}}input-group__prefix" ref="prefix">
```

src/formio/templates/text.ejs
```diff
@@ -24,13 +24,13 @@
 <span class="wordcount" ref="wordcount"></span>
 {% } %}
 {% if (ctx.suffix) { %}
-<div class="suffix" ref="">
+<span class="{{ctx.ofPrefix}}group {{ctx.ofPrefix}}group--suffix" ref="suffix">
```
sergei-maertens, line 27:
```suggestion
<span class="{{ctx.ofPrefix}}input-group__suffix" ref="suffix">
```

sergei-maertens (changes requested):
Some BEM naming errors:

- the block is the `openforms-input-group`
- the elements are `openforms-input-group__prefix` and openforms-input-group__suffix`
- technically there's also a child `openforms-input-group__input` but we don't attach any styling to that currently, so let's not bother

I haven't looked at the CSS yet

sergei-maertens (approved):

PR #516, opened 2023-07-25
Fix datefield month january

### LINKED ISSUE #515: Date of birth is not properly displayed on hard refresh
The split date input (day/month/year) chokes om something - when entering 1/1/1980, submitting the step and then navigating back to step (potentially also hard refreshing it) seems to clear the month field, while the date entered *is* valid.

Potentially this has something to do with Javascript's bizarre month-0 indexing/numbering?

sergei-maertens:
Closes #515

### DIFF
--- a/src/components/forms/DateField/utils.js
+++ b/src/components/forms/DateField/utils.js
@@ -47,7 +47,7 @@ export const getDateLocaleMeta = locale => {
 };
 
 export const convertMonth = (month, toAdd) => {
-  if (!month) return '';
+  if (month === '' || month == null) return '';
   const monthNumber = parseInt(month);
   return String(monthNumber + toAdd);
 };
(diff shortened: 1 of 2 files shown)


src/components/forms/DateField/utils.js
```diff
@@ -47,7 +47,7 @@ export const getDateLocaleMeta = locale => {
 };
 
 export const convertMonth = (month, toAdd) => {
-  if (!month) return '';
+  if (month === '' || month == null) return '';
   const monthNumber = parseInt(month);
   return String(monthNumber + toAdd);
```
CharString, line 52:
```suggestion
  const monthNumber = parseInt(month);
  return isNaN(monthNumber)?'':String(monthNumber + toAdd);
```

This catches `undefined` and whatever JS idiosyncrasies  I haven't thought of.
Is it okay for this to go out of bounds to 13? :thinking: 12?

CharString (changes requested):

src/components/forms/DateField/utils.js
sergei-maertens: Undefined is already caught with the non-strict equal null. This is ugly but idiomatic javascript.

13 and the likes should be covered later when a date instance is constructed and checked against isNaN

CharString (approved):

PR #519, opened 2023-07-28
:heavy_plus_sign: [#2176] added gesture handling for leaflet

bart-maykin:
Fixes https://github.com/open-formulieren/open-forms/issues/2176

- Added leaflet-gesture-handling
- Implemented leaflet-gesture-handling
- removed zoomlevel as a argument from the setView in the MapView component

### DIFF
--- a/src/components/Map/index.js
+++ b/src/components/Map/index.js
@@ -1,5 +1,4 @@
 import {GeoSearchControl} from 'leaflet-geosearch';
-import {GestureHandling} from 'leaflet-gesture-handling';
 import isEqual from 'lodash/isEqual';
 import PropTypes from 'prop-types';
 import React, {useCallback, useContext, useEffect} from 'react';
@@ -88,6 +87,7 @@ const LeaftletMap = ({
       crs={MAP_DEFAULTS.crs}
       attributionControl
       className={className}
+      gestureHandling
       gestureHandlingOptions={{
         text: {
           touch: intl.formatMessage(leafletGestureHandlingText.touch),
@@ -119,7 +119,6 @@ const LeaftletMap = ({
         }}
       />
       {disabled ? <DisabledMapControls /> : <CaptureClick setMarker={onMarkerSet} />}
-      <GestureHandlingSetter />
     </MapContainer>
   );
 };
@@ -274,15 +273,6 @@ const CaptureClick = ({setMarker}) => {
 
 CaptureClick.propTypes = {
   setMarker: PropTypes.func.isRequired,
-};
-
-const GestureHandlingSetter = () => {
-  const map = useMap();
-  useEffect(() => {
-    map.gestureHandling.enable();
-    map.addHandler('gestureHandling', GestureHandling);
-  }, [map]);
-  return null;
 };
 
 export default LeaftletMap;
--- a/src/components/Map/index.js
+++ b/src/components/Map/index.js
@@ -1,4 +1,5 @@
 import {GeoSearchControl} from 'leaflet-geosearch';
+import {GestureHandling} from 'leaflet-gesture-handling';
 import isEqual from 'lodash/isEqual';
 import PropTypes from 'prop-types';
 import React, {useCallback, useContext, useEffect} from 'react';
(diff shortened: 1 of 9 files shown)


bart-maykin:
Jest tests seem to be failing because Leaflet (L) is not defined in the index.js file.

Screenshot from 2023-07-28 19-29-03

src/components/Map/index.js
```diff
@@ -251,4 +254,11 @@ CaptureClick.propTypes = {
   setMarker: PropTypes.func.isRequired,
 };
 
+const GestureHandlingSetter = () => {
+  const map = useMap();
+  map.gestureHandling.enable();
+  map.addHandler('gestureHandling', GestureHandling);
```
sergei-maertens, line 260: this is really something that you can't do with react components - every render means this function gets executed every time, and we are not (nor shouldn't be) in control of when renders happen - that's a React internal.

Effectively, running this kind of initialization during the render is a bad idea - it belongs in a `useEffect` that only runs once, or better (from https://github.com/elmarquis/Leaflet.GestureHandling/issues/65), it appears you can run:

```js
L.Map.addInitHook("addHandler", "gestureHandling", GestureHandling);
```
this looks a lot like a global, so perhaps we can even move that out of react alltogether (and put it in `src/map/index.js`) and handle it similarly to `fixIconUrls`.

src/components/Map/index.js
```diff
@@ -113,6 +115,7 @@ const MapView = ({coordinates = null}) => {
     if (!coordinates.filter(value => isFinite(value)).length === 2) return;
     map.setView(coordinates);
   }, [map, coordinates]);
+
```
sergei-maertens, line 118: so the previous PR removed the blank line, and this one adds it back? :stuck_out_tongue: 

Please avoid these kind of cosmetic changes that only clutter the diff :wink:

src/scss/components/_leaflet-map.scss
```diff
@@ -2,6 +2,7 @@
 @import 'leaflet-geosearch/dist/geosearch.css';
 
 @import '../mixins/prefix';
+@import 'leaflet-gesture-handling/dist/leaflet-gesture-handling.css';
```
sergei-maertens, line 5: please pay attention to the import ordering - you can see that there already is a leaflet-related CSS import at the top. This belongs directly below it (sorting alphabetically).

In general, we follow the pattern: stdlib / third party dependencies / own project / relative imports for the import groups.

sergei-maertens (changes requested):
Some more general remarks:

* https://github.com/elmarquis/Leaflet.GestureHandling/issues/86 -> when switching the language in storybook, I'm only seeing English messages and never Dutch, that's wrong behaviour
* There's not a lot of activity in the library/repository, so this is a bit of a risk :thinking: if it works well, we can accept the risk
* https://github.com/elmarquis/Leaflet.GestureHandling/issues/65 is concerning, have you seen similar behaviour?

bart-maykin:
> Some more general remarks:
> 
>     * [Wrong default language - WCAG 3.1.2 failure elmarquis/Leaflet.GestureHandling#86](https://github.com/elmarquis/Leaflet.GestureHandling/issues/86) -> when switching the language in storybook, I'm only seeing English messages and never Dutch, that's wrong behaviour
> 
>     * There's not a lot of activity in the library/repository, so this is a bit of a risk thinking if it works well, we can accept the risk
> 
>     * [two finger drag / page scroll prevention and glitchy panning elmarquis/Leaflet.GestureHandling#65](https://github.com/elmarquis/Leaflet.GestureHandling/issues/65) is concerning, have you seen similar behaviour?

- I added react-intel translations which seem to fix the translation problem.
- I looked at the video capture that he provided and from what I can see we are not running into the same problems. But maybe it would be good if you double check for me with the storybook build in this pr.

sergei-maertens (approved):

sergei-maertens:
Cleaned up a little bit to register the handler globally and enable it via option so we don't need the component with side-effect. Tested on mobile with storybook and the panning/dragging is smooth for me too.

PR #520, opened 2023-07-31
Always allow formik submission and validate on submit

### LINKED ISSUE #514: Submit button is blocked on a hard refresh with valid data
This brings the formik form in a non-dirty state, which causes the button to be disabled.

Not sure yet how to fix this...

sergei-maertens:
Closes #514

Due to situations (hard refresh) where valid data is filled out, but no changes are made, the form is not considered to be dirty, which blocked the submit button. Relying on the valid state is not an alternative option either, as this requires validation to run on mount, which in turn discards the initialErrors set via the backend.

Additionally, running validation on blur/on change has some drawbacks in the UI:
- browser autocomplete does not fire onBlur, which requires another click or [Tab] by the user to blur the active input for validadtion to trigger
- blur on the datepicker input triggers validation to early when the user is still selecting a date in the calendar (possibly navigating to the next/previous month)
- blur on the date input group may trigger validation too soon when a date part like day/month is being entered and no full date is yet available to validate
- running validation on change yields intrusive validation errors for email address input while the user is typing

Deferring the client-side validation to the actual submit action is the easiest solution in this case.

Possible future improvement: scroll the first validation error into view.

### DIFF
--- a/src/components/appointments/ChooseProductStep.js
+++ b/src/components/appointments/ChooseProductStep.js
@@ -94,7 +94,7 @@ const ChooseProductStepFields = ({values: {products = []}, validateForm, isValid
       </FieldArray>
 
       <SubmitRow
-        canSubmit={dirty && isValid}
+        canSubmit
         nextText={intl.formatMessage({
           description: 'Appointments products step: next step text',
           defaultMessage: 'Confirm products',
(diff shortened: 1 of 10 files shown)


CharString (commented):
Am I holding it wrong?
[Afspraak-57bfed81460.webm](https://github.com/open-formulieren/open-forms-sdk/assets/325643/c44886fd-3a14-44d7-8119-22b8a7579b52)

sergei-maertens:
> Am I holding it wrong? [Afspraak-57bfed81460.webm](https://github.com/open-formulieren/open-forms-sdk/assets/325643/c44886fd-3a14-44d7-8119-22b8a7579b52)

How did I manage to f*** this up so badly

sergei-maertens:
@CharString should be fixed now! There was a validation side effect causing this that can now be removed because of the nature of this patch/ticket.

CharString (approved):

PR #522, opened 2023-08-03
Only allow selecting valid product combinations

sergei-maertens:
Partially closes open-formulieren/open-forms#3332

Backend PR: https://github.com/open-formulieren/open-forms/pull/3338

When selecting one or more products and adding another product, the available options must be limited to valid options in combination with the earlier selected products.

### DIFF
--- a/src/components/appointments/mocks.js
+++ b/src/components/appointments/mocks.js
@@ -1,4 +1,5 @@
 import {addDays, formatISO} from 'date-fns';
+import isEqual from 'lodash/isEqual';
 import {rest} from 'msw';
 
 import {BASE_URL} from 'api-mocks';
@@ -6,6 +7,7 @@ import {BASE_URL} from 'api-mocks';
 const DEFAULT_PRODUCTS = [
   {code: 'PASAAN', identifier: '166a5c79', name: 'Paspoort aanvraag'},
   {code: 'RIJAAN', identifier: 'e8e045ab', name: 'Rijbewijs aanvraag (Drivers license)'},
+  {code: 'LIMITED', identifier: 'ea04db83', name: 'Not available with drivers license'},
 ];
 
 const LOCATIONS = [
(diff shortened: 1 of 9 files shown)


src/components/appointments/mocks.js
```diff
@@ -43,7 +77,15 @@ const DATES = [
...
-    return res(ctx.json(DEFAULT_PRODUCTS));
+    let products = DEFAULT_PRODUCTS;
+    const productIds = req.url.searchParams.getAll('product_id');
+    if (productIds.length) {
+      // find the matching configuration
+      const config = PRODUCT_BY_ID.find(mapping => isSameIds(mapping.other, productIds));
+      if (!config) throw new Error('Invalid querystring combination used.');
+      products = products.filter(p => config.products.includes(p.identifier));
+    }
+    return res(ctx.json(products));
```
CharString, line 88: I find the whole lookup table a bit confusing. Does it do this: ?
```suggestion
    // return all the unpicked products
    const knownIds = DEFAULT_PRODUCTS.map(p => p.identifier);
    // excludes the ids already picked
    const excludeIds = req.url.searchParams.getAll('product_id');
    // check for typos
    if (excludeIds.reduce((anyUnknown, id) => anyUnknown || !knownIds.includes(id), false))
      throw new Error('Unknown id in querystring used.');
    // 'e8e045ab' 'ea04db83' are mutually exclusive
    if (excludeIds.includes('e8e045ab') && excludeIds.includes('ea04db83'))
      throw new Error('Invalid querystring combination used.');
    if (excludeIds.includes('e8e045ab')) excludeIds.push('ea04db83');
    if (excludeIds.includes('ea04db83')) excludeIds.push('e8e045ab');

    const products = DEFAULT_PRODUCTS.filter(p => !excludeIds.includes(p.identifier));
    return res(ctx.json(products));
```

CharString (changes requested):

src/components/appointments/mocks.js
sergei-maertens: not exactly - I didn't want to bake the combinations into the mock function but put them into a constant instead for explicit configuration. I'm having trouble understanding this code.

It also seems with JCC that the exclusions/combinations are not associative (?), e.g. given there are products A, B, C and D, when I pick product A, for the second product I can pick B or C and when I pick B as second product I can only pick C.

But, when I pick product C as first product, I can then only pick product B as second product (and A is not available).

The lookup table is: given a set of select product IDs, which other product IDs are then still available.

PR #527, opened 2023-08-09
[OF#3203] Support multiple declaration checkboxes

SilviaAmAm:
Partly fixes open-formulieren/open-forms#3203

Depends on https://github.com/open-formulieren/open-forms/pull/3344


src/components/DeclarationCheckboxes/DeclarationCheckbox.js
SilviaAmAm, line 1: The `DeclarationCheckbox.*` files are the moved `PrivacyCheckbox.*` files (and the name of the component `PrivacyCheckbox` has been changed to `DeclarationCheckbox`)

src/components/Summary/test.spec.js
```diff
@@ -69,15 +71,13 @@ it('Summary displays logout button if isAuthenticated is true', () => {
...
-        <IntlProvider locale="nl" messages={messagesNL}>
-          <SubmissionSummary
-            form={testForm}
-            submission={SUBMISSION}
-            onConfirm={onConfirm}
-            onLogout={onLogout}
-            onClearProcessingErrors={() => {}}
-          />
-        </IntlProvider>
+        <SubmissionSummary
```
SilviaAmAm, line 74: The `IntlProvider` is in the `Wrap`

src/components/Summary/GenericSummary.js
```diff
@@ -42,7 +42,7 @@ const GenericSummary = ({
         <ErrorMessage key={error}>{error}</ErrorMessage>
       ))}
       <Formik
-        initialValues={{privacy: false}}
+        initialValues={{}}
```
sergei-maertens, line 45: Formik requires initial values to be specified for all the fields that are managed by it, otherwise you get react warnings that mention either:

- `undefined` is not a valid value for a field (inputs must have a string value, typically, for checkboxes this translated to the `checked` prop)
- form field is changing from uncontrolled to controlled, which is an anti-pattern

So I think we should just include `privacy` and `truth` as values here (mapping to the appropriate form fields) with the default false, even if they are not relevant for this particular form.

In the future, I think we should leverage the react-router datarouters where we can configure a `loader` at the route level. This should then consult the API endpoint and provide that data, so the GenericSummary component can then get the configuration from `useLoaderData` hook. However, that's too involved too tackle now, so that's why I'd opt for just including them in a hardcoded way for now.

src/components/SummaryConfirmation/index.js
```diff
@@ -1,55 +1,113 @@
...
+  const showWarnings = formikValues => {
+    const declarationValues = getDeclarationValues(declarationsInfo, formikValues);
+
+    let updatedWarnings = {...declarationsWarnings};
+    Object.entries(declarationValues).forEach(([declarationKey, declarationValue]) => {
+      if (!declarationValue) {
+        updatedWarnings = {...updatedWarnings, [declarationKey]: true};
+      }
+    });
+    setDeclarationWarnings(updatedWarnings);
```
sergei-maertens, line 83: perhaps a reducer is more readable here? now it looks like we're `zip`'ing arrays to match field name + value + showWarning yes/no while we can probably make that more self-contained?

src/components/SummaryConfirmation/index.js
```diff
@@ -1,55 +1,113 @@
...
+  };
+
+  const declarationValues = getDeclarationValues(declarationsInfo, formikValues);
+  const previousDeclarationValues = usePrevious(declarationValues);
+
+  useEffect(() => {
```
sergei-maertens, line 89: I've learned this myself recently, but you don't actually need `useEffect` to update state directly in this component, you can do it directly without the `useEffect` and that improves reactivity because you skip a render cycle (and it also makes the code a bit easier to follow).

This only works if you're calling the `setState` from `useState` hook in the same component, see also: https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

src/components/SummaryConfirmation/index.js
```diff
@@ -65,7 +123,7 @@ const SummaryConfirmation = ({submissionAllowed, onPrevPage}) => {
               variant="primary"
               name="confirm"
               disabled={loading || submitDisabled}
-              onDisabledClick={() => !loading && setWarningUncheckedPrivacy(true)}
+              onDisabledClick={() => !loading && showWarnings(formikValues)}
```
sergei-maertens, line 126: looking at this and the code above, I think this should still be boolean state? Either warnings should be shown or not. However, the single checkbox requires some additional display logic:

1. StatementCheckbox takes props `label`, `name` and `showWarning`
2. each checkbox can use `useFormikContext().getFieldProps(name).value` to check whether it's checked or not
3. only display the warning if `showWarning && !value` (so don't display the warning if it's checked and if it's unchecked only display when the disabled submit button was clicked)

Clicking a checkbox to enable it will then make the warning disappear for that particular field. The `setWarning` would then be called if all checkbox fields are checked.

src/components/appointments/CreateAppointment/CreateAppointment.stories.js
```diff
@@ -202,6 +202,17 @@ export const HappyFlow = {
...
+    await step('Wait for the declarations to load', async () => {
+      await waitFor(
+        async () =>
+          await expect(
+            await canvas.getByLabelText(
+              /I accept the privacy policy and consent to the processing of my personal data/
+            )
+          )
+      );
+    });
```
sergei-maertens, line 214: I think you can use `await canvas.findByLabelText` which will wait for everything to load, but I'm not 100%. So that might help you get rid of the `waitFor` (it should be similar to handling the dynamic product/location list dropdowns).

src/components/SummaryConfirmation/index.js
```diff
@@ -24,32 +34,30 @@ const SummaryConfirmation = ({submissionAllowed, onPrevPage}) => {
...
-      {displayPrivacyNotice && (
-        <PrivacyCheckbox label={privacyLabel} showWarning={warningUncheckedPrivacy} />
-      )}
       {loading && <Loader />}
+      <StatementCheckboxes
+        loading={loading}
+        canSubmit={canSubmit}
+        statementsInfo={statementsInfo}
+        showWarnings={showStatementWarnings}
+      />
```
sergei-maertens, line 60: I think the checkboxes themselves don't really need to know about `canSubmit` and we could rather do something like:

```jsx
{canSubmit && <StatementCheckboxes ... />}
```

Probably we should also incorporate `loading` in here?

src/components/StatementCheckboxes/index.js
```diff
@@ -0,0 +1,41 @@
+import {useFormikContext} from 'formik';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import StatementCheckbox from 'components/StatementCheckboxes/StatementCheckbox';
```
sergei-maertens, line 5:
```suggestion
import StatementCheckbox from './StatementCheckbox';
```
I'd say a relative import is a bit more appropriate since it's all grouped around the same main public component?

sergei-maertens (approved):

PR #529, opened 2023-08-14
Fix order of SCSS imports for radio field

sergei-maertens:
Partly closes open-formulieren/open-forms#3327

The radio background styles were being overridden by the form-field overrides (open-forms theme specific) due to the CSS precedence order.

Forcing the radio styles to come after the form-field styles resolves this, but it's not an ideal fix. The proper long-term solution is to have a separate stylesheet with our open-forms theme specific overrides rather than baking this into the main bundle.


CharString (commented):
LFTM: I don't understand the code that was removed.
But I do expect Chromatic changes, if this is a visual fix.

sergei-maertens:
> LFTM: I don't understand the code that was removed. But I do expect Chromatic changes, if this is a visual fix.

Storybook loads the styles slightly different, which caused the problem not to occur, therefore there are no visual changes!

There wasn't really code being removed, just different places where the sass is imported. I was hoping we could bundle styles and JS together, but unfortunately that causes some issues :(

PR #530, opened 2023-08-16
Support statement checkboxes per form

sergei-maertens:
Closes open-formulieren/open-forms#3372

Reworked statement checkboxes to not hit the global configuration endpoint, but instead read the configuration from the form data retrieved from the form detail endpoint.

This also sets up the accessible message texts which were a bit too generic.

### DIFF
--- a/src/Context.js
+++ b/src/Context.js
@@ -10,6 +10,7 @@ const FormContext = React.createContext({
   maintenanceMode: false,
   showProgressIndicator: true,
   submissionAllowed: 'yes',
+  submissionStatementsConfiguration: [],
   literals: {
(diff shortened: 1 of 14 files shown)


src/Context.js
```diff
@@ -10,6 +10,7 @@ const FormContext = React.createContext({
   maintenanceMode: false,
   showProgressIndicator: true,
   submissionAllowed: 'yes',
+  submissionStatementsConfiguration: [],
```
SilviaAmAm, line 13: Same remark about the naming as on the backend PR :smile:

SilviaAmAm (approved):

PR #533, opened 2023-08-21
Make amount field readonly when multiple products is not supported.

sergei-maertens:
Closes open-formulieren/open-forms#3299


CharString (approved):
LGTM Can I approve the storybook tests too?
And is this already managed in the Context?

PR #534, opened 2023-08-21
Reset session storage / cache on session expiry reset

### LINKED ISSUE #505: Appointment form - ensure session expiration is handled correctly

sergei-maertens:
Closes #505

Writing this test was *hard*

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -20,7 +20,7 @@ const SessionExpiresInHeader = 'X-Session-Expires-In';
 
 let sessionExpiresAt = createGlobalstate({expiry: null});
 
-const updateSesionExpiry = seconds => {
+export const updateSessionExpiry = seconds => {
   const newExpiry = new Date();
   newExpiry.setSeconds(newExpiry.getSeconds() + seconds);
   sessionExpiresAt.setValue({expiry: newExpiry});
@@ -97,7 +97,7 @@ const addHeaders = (headers, method) => {
 const updateStoredHeadersValues = headers => {
   const sessionExpiry = headers.get(SessionExpiresInHeader);
   if (sessionExpiry) {
-    updateSesionExpiry(parseInt(sessionExpiry), 10);
+    updateSessionExpiry(parseInt(sessionExpiry), 10);
   }
 
   const CSRFTokenValue = headers.get(CSRFToken.headerName);
--- a/src/components/appointments/CreateAppointment/CreateAppointment.spec.js
+++ b/src/components/appointments/CreateAppointment/CreateAppointment.spec.js
@@ -1,0 +1,118 @@
+import {act, render, screen, waitFor} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import messagesEN from 'i18n/compiled/en.json';
+import {IntlProvider} from 'react-intl';
+import {RouterProvider, createMemoryRouter} from 'react-router-dom';
+
+import {ConfigContext, FormContext} from 'Context';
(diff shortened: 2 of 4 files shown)


src/api.js
```diff
@@ -20,7 +20,7 @@ const SessionExpiresInHeader = 'X-Session-Expires-In';
 
 let sessionExpiresAt = createGlobalstate({expiry: null});
 
-const updateSesionExpiry = seconds => {
+export const updateSesionExpiry = seconds => {
```
CharString, line 23: At first I wasn't delighted to see this exported, but I guess that state pool was exposed anyway.

My concern was that I say myself chasing timeouts because somewhere, some component mutates the expiry time outside of the fetch requests flow.

Maybe fix the typo... `updateSessionExpiry`

src/components/appointments/CreateAppointment/CreateAppointment.js
```diff
@@ -49,6 +49,8 @@ const CreateAppointment = () => {
   const reset = () => {
     clearSubmission();
     resetSession();
+    // clear any cached data or lingering submission state
+    window.sessionStorage.clear();
```
CharString, line 53: Clearing the whole sessionStorage doesn't have any unwanted side-effects? This is per browser-tab, right?

CharString (changes requested):

src/components/appointments/CreateAppointment/CreateAppointment.spec.js
```diff
@@ -0,0 +1,118 @@
...
+    // console.log(Object.entries(sessionStorage));
+
+    // introspect session storage
+    await waitFor(() => {
+      const formData = sessionStorage.getItem('appointment|formData');
+      expect(formData).toBe('{}');
+    });
+    await screen.findByText('Select your product(s)');
+
+    const productDropdown = await screen.getByRole('combobox');
```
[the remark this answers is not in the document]
sergei-maertens: ah this was not intentional!

src/api.js
sergei-maertens: Yeah it hurt me too, I wish there was a mechanism to limit exports to tests :(

Re-implementing it directly on the state pool seemed like a bigger evil.

> My concern was that I say myself chasing timeouts because somewhere, some component mutates the expiry time outside of the fetch requests flow.

This already is a "problem" - the session cookie already is a global state so managing the expiry in the app also has to be global. It's a side effect of every API call. I do have a draft issue on the board to refactor the session expiry component into it's own explicit route to simplify/decouple things a bit more.

Will fix the typo!

src/components/appointments/CreateAppointment/CreateAppointment.js
sergei-maertens: It is per browser tab indeed, and it should not have any unwanted side effects. The react-use hooks even deal with browser storage API's being blocked an in that case all state is purely in memory (using `useState`). So we should *never* assume that sessionStorage is guaranteed to be present.

CharString (approved):

PR #535, opened 2023-08-22
Display backend processing error

### LINKED ISSUE #506: Appointment form - handle backend processing error state
On the confirmation page, it's possible the status reported by the backend returns errors (`onProcessingFailure` prop, which receives an error message).

Desired flow:

1. Fill out appointment form
2. Submit the summary page
3. Wait for loader on confirmation to resolve
4. When the status is error, redirect the user back to the summary page and display the error message.

Additionally, when re-submitting the data, the backend now has an IntegrityError because the existing appointment record is not updated and the backend tries to create another one for the same submission.

sergei-maertens:
Closes #506  - depends on #534

### DIFF
--- a/src/api-mocks/submissions.js
+++ b/src/api-mocks/submissions.js
@@ -77,3 +77,20 @@ export const mockSubmissionProcessingStatusPendingGet = rest.get(
       })
     )
 );
+
+export const mockSubmissionProcessingStatusErrorGet = rest.get(
+  `${BASE_URL}submissions/:uuid/:token/status`,
+  (req, res, ctx) =>
+    res(
+      ctx.json({
+        status: 'done',
+        result: 'failed',
+        errorMessage: 'Computer says no.',
+        publicReference: '',
+        confirmationPageContent: '',
+        reportDownloadUrl: '',
+        paymentUrl: '',
+        mainWebsiteUrl: '',
+      })
+    )
+);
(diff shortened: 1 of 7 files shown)


src/api-mocks/submissions.js
```diff
@@ -77,3 +77,20 @@ export const mockSubmissionProcessingStatusPendingGet = rest.get(
       })
     )
 );
+
+export const mockSubmissionProcessingStatusErrorGet = rest.get(
```
CharString, line 81: When you ask Yoda, instead of GPT to name your things. :)

Consistent it is. Like it, I do.

Beats spelling out "Request" on every mock, as it's implied.

src/components/appointments/CreateAppointment/Summary.spec.js
```diff
@@ -18,22 +18,37 @@ import {
...
+      location: '1396f17c',
+      date: '2023-07-12',
+      datetime: `2023-07-12T08:00:00Z`,
+    },
+    contactgegevens: {
+      lastName: 'Kundera',
+      dateOfBirth: '1929-04-01',
+      email: 'milan@kundera.cz',
+      phone: '12345678',
+      bsn: '123456782',
```
CharString, line 39: I'd expect an Eidas psuedo ID here. :stuck_out_tongue_winking_eye:  Sorry for being so unbearable.

CharString (approved):
A more meta question I get from reviewing React code is that: given that react components are supposed to be pure functions, why do our tests of them end up looking so complex?

PR #536, opened 2023-08-22
Handle invalid states when accessing URLs directly

### LINKED ISSUE #508: Appointment form - handle invalid states at URLs
There are various ways that you can go to 'step' URLs in the appointment form while the expected state is not in place yet:

* manually entering urls such as `.../kalender`
* on the confirmation screen clicking the progress indicator links

The products, calendar, contact details and summary pages require a submission and appointment data to be in the state. If this is not the case, they should redirect back to the start of the form.

The confirmation screen requires the `statusUrl` parameter (now it throws when it's missing) -> it should be verified & tested that the error boundary is sufficient.

sergei-maertens:
Classic OWASP thing, fiddling with the URL and trying to bypass checks :)

Closes #508, depends on #535

### DIFF
--- a/src/components/appointments/ContactDetailsStep.js
+++ b/src/components/appointments/ContactDetailsStep.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import React, {useContext, useMemo} from 'react';
 import {flushSync} from 'react-dom';
 import {FormattedMessage, useIntl} from 'react-intl';
-import {useNavigate} from 'react-router-dom';
+import {Navigate, useNavigate} from 'react-router-dom';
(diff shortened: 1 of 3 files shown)


CharString (approved):
What's the use of a route if I can't bookmark it or navigate with it?

hmmm OWASP, interesting way of framing it: a backend in the frontend.

sergei-maertens:
> What's the use of a route if I can't bookmark it or navigate with it?
> 
> hmmm OWASP, interesting way of framing it: a backend in the frontend.

So you're saying that HTTP 3XX is now obsolete? :wink:

PR #544, opened 2023-09-04
Pre-select product via query string

sergei-maertens:
Closes open-formulieren/open-forms#3300

### DIFF
--- a/src/components/appointments/CreateAppointment/CreateAppointment.mdx
+++ b/src/components/appointments/CreateAppointment/CreateAppointment.mdx
@@ -11,9 +11,29 @@ The create appointment UI is used when `form.appointmentOptions.isAppointment` i
 brings the user in a (mostly) fixed form step progression to select their products, appointment
 details and provide contact details.
 
-**NOTE**
+## Features
 
-This is currently under development.
+**Multi vs. single product flow**
+
+The backend exposes information whether the enabled plugin for appointments supports multi-product
+appointments or not. This drives the appointment flow - for multi-product appointments, you can add
+additional products and the list of available products is fetched while passing the already selected
+products to the backend - this allows only valid combinations to be returned.
+
+For single-product, the UI is slimmed down a little bit to remove the multi-product controls, and
+the "amount" field is fixed to the number "1".
+
+**Pre-selecting a product**
+
+You can link to the appointments flow with an optional query string parameter `product`, which
+should have the value of a valid product ID. If provided, that product will be pre-selected in the
+list of all products. Note that this only applies to the first product in the case of multi-product
+forms.
+
+For example, the following URLs will pre-select a product:
+
+- `https://example.com/appointment/?product=my-product-id`, for the regular routing
+- `https://example.com/#/appointment/?product=my-product-id`, for hash-based routing
 
 <Story of={CreateAppointmentStories.Default} />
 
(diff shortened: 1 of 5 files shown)


src/components/appointments/CreateAppointment/CreateAppointment.mdx
```diff
@@ -11,9 +11,28 @@ The create appointment UI is used when `form.appointmentOptions.isAppointment` i
...
+appointments or not. This drives the appointment flow - for multi-product appointments, you can add
+additional products and the list of available products is fetched while passing the already products
+to the backend - this allows only valid combinations to be returned.
+
+For single-product, the UI is slimmed down a little bit to remove the multi-product controls, and
+the "amount" field is fixed to the number "1".
+
+**Pre-selecting a product**
+
+You can link to the appointments flow with an optional querystring parameter `product`, which should
```
CharString, line 28:
```suggestion
You can link to the appointments flow with an optional query string parameter `product`, which should
```
It's two word in English.

src/components/appointments/CreateAppointment/CreateAppointment.mdx
```diff
@@ -11,9 +11,28 @@ The create appointment UI is used when `form.appointmentOptions.isAppointment` i
...
 
-**NOTE**
+## Features
 
-This is currently under development.
+**Multi vs. single product flow**
+
+The backend exposes information whether the enabled plugin for appointments supports multi-product
+appointments or not. This drives the appointment flow - for multi-product appointments, you can add
+additional products and the list of available products is fetched while passing the already products
```
CharString, line 20:
```suggestion
additional products and the list of available products is fetched while passing the already selected products
```

src/components/appointments/CreateAppointment/CreateAppointment.mdx
```diff
@@ -11,9 +11,28 @@ The create appointment UI is used when `form.appointmentOptions.isAppointment` i
...
+**Pre-selecting a product**
+
+You can link to the appointments flow with an optional querystring parameter `product`, which should
+have the value of a valid product ID. If provided, that product will be pre-selected in the list of
+all products. Note that this only applies to the first product in the case of multi-product forms.
+
+For example, the following URLs will pre-select a product:
+
+- `https://example.com/appointment/?product=my-product-id`, for the regular routing
+- `https://example.com/#/appointment/?product=my-product-id`, for hash-based routing
```
CharString, line 35: Maybe instead of a wordy explanation 2 stories named Multi and single product flow, that show the two cases is clearer. E.g. it clears up whether multi-products is about putting multiple different products "in your basket" or whether it's about providing choice of multiple products in a drop down.

CharString (changes requested):
I don't recognise hash based routing tests, is that something we get for free from some library by setting `REACT_APP_USE_HASH_ROUTING=true` in the env?

src/components/appointments/CreateAppointment/CreateAppointment.mdx
sergei-maertens: These stories already exist: https://open-formulieren.github.io/open-forms-sdk/?path=/story/private-api-appointments-product-and-amount--no-multiple-products-support and for the full flow there is a an arg + control to toggle between modes. 
image

sergei-maertens:
> I don't recognise hash based routing tests, is that something we get for free from some library by setting `REACT_APP_USE_HASH_ROUTING=true` in the env?

It's something we handle ourselves in `sdk.js` by selecting a different router:

```js
    const createRouter = this.useHashRouting ? createHashRouter : createBrowserRouter;
    const router = createRouter(routes, {basename: this.basePath});
```

so that makes it impossible to test, since we use `createMemoryRouter` in tests. I've verified it manually and I suppose that's the best we can do, short of an end-to-end tests with a real backend.

CharString (approved):

PR #545, opened 2023-09-04
[OF#1884] Fine grained custom errors for time components

SilviaAmAm:
Fixes https://github.com/open-formulieren/open-forms/issues/1884
Screenshot from 2023-09-04 15-17-57

### DIFF
--- a/src/formio/validators/MinMaxTimeValidator.js
+++ b/src/formio/validators/MinMaxTimeValidator.js
@@ -1,4 +1,12 @@
+import set from 'lodash/set';
 import moment from 'moment';
+
+const ERROR_CASES = {
+  ONLY_MIN: 'onlyMin',
+  ONLY_MAX: 'onlyMax',
+  MIN_SMALLER_THAN_MAX: 'minSmallerThanMax',
+  MAX_SMALLER_THAN_MIN: 'maxSmallerThanMin',
+};
 
 const validateTimeBoundaries = (minBoundary, maxBoundary, timeValue) => {
   const minTime = minBoundary ? moment(minBoundary, 'HH:mm:ss') : null;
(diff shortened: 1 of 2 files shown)


src/formio/validators/MinMaxTimeValidator.js
```diff
@@ -40,7 +76,16 @@ const MinMaxTimeValidator = {
...
-    return validateTimeBoundaries(component.component.minTime, component.component.maxTime, value);
+    const {isValid, error} = validateTimeBoundaries(
+      component.component.minTime,
+      component.component.maxTime,
+      value
+    );
+
+    if (!isValid) {
+      set(component, 'openForms.validationErrorContext', {timeError: error});
+    }
```
sergei-maertens, line 87: I'm guessing that the validator instance is only created once and you can't put this on the `this` object?

In this form, wouldn't there also be a race condition risk if multiple validators run in parallel (due to `await` syntax being used or `validateAsync` being called)? What happens if the component is part of a repeating group, does this create separate component instances or is this context shared between all of them?

SilviaAmAm: > I'm guessing that the validator instance is only created once and you can't put this on the this object?

Yes, this is why I put it in the component!

> wouldn't there also be a race condition risk if multiple validators run in parallel (due to await syntax being used or validateAsync being called)

I don't think so :thinking: For the time component there is only the minTime/maxTime validation (none of the async validators apply, and there aren't any other validators available). If we were to use the same approach for other components that have more validators, then maybe the key of the errors in the context should be based on the validator type rather than the component type (i.e. `MinMaxTimeValidatorError` instead of `timeError`?) :thinking: 

> What happens if the component is part of a repeating group, does this create separate component instances or is this context shared between all of them?

It goes in the context of the component (it is not shared by other components in the same group).
Screenshot from 2023-09-06 10-33-39

sergei-maertens (approved):

PR #547, opened 2023-09-06
:truck: Group appointment components for more logical code organization

sergei-maertens:
Flat is better than nested but this was getting borderline unmaintainable.


CharString (approved):
Sure, if you and CI agree. :)

PR #548, opened 2023-09-11
[OF#3443] Date validation

SilviaAmAm:
Fixes open-formulieren/open-forms#3443

Depends on https://github.com/open-formulieren/open-forms/pull/3479

I give up on the tests :cry:

### DIFF
--- a/src/formio/components/DateField.stories.js
+++ b/src/formio/components/DateField.stories.js
@@ -1,3 +1,6 @@
+import {expect} from '@storybook/jest';
+import {userEvent, within} from '@storybook/testing-library';
+
 import {withUtrechtDocument} from 'story-utils/decorators';
 
 import {SingleFormioComponent} from './story-util';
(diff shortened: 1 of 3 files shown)


src/formio/components/DateField.stories.js
```diff
@@ -38,5 +41,61 @@ export const DateField = {
...
+        maxDate: null,
+      },
+    },
+  },
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+
+    const dateInput = canvas.getByRole('textbox');
+
+    userEvent.type(dateInput, '06-06-2006');
```
sergei-maertens, line 59:
```suggestion
    await userEvent.type(dateInput, '06-06-2006');
```
this is async iirc

src/formio/components/DateField.stories.js
```diff
@@ -38,5 +41,61 @@ export const DateField = {
...
+        dateMinMax: true,
+      },
+    },
+  },
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+
+    const dateInput = canvas.getByRole('textbox');
+
+    userEvent.type(dateInput, '06-06-2006');
```
sergei-maertens, line 94:
```suggestion
    await userEvent.type(dateInput, '06-06-2006');
```

src/formio/components/DateField.stories.js
```diff
@@ -38,5 +41,61 @@ export const DateField = {
...
+    await expect(dateInput).toHaveDisplayValue('06-06-2006');
```
sergei-maertens, line 95:
```suggestion
    expect(dateInput).toHaveDisplayValue('06-06-2006');
```
and this one is sync :)

src/formio/components/DateField.stories.js
```diff
@@ -38,5 +41,61 @@ export const DateField = {
...
+
+    // TODO: I cannot get this to work. If you do it manually in storybook, it works... (it shows the error).
+    // const error = canvas.queryByText('minDate');
+    // await expect(error).not.toBeNull();
```
sergei-maertens, line 99: could be because of the missing `await`?

Have you tried `const error = await canvas.findByText('minDate')`? Possibly wrap it in `await waitFor(...);`? Probably the answer is yes :P

src/formio/validators/minMaxDateValidator.js
sergei-maertens, line 1: Perhaps we can add some jest unit tests for the validator itself so we at least have coverage on them?

src/formio/components/DateField.stories.js
SilviaAmAm: From the docs I saw that it is async if you use the delay option (which then adds a delay between each key pressed).

However, with the async behaviour, it looks like the flatpickr instance is reinitialised and so you loose what was filled in :S

src/formio/validators/minMaxDateValidator.js
SilviaAmAm: This is a good idea, I will try!

src/formio/components/DateField.stories.js
SilviaAmAm: I tried, but I couldn't get it to work :(

src/formio/components/DateField.stories.js
sergei-maertens: towel, ring... etc :) let's leave it at that

src/formio/components/DateField.stories.js
sergei-maertens: date pickers are evil :)

PR #549, opened 2023-09-15
[OF#3414] Change co-sign section concerning styling

vaszig:
Fixes open-formulieren/open-forms#3414

I have chosen to add a padding to the whole div and align it with the buttons above. If you think it would be better to align all the buttons and not the div please let me know.

image

### DIFF
--- a/src/components/LoginOptions/LoginOptionsDisplay.js
+++ b/src/components/LoginOptions/LoginOptionsDisplay.js
@@ -37,18 +37,16 @@ const LoginOptionsDisplay = ({
       )}
 
       {cosignLoginInfo && (
-        <>
-          <div className={getBEMClassName('login-options__cosign')}>
-            <h2 className={getBEMClassName('login-options__caption')}>
-              <FormattedMessage
-                description="Log in to co-sign the form title"
-                defaultMessage="Log in to co-sign the form"
-              />
-            </h2>
+        <div className={getBEMClassName('login-options__cosign')}>
+          <h2 className={getBEMClassName('login-options__caption')}>
+            <FormattedMessage
+              description="Log in to co-sign the form title"
+              defaultMessage="Log in to co-sign the form"
+            />
+          </h2>
 
-            <LoginButton key={cosignLoginInfo.identifier} option={cosignLoginInfo} />
-          </div>
-        </>
+          <LoginButton key={cosignLoginInfo.identifier} option={cosignLoginInfo} />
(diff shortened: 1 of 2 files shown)


src/components/LoginOptions/LoginOptionsDisplay.js
```diff
@@ -38,14 +38,16 @@ const LoginOptionsDisplay = ({
 
       {cosignLoginInfo && (
         <>
```
sergei-maertens, line 40: this outer fragment can be removed since everything is contained in a div now

src/scss/components/_login-options.scss
```diff
@@ -16,3 +16,8 @@
     @include margin(auto);
   }
 }
+
+.#{prefix(login-options__cosign)} {
+  padding: 0.6em;
+  background-color: var(--of-color-read-only-bg);
```
sergei-maertens, line 22: design tokens should always be made for the component itself, you are not allowed to use design tokens of other components (or utility tokens) for a specific goal.

So the correct approach here would be:

```suggestion
  background-color: var(--of-login-options-cosign-background-color);
```

Then, you can define that design token in https://github.com/open-formulieren/design-tokens, get a new version of that package published and use it in the SDK.

src/scss/components/_login-options.scss
```diff
@@ -16,3 +16,8 @@
     @include margin(auto);
   }
 }
+
+.#{prefix(login-options__cosign)} {
```
sergei-maertens, line 20: this belongs in the code block above, instead of making these BEM classnames manually, you should use `@include bem.element('cosign')` instead and define the style rules in there.

src/scss/components/_login-options.scss
```diff
@@ -16,3 +16,8 @@
...
+  padding: 0.6em;
```
sergei-maertens, line 21: magic numbers are not themeable - can you please use a design token for this and set a value in the design tokens repository?

the following will do:

```scss
padding-block: var(--of-login-options-cosign-padding-block);
padding-inline: var(--of-login-options-cosign-padding-inline);
```

Also note the use of `block` and `inline` rather than top/bottom/left/right - as those naturally deal correctly with LTR/RTL and horizontal/vertical text writing modes.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #550, opened 2023-09-15
:lipstick: Fix width of progress indicator

sergei-maertens:
The bar was not taking up full width due to the negative margin-left and margin-right values, combined with the width: 100% of the parent. Setting the width to calc(100% + 2 * margin) results in a horizontal scrollbar, but unsetting the width fixes the styling on mobile while not destroying it on desktop.


SilviaAmAm (approved):

sergei-maertens:
Backports: 994ecc36249bd334c4cfa540659d92fec9190a77, a66e1b99cbf0cc878e6d96b47ea0e9cb484762f6 and c9612d2962cb05b75e3b59600cacaa2d2899fd96

PR #553, opened 2023-09-21
 [OF#3383] Refactor phone number validation

vaszig:
Fixes open-formulieren/open-forms#3383

### DIFF
--- a/src/formio/validators/plugins.js
+++ b/src/formio/validators/plugins.js
@@ -2,12 +2,6 @@ import {post} from '../../api';
 
 export const pluginsAPIValidator = {
   key: `validate.backendApi`,
-  message(component) {
-    return component.t(component.errorMessage('Invalid'), {
-      field: component.errorLabel,
-      data: component.data,
-    });
-  },
   check(component, setting, value) {
     if (!value) return true;
 
@@ -21,13 +15,15 @@ export const pluginsAPIValidator = {
         return valid ? true : response.data.messages.join('<br>');
       });
     });
-    return Promise.all(promises).then(results => {
-      const anyValid = results.some(result => result === true);
+    return Promise.all(promises)
+      .then(results => {
+        const anyValid = results.some(result => result === true);
 
-      if (anyValid) return true;
+        if (anyValid) return true;
 
-      return results.join('<br>');
-    });
+        return results.join('<br>');
+      })
+      .catch(() => false);
   },
 };
 
--- a/src/formio/validators/plugins.js
+++ b/src/formio/validators/plugins.js
@@ -1,41 +1,34 @@
 import {post} from '../../api';
 
-const errorMessageMap = {
-  'kvk-kvkNumber': 'Invalid Kvk Number',
-  'kvk-rsin': 'Invalid RSIN',
-  'kvk-branchNumber': 'Invalid Branch Number',
-  'phonenumber-international': 'Invalid international phonenumber',
-  'phonenumber-nl': 'Invalid Dutch phonenumber',
-};
+export const pluginsAPIValidator = {
+  key: `validate.backendApi`,
+  message(component) {
+    return component.t(component.errorMessage('Invalid'), {
+      field: component.errorLabel,
+      data: component.data,
+    });
+  },
+  check(component, setting, value) {
+    if (!value) return true;
 
-const pluginAPIValidator = plugin => {
-  let defaultMsg = errorMessageMap[plugin];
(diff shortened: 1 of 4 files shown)


src/formio/validators/plugins.js
```diff
@@ -1,41 +1,34 @@
...
-        .then(response => {
-          const valid = response.data.isValid;
-          return valid ? true : response.data.messages.join('<br>');
-        })
-        .catch(() => false);
-    },
-  };
+      if (anyValid) return true;
+
+      return results.join('<br>');
```
SilviaAmAm, line 29: When testing I don't see the new line :thinking: 
errors

src/formio/validators/plugins.js
```diff
@@ -1,41 +1,34 @@
...
+    });
```
SilviaAmAm, line 30:
```suggestion
    }).catch(() => false);
```
I think this was removed by accident?

src/formio/validators/plugins.js
```diff
@@ -1,41 +1,34 @@
...
-};
+export const pluginsAPIValidator = {
+  key: `validate.backendApi`,
+  message(component) {
+    return component.t(component.errorMessage('Invalid'), {
+      field: component.errorLabel,
+      data: component.data,
+    });
+  },
+  check(component, setting, value) {
```
SilviaAmAm, line 11: In the time validator, we used the pattern:
- `check` returns true/false and sets on the component (under `openForms.validationErrorContext`) a list of error keys.
- `message` processes what is in the `openForms.validationErrorContext` and then decides what error to display.

Not sure if this is good for this case, since the validator returns error messages rather than error messages keys :thinking:

src/jstests/formio/components/fixtures/phonenumber.js
```diff
@@ -0,0 +1,15 @@
...
+  key: 'phonenumber',
+  type: 'phoneNumber',
+  input: true,
+  validate: {
+    backendApi: true,
+    plugins: ['phonenumber-international', 'phonenumber-nl'],
+  },
+};
+
+const validSamples = ['630123456', '+31630123456'];
```
SilviaAmAm, line 12:
```suggestion
const validSamples = ['0630123456', '+31630123456'];
```
Doesn't the first one need a 0?

src/jstests/formio/validators/pluginapivalidator.spec.js
SilviaAmAm, line 1: Maybe also good to write a test for the `message` method of the validator, so that it is also covered?

src/jstests/formio/validators/mocks.js
```diff
@@ -0,0 +1,31 @@
+import {rest} from 'msw';
+
+import {BASE_URL} from 'api-mocks';
+
+const INTERNATIONAL_VALIDATION_ENDPOINT = `${BASE_URL}validation/plugins/phonenumber-international`;
+const DUTCH_VALIDATION_ENDPOINT = `${BASE_URL}validation/plugins/phonenumber-nl`;
+
+export const mockInternationalValidPost = rest.post(
```
sergei-maertens, line 8: I'd prefer to have the validator names in here - `mockValidInternationalPhonenumberPost` - if we add other validators in the future, then the names won't collide.

You could also namespace this:

```js
export const phoneNumberValidations = {
  internationalValidPost: ...,
  internationalInvalidPost: ...,
  dutchValidPost: ...,
  dutchInvalidPost: ...,
};
```

src/formio/validators/plugins.js
vaszig: Discussed with @SilviaAmAm  and @sergei-maertens - a separate issue(https://github.com/open-formulieren/open-forms-sdk/issues/558) has been created for this.

src/jstests/formio/validators/pluginapivalidator.spec.js
vaszig: As I told to Silvia earlier, I don't see the usage of `message` method here. I tested this with all the validators enabled and still doesn't go through this piece of code. The only `message` method which is used is the one of the `PhoneNumberValidator` object in the `PhoneNumberField.js`  (you can see the results below.)

If you disagree, of course I can add a test about it.

image

sergei-maertens: Let's wrap this up and leave it for what it is. Nobody gets any happier working on these parts of the codebase

sergei-maertens (approved):

PR #559, opened 2023-09-29
Fix flickering inputs on Webkit browsers

sergei-maertens:
Closes open-formulieren/open-forms#3511

Some difference in the JS engine in WebKit (the browser engine behind Safari or any other browser on iOS/MacOS) vs. Gecko (Firefox) and Blink (Chromium, Edge) seems to cause a different (debounce) behaviour and or event firing/acting on events w/r to the Formio renderer.

On Webkit, the formio initialized event triggered with mismatching backend data, causing the form field inputs to be cleared and the submission to be updated, which in turn triggers the logic check again due to a change event being fired. This then leads again to a change in the formio configuration and the formio renderer re-initializing, ad infinitum. Somewhere the form data and form configuration state goes out of sync and it's incredibly hard to manage this, especially due to the large number of related issues in the past.

Since we use the formio initialized event pretty much only to form step data from earlier submissions or when navigating between steps, we avoid setting updated form data at all in formio initialized events triggered by logic evaluations. The logic evaluation itself already manages the form instance submission data, in case that the backend returns calculated/derived values, for example.

I truly hope this doesn't break anything else, but due to the lack of tests (granted, it's near impossible to test this mess) and complexity of this part of the code base, I can't say I'm very confident in this fix.

### DIFF
--- a/src/components/FormStep/index.js
+++ b/src/components/FormStep/index.js
@@ -180,6 +180,7 @@ class AbortedLogicCheck extends Error {
 const initialState = {
   configuration: null,
   backendData: {},
+  formioInitialized: false,
   canSubmit: false,
   logicChecking: false,
   isFormSaveModalOpen: false,
@@ -212,6 +213,7 @@ const reducer = (draft, action) => {
       } = action.payload;
       draft.configuration = configuration;
       draft.backendData = data;
+      draft.formioInitialized = false;
       draft.canSubmit = canSubmit;
       draft.logicChecking = false;
       draft.isNavigating = false;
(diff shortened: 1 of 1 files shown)


sergei-maertens:
Needs backporting to 1.4.x and 1.5.x - 1.3.x is React 17 based, which doesn't seem to have this bug. The React 18 upgrade doing the batching of state updates is suspected to not play nice with the formio events.

sergei-maertens:
Backports: e38931fa0131711f7e3a7c3f1addd4607c662a8f, 9b68e061401bb3ce92a1a046439f12be9eae795f

PR #562, opened 2023-10-06
[OF#3510] Fix overlapping address under map component

vaszig:
Fixes open-formulieren/open-forms#3510


vaszig:
I haven't managed to solve this by defining different position in the elements. I added the flex property to the map and below you can see the results:

### Desktop

desktop

### Mobile

mobile

sergei-maertens (approved):

PR #564, opened 2023-10-13
[#462] Refactor button

### LINKED ISSUE #462: Refactor buttons to use utrecht-button + extensions
The current `.openforms-button` styles are preventing proper use of the existing NL DS button component (from Utrecht), due to a number of hardcoded styles and the use of `!important`.

This needs to be rebuilt around the Utrecht components, while any Open Forms-specific theming needs to be gated behind a `--openforms` extension or similar.

This affects:

* Buttons
* LoginButton / login options
* Buttons in the edit group component (add another, remove)

Important: the cookie pages in the backend may not break, so possibly/likely classnames need to be updated there too!

SilviaAmAm:
Fixes #462 

To do:
- [x] Deal with the submit button at the end of the form
- [x] Deal with stories
- [x] Deal with jest tests
- [x] Deal with button borders
- [x] Remove old button styles and change the Button story that displays them
- [x] Disabled button aria
- [x] Check buttons on start page
- [x] Add instructions on what tokens need to set to get the new styles

Tokens PR: https://github.com/open-formulieren/design-tokens/pull/35 https://github.com/open-formulieren/design-tokens/pull/37


src/components/Button/ButtonWithAccessibleDisabledState.js
```diff
@@ -0,0 +1,19 @@
+import {Button as UtrechtButton} from '@utrecht/component-library-react';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+// Temporary until the aria-disabled is set on the Utrecht button
+const ButtonWithAccessibleDisabledState = ({disabled, children, ...extraProps}) => {
+  return (
+    <UtrechtButton disabled={disabled} aria-disabled={disabled} {...extraProps}>
+      {children}
+    </UtrechtButton>
```
SilviaAmAm, line 10: This will also have the disabled attribute set :/

sergei-maertens: I'm wondering if we instead should still have our own `Button` component that wraps `UtrechtButton`, but all it does is grab the `disabled` prop and turn it into `aria-disabled="true"`, install an `onClick` handler to prevent actual submitting and maybe manually set the disabled utrecht class name (if that's relevant).

So then we'd still everywhere import our "own" button, but it's just a thin shell addressing limitations in the upstream component.

src/scss/components/_button.scss
```diff
@@ -1,229 +1,75 @@
...
-      @if $color-background-hover {
-        background-color: $color-background-hover;
-      }
-
-      @if $color-border-hover {
-        // TODO: can be removed once there are utrecht design tokens for the bottom/right
-        // border color
-        border-right-color: $color-border-hover !important;
-        border-bottom-color: $color-border-hover !important;
+// TODO: can be removed once there are utrecht design tokens for the bottom/right border color
```
sergei-maertens, line 6: these design tokens are not going to happen

the linked issue is about other design tokens

src/scss/components/_button.scss
```diff
@@ -1,229 +1,75 @@
...
+// issue: https: //github.com/nl-design-system/utrecht/issues/1406
+.utrecht-button {
```
sergei-maertens, line 8: these styles now affect all themes - I'd instead opt for a simpler approach where we configure the design tokens for full borders and then apply CSS overrides scoped to our own theme:

```scss
.openforms-theme {
  .utrecht-button {
    border-block-start-color: transparent;
    border-inline-start-color: transparent;
  }
}
```

etc.

SilviaAmAm:
Note: this needs the Tokens PR to be merged and the new version of the design-tokens package to be released + upgraded

src/components/Button/OFButton.js
```diff
@@ -0,0 +1,29 @@
...
+  if (disabled) otherProps.className = 'utrecht-button--disabled';
+
+  const onClick = event => {
+    if (disabled) event.preventDefault();
+
+    if (onClickHandler) onClickHandler(event);
+  };
+
+  return (
+    <UtrechtButton aria-disabled={disabled} onClick={onClick} {...otherProps}>
```
sergei-maertens, line 18: is the `aria-disabled` properly stringified here? It's a bit weird that we're passing a boolean but the rendered HTML should be the string `"true"` or `"false"`.

sergei-maertens (approved):
Let's see what Chromatic has to say :grimacing:

PR #565, opened 2023-10-16
Implement cursor start/end on Home/End keypress

### LINKED ISSUE #437: New Select component: home/end keys have no effect on search input
Expected behaviour: pressing <kbd>Home</kbd> resets the cursor to the start of the input, while <kbd>End</kbd> moves the cursor to the end.

Currently the keys have no effect at all.

sergei-maertens:
Closes #437

Solution taken from react-select issue/discussion.

There is some ambiguity on what the behaviour should be on MacOS, but due to lack of appropriate device we're leaving that as it is, for now.

### DIFF
--- a/src/components/forms/SelectField/SelectField.js
+++ b/src/components/forms/SelectField/SelectField.js
@@ -55,6 +55,25 @@ const SelectField = ({
     [autoSelectOnlyOption, isSingle, value, JSON.stringify(options), setValue, valueProperty]
   );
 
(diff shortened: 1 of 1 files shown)


CharString (approved):
I know of 3 colleagues running MacOS as their daily driver. I've only had 2 Macintoshes and the last one I ran Linux on after support for my MacOS ran out and I had to upgrade.

PR #567, opened 2023-10-18
[OF#3523] Fix cosign when no privacy concent enabled

vaszig:
Fixes https://github.com/open-formulieren/open-forms/issues/3523


sergei-maertens:
Checked if 1.5.x or 1.3.x are affected to be sure, but the code is vastly different (and doesn't exist in 1.3.x) so no extra backports needed.

sergei-maertens (approved):

PR #569, opened 2023-10-19
Fix cascade of requests firing

### LINKED ISSUE #546: Appointments: multi-product makes duplicate requests
The fetch-products hook is getting called multiple times and with every additional product, the amount of calls increases.

There's probably something broken in the hooks dependencies.

Maybe this would be the time to introduce tanstack query as it should handle caching/de-duplication out of the box?

sergei-maertens:
Closes #546

Hook dependencies operate on the identity, and the identity of the selectedProducts changes every time on each render, while the contents stay the same.

JSON.stringify is the recommended approach to mitigate using non-primivites in hook dependency arrays.

Note that the caching masked the issue partially - it caused the request cascade to stop, but the problem still showed with multi- product appointments.


SilviaAmAm (approved):

sergei-maertens:
Backport: 78b1b76889c77271906b3816e3db3f3fa48414bf

1.4.x and older are not affected, new appointments flow is since 1.5.x

PR #572, opened 2023-10-20
Remove need for custom form-control CSS

### LINKED ISSUE #571: Refactor out openforms-form-control CSS
I took a quick look at the current state - while the React side of things is fairly straightforward to get rid of openforms-form-control wrapper (the `utrecht-form-field` fulfills the same role), the Formio templates are a different beast. It requires the `ref="component"` to be present, but for that we need to move the (validation error) message container inside the `utrecht-form-field` container. Applying this container in the `component.ejs` is likely a bit too optimistic because radio groups/checkboxes use other container classes :thinking:

sergei-maertens:
Closes #571

Goal: getting rid of custom CSS for the form-control wrapper, which is arguably already performed by the `utrecht-form-field` component.

There are some challenges, in particular because formio requires the `ref="component"` to be present and still applying the relevant styles & design tokens.


sergei-maertens:
Marked ready for review to get visual regression testing, this is by no means complete :sweat_smile: 

AKA "Chromatic take the wheel"

src/components/forms/Label.js
```diff
@@ -13,7 +13,7 @@ export const LabelContent = ({id, disabled = false, isRequired = false, type, ch
       htmlFor={id}
       disabled={disabled}
       className={classNames({
-        'utrecht-form-label--openforms-required': isRequired,
+        'utrecht-form-label--openforms-required': isRequired && requiredFieldsWithAsterisk,
```
CharString, line 16: I expected  this fix to trigger a diff in chromatic.

CharString (changes requested):
Also check the questions on Chromatic

src/components/forms/Label.js
sergei-maertens: we didn't really have any stories with `validate.required` or the explicit config to not use asterisks for required fields, so it makes sense there's no diff.

Or do you mean markup diffs?

CharString (approved):

PR #574, opened 2023-10-23
Fix support for address autofill in editgrid

sergei-maertens:
Closes open-formulieren/open-forms#2656

Simple fix - now the local scope of data is used instead of the global bucket of data.


sergei-maertens:
Backports: 787c673249ad30322b82419058052437378a3132, dcd5c4bb1784661f50c9091df89ae06859433572 and 292e6311bbeb667fbec0d8f7112ea07f950cc994

PR #576, opened 2023-10-27
[#524] Improve input element accessibility

### LINKED ISSUE #524: Input affixes: improve accessibility
MVP approach: `aria-labelled-by` for the prefix/suffix elements for the input container.

So this could read out "Surface area" followed by "m2".

@Robbert @yolijn @Hidde feel free to add additional suggestions/input

Related: https://nl-design-system.github.io/rvo/?path=/story/componenten-form-text-input-prefix-suffix--prefix

vaszig:
Fixes #524

A new attribute has been added (`aria-labelledby`) in order to combine multiple elements (`label`, `span`) and have a proper accessible name. This attribute identifies the element (or elements) that labels the element it is applied to. This was tested with Orca screen reader in Ubuntu (`super+alt+s` to enable it).

**References**:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby
https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA16 (example 3)
https://help.gnome.org/users/orca/stable/index.html.en

**Some remarks:**

- `aria-labelledby` can be used to reference another element to define its accessible name, when an element's accessible name needs to use content from elsewhere in the DOM

- The `aria-labelledby` property value can include content from elements that aren't even visible (CSS display: none, and CSS visibility: hidden)

### DIFF
--- a/src/formio/templates/text.ejs
+++ b/src/formio/templates/text.ejs
@@ -16,6 +16,7 @@
   {{attr}}="{{ctx.input.attr[attr]}}"
   {% } %}
   id="{{ctx.instance.id}}-{{ctx.component.key}}"
+  aria-labelledby="label-{{ctx.instance.id}}-{{ctx.component.key}} {{ctx.instance.id}}-{{ctx.component.key}} suffix-{{ctx.instance.id}}-{{ctx.component.key}}"
 >{{ctx.input.content}}</{{ctx.input.type}}>
 {% if (ctx.component.showCharCount) { %}
 <span class="charcount" ref="charcount"></span>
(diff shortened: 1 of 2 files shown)


src/formio/templates/text.ejs
```diff
@@ -16,6 +16,7 @@
   {{attr}}="{{ctx.input.attr[attr]}}"
   {% } %}
   id="{{ctx.instance.id}}-{{ctx.component.key}}"
+  aria-labelledby="label-{{ctx.instance.id}}-{{ctx.component.key}} {{ctx.instance.id}}-{{ctx.component.key}} suffix-{{ctx.instance.id}}-{{ctx.component.key}}"
```
sergei-maertens, line 19: I think we should only do this when there is a suffix, this template is also rendered for a `textfield` component which doesn't have suffixes. So the suffix `<span>` will not exist, and the ID therefor will not exist, so these would be broken references. I also think if there's no suffix, it doesn't make sense to use a different label than the default.

sergei-maertens (approved):

PR #577, opened 2023-10-27
[#464] Replace previous page button with anchor

### LINKED ISSUE #464: Step previous button should become a proper link
Button is not appropriate because it doesn't do any side effects (there is no saving of the data).

We should be able to properly create an `Anchor` with a `href` to navigate back, which makes the link accessible to keyboard navigation.

SilviaAmAm:
Fixes #464

### DIFF
--- a/src/components/ButtonsToolbar/index.js
+++ b/src/components/ButtonsToolbar/index.js
@@ -1,8 +1,8 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 
-import Anchor from 'components/Anchor';
 import {OFButton} from 'components/Button';
+import Link from 'components/Link';
 import {Literal, LiteralsProvider} from 'components/Literal';
 import Loader from 'components/Loader';
 import LogoutButton from 'components/LogoutButton';
@@ -29,10 +29,10 @@ const ButtonsToolbar = ({
       <LiteralsProvider literals={literals}>
         <Toolbar modifiers={['mobile-reverse-order', 'bottom']}>
           <ToolbarList>
-            {onNavigatePrevPage && (
-              <Anchor href={previousPage} onClick={onNavigatePrevPage}>
+            {previousPage && (
+              <Link to={previousPage} onClick={onNavigatePrevPage}>
                 <Literal name="previousText" />
-              </Anchor>
+              </Link>
             )}
           </ToolbarList>
           <ToolbarList>
@@ -81,7 +81,7 @@ ButtonsToolbar.propTypes = {
   onNavigatePrevPage: PropTypes.func,
   onFormSave: PropTypes.func.isRequired,
   onLogout: PropTypes.func.isRequired,
-  previousPage: PropTypes.string.isRequired,
+  previousPage: PropTypes.string,
 };
 
 export default ButtonsToolbar;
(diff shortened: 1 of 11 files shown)


SilviaAmAm:
The link is now of a smaller font size, but I don't think that there is a token that controls the link font-size: https://github.com/nl-design-system/utrecht/blob/main/components/link/css/_mixin.scss :thinking:

src/components/ButtonsToolbar/index.js
```diff
@@ -29,9 +30,9 @@ const ButtonsToolbar = ({
         <Toolbar modifiers={['mobile-reverse-order', 'bottom']}>
           <ToolbarList>
             {onNavigatePrevPage && (
-              <UtrechtLinkButton onClick={onNavigatePrevPage} appearance="subtle-button">
+              <Anchor href={previousPage} onClick={onNavigatePrevPage}>
```
sergei-maertens, line 33: I think we can use our [`Link`](https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/Link.js#L9) component and drop the `onNavigatePrevPage` entirely, but I haven't verified that

src/components/appointments/CreateAppointment/CreateAppointment.stories.js
```diff
@@ -131,7 +131,7 @@ export const HappyFlow = {
     await step('Fill out the location and time', async () => {
       expect(await canvas.findByText('Locatie en tijd')).toBeVisible();
       await waitFor(async () => {
-        await expect(canvas.getByRole('button', {name: 'Terug naar producten'})).toBeVisible();
+        await expect(canvas.getByText('Terug naar producten')).toBeVisible();
```
sergei-maertens, line 134: I'd use `getByRole('link', {name: ...})` to lock in the accessible behavior via tests

src/scss/components/_toolbar.scss
```diff
@@ -7,7 +7,7 @@
 
 .#{prefix(toolbar)} {
   display: flex;
-  align-items: flex-start;
+  align-items: center;
```
sergei-maertens, line 10: have you checked the impact on mobile?

src/components/ButtonsToolbar/index.js
```diff
@@ -28,10 +28,10 @@ const ButtonsToolbar = ({
       <LiteralsProvider literals={literals}>
         <Toolbar modifiers={['mobile-reverse-order', 'bottom']}>
           <ToolbarList>
-            {onNavigatePrevPage && (
-              <UtrechtLinkButton onClick={onNavigatePrevPage} appearance="subtle-button">
+            {previousPage && (
+              <Link to={previousPage}>
```
sergei-maertens, line 32: I think we still want to the `onClick` so that the parent component `NAVIGATE` is still dispatched, no?

src/scss/components/_toolbar.scss
SilviaAmAm: This is the storybook from main:
Screenshot from 2023-11-01 15-49-57

This is from this branch:
Screenshot from 2023-11-01 15-50-14

(Both look bad, but unrelated to this change)

src/components/ButtonsToolbar/index.js
SilviaAmAm: Ah I think I misunderstood your previous comment then, I thought you wanted me to remove it. I will add it again.

sergei-maertens (approved):

PR #580, opened 2023-10-31
[#467] Alert refactor

### LINKED ISSUE #467: Replace openforms-alert with utrecht-alert
The utrecht component supports icon + text.

There's an accessibility discussion about interactive elements inside the text, so likely we'll break this up.

Upstream is working on a variant with actions.

SilviaAmAm:
Fixes #467 

Design tokens PR: https://github.com/open-formulieren/design-tokens/pull/39


src/components/auth/AuthenticationErrors/index.js
```diff
@@ -75,11 +75,12 @@ const AuthenticationErrors = ({parameters}) => {
     }
   }
 
-  return <ErrorMessage modifiers={['error', 'with-margin']}>{messagesToDisplay[0]}</ErrorMessage>;
+  return <ErrorMessage modifier="error">{messagesToDisplay[0]}</ErrorMessage>;
```
SilviaAmAm, line 78: Self reminder: deal with 'with-margin'

src/components/auth/AuthenticationOutage.js
```diff
@@ -15,20 +13,15 @@ export const useDetectAuthenticationOutage = () => {
...
 const AuthenticationOutage = ({loginOption}) => (
-  <div className={getBEMClassName('alert', ['error'])}>
-    <span className={getBEMClassName('alert__icon', ['wide'])}>
-      <FAIcon icon="exclamation-circle" />
-    </span>
-    <Body>
-      <FormattedMessage
-        description="Authentication outage message"
-        defaultMessage={`
+  <ErrorMessage>
```
SilviaAmAm, line 16: Self-reminder: deal with 'wide icon'

SilviaAmAm: Before:
Screenshot from 2023-11-06 11-52-16
After:
Screenshot from 2023-11-06 11-53-09
I dont think there is a need to make the icon wider? :thinking:

package.json
```diff
@@ -137,10 +138,10 @@
...
     "@testing-library/react": "^14.0.0",
     "@testing-library/user-event": "^14.4.3",
-    "@utrecht/component-library-css": "^1.0.0-alpha.526",
-    "@utrecht/component-library-react": "^1.0.0-alpha.353",
-    "@utrecht/components": "^1.0.0-alpha.473",
-    "@utrecht/design-tokens": "^1.0.0-alpha.505",
+    "@utrecht/component-library-css": "^1.0.0-alpha.604",
+    "@utrecht/component-library-react": "^1.0.0-alpha.413",
+    "@utrecht/components": "^1.0.0-alpha.565",
+    "@utrecht/design-tokens": "^1.0.0-alpha.597",
```
sergei-maertens, line 144: something I noticed working on my own branch - these dependencies must be pinned instead of allowing a range (went wrong in my earlier commit):

```suggestion
    "@utrecht/component-library-css": "1.0.0-alpha.604",
    "@utrecht/component-library-react": "1.0.0-alpha.413",
    "@utrecht/components": "1.0.0-alpha.565",
    "@utrecht/design-tokens": "1.0.0-alpha.597",
```

src/components/Errors/Errors.stories.js
```diff
@@ -0,0 +1,49 @@
+import ErrorMessage from './ErrorMessage';
+
+const ALERT_MODIFIERS = ['info', 'warning', 'error', 'ok'];
```
sergei-maertens, line 3: better to move this to (and export from) `ErrorMessage.js` so that you can use it in the proptypes:

```js
modifier: PropTypes.oneOf(ALERT_MODIFIERS),
```

src/components/auth/AuthenticationOutage.js
sergei-maertens: looks fine to me

src/components/auth/AuthenticationOutage.stories.js
```diff
@@ -0,0 +1,13 @@
+import AuthenticationOutage from './AuthenticationOutage';
+
+export default {
+  title: 'Private API / Authentication Outage',
+  render: ({label}) => <AuthenticationOutage loginOption={{label}} />,
+};
```
sergei-maertens, line 6:
```suggestion
export default {
  title: 'Private API / Authentication Outage',
  component: AuthenticationOutage,
  args: {
    loginOption: {
      label: 'DigiD,
    },
  },
};
```
why not this?

The default storybook render is to pass the `args` into the `component` as props.

src/scss/components/_alert.scss
```diff
@@ -1,44 +1,30 @@
...
-    background-color: var(--of-alert-error-bg);
-
-    .fa-icon {
-      color: var(--of-color-danger);
+// Temporary CSS override to get the colour of the icon to work
+.openforms-theme {
+  .utrecht-alert {
+    @include bem.modifier('error') {
+      .utrecht-alert__icon {
+        color: var(--utrecht-alert-icon-error-color);
```
sergei-maertens, line 8: this is not an existing design token right? I'd namespace it under our own design tokens then: `of.utrecht-alert.icon.error.color`.

src/scss/components/_authentication-errors.scss
```diff
@@ -0,0 +1,3 @@
+.openforms-authentication-errors {
+  margin: 1rem 0;
```
sergei-maertens, line 2: please don't use shorthands so that we can use styles that are writing direction independent:

```css
margin-block-end: 1rem;
margin-block-start: 1rem;
margin-inline-end: 0;
margin-inline-start: 0;
```
also - does this need to be margins? Blocks should not specify margins, but paddings to create vertical spacing so that the components can be rearranged properly.

Perhaps a better solution to create vertical spacing would also be to apply a flexbox layout to the container with appropriate `gap` values set?

sergei-maertens (changes requested):
Can you document the mapping of design tokens from old -> new?

We can also add a backwards compatibility layer for the old tokens we had

```scss
// TODO: remove in SDK 3.0
.utrecht-alert {
  @include bem.modifier('info') {
    background-color: var(--of-alert-info-bg, var(--utrecht-alert-info-background-color));
   }
  // etc...
}
```

src/scss/components/_alert.scss
SilviaAmAm: This is an existing token! But it doesn't work yet

src/scss/components/_authentication-errors.scss
SilviaAmAm: I don't really know why they were margins :thinking: I tried looking on our test environment and the one for DH, but couldn't get this alert to be rendered in order to check it. So I kept as close as possible to what we had before :S

sergei-maertens (approved):

PR #581, opened 2023-10-31
A couple more button tweaks

sergei-maertens:
I wasn't entirely satisfied yet with the changed focus ring on the button, and as there is no border-width design token for focus states, I think a CSS override in our own theme is the only viable option, but please convince me that I'm wrong!

I've also updated the design token documentation in storybook.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "@floating-ui/react": "^0.24.2",
     "@formio/protected-eval": "^1.2.1",
     "@fortawesome/fontawesome-free": "^6.1.1",
-    "@open-formulieren/design-tokens": "^0.45.0",
+    "@open-formulieren/design-tokens": "^0.45.1",
     "@sentry/react": "^6.13.2",
     "@sentry/tracing": "^6.13.2",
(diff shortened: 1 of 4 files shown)


SilviaAmAm (commented):
Currently the focus outline for the primary button and the primary button in "danger mode"/"disabled mode" is not really visible. Maybe we need to add design tokens for this?

sergei-maertens:
> Currently the focus outline for the primary button and the primary button in "danger mode"/"disabled mode" is not really visible. Maybe we need to add design tokens for this?

Hrm that's an unintended regression, I'll check locally with storybook. There are some outline design tokens available already, just need to find the right combination :grimacing:

sergei-maertens:
> Currently the focus outline for the primary button and the primary button in "danger mode"/"disabled mode" is not really visible. Maybe we need to add design tokens for this?

I see - these "buttons" are links-that-look-like-a-button and something was missed in there, I'm checking what I can do to resolve this!

SilviaAmAm (approved):

PR #582, opened 2023-11-01
Fix appointments crash

sergei-maertens:
Closes open-formulieren/open-forms#3536

This one kinda sucks because we need to pass (unvalidated) form data/state down to other form fields. Normalizing this with zod works as long as the data is valid, but crashes if the data is not valid. Allowing these empty/invalid inputs (-3 for example) in the validation schema weakens our actual validation and causes other problems downstream.

I think when we revisit this, we should not pass down the entire data structure or should use the context to obtain the data and bypass proptypes checking? I don't really see an elegant solution, so at least this patch will improve the UX.

### DIFF
--- a/src/components/appointments/CreateAppointment/CreateAppointment.spec.js
+++ b/src/components/appointments/CreateAppointment/CreateAppointment.spec.js
@@ -233,3 +233,25 @@ describe('Preselecting a product via querystring', () => {
     expect(await screen.findByText('Paspoort aanvraag')).toBeVisible();
   });
 });
+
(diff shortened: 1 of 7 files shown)


Viicos:
What will happen if an invalid value for `amount` is set (e.g. 0 or -1) when the backend will be requested?

Viicos (approved):

sergei-maertens:
Backport: 98790c2b9c1a13d4730b17dd7846b8fd5b9b476b

PR #583, opened 2023-11-01
Fix mobile button styling

sergei-maertens:
Depends on #581

### DIFF
--- a/src/scss/components/_button.scss
+++ b/src/scss/components/_button.scss
@@ -22,3 +22,14 @@
     }
   }
 }
+
+@include mobile-only {
+  .openforms-theme {
+    .utrecht-button,
+    .utrecht-button-link {
+      display: flex;
+      inline-size: 100%;
+      justify-content: center;
+    }
+  }
+}
--- a/src/scss/components/_button.scss
+++ b/src/scss/components/_button.scss
@@ -18,8 +18,7 @@
 
     // approximate browser default behaviour for focus outline
     &:focus-visible {
-      border-radius: 3px;
-      border-color: transparent;
+      --utrecht-button-border-radius: 2px;
     }
   }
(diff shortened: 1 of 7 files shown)


src/scss/components/_button.scss
```diff
@@ -15,5 +15,21 @@
...
+
+    // approximate browser default behaviour for focus outline
+    &:focus-visible {
+      --utrecht-button-border-radius: 2px;
+    }
+  }
+}
+
+@include mobile-only {
+  .openforms-theme {
```
SilviaAmAm, line 27: How come we put the .open-forms theme inside the `mobile-only` instead of adding the `mobile-only` in the rules above?

src/scss/components/_button.scss
```diff
@@ -15,5 +15,21 @@
...
+    .utrecht-button,
+    .utrecht-button-link {
+      display: flex;
+      inline-size: 100%;
```
SilviaAmAm, line 31: What is the reason to use inline-size instead of width?

SilviaAmAm (approved):

src/scss/components/_button.scss
sergei-maertens: I tend to isolate/group mobile styles from the rest of the styles so that it's a bit more clear which CSS exactly is responsible for what, but it's not exact science :grimacing:

src/scss/components/_button.scss
sergei-maertens: inline-size accounts for vertical writing directions like Japanese instead of assuming horizontal writing direction.

It's similar to the `padding-block-start` which maps to `padding-top` for horizontal writing and `padding-inline-end` which maps to `padding-left` right-to-left writing mode.

PR #585, opened 2023-11-02
🐛 Fix potential race condition on WebKit

Viicos:
Fixes https://github.com/open-formulieren/open-forms/issues/3572.

The race condition happens after `onFormIOChange` is called a first time:
- Firefox
image
- WebKit
image

`onFormIOChange` will schedule a call to `evaluateFormLogic` that will in turn dispatch `BLOCK_SUBMISSION` (thus setting `canSubmit=false`). Logic check is performed, and `LOGIC_CHECK_DONE` is about to be dispatched (which will set `canSubmit` to `true` if the logic check allows it).

For some reason, having a colored HTML element triggers `onFormIOChange` at this specific time. On Firefox, `LOGIC_CHECK_DONE` was fired in time so it had time to set `canSubmit` to `true` again (as you can see `logicChecking` is `false` when this `onFormIOChange` call happens). However, WebKit seems to be to slow and `onFormIOChange` is called _before_ `LOGIC_CHECK_DONE` is fired.

For now my fix is to return if we are still _formChecking_, but ideally a more robust solution should be found (e.g. a queue handling all the form change events?)

### DIFF
--- a/src/components/FormStep/index.js
+++ b/src/components/FormStep/index.js
@@ -414,7 +414,7 @@ const FormStep = ({
    *
    *   - `LOGIC_CHECK_INTERRUPTED` When there's no point/change in state to be expected by the logic
    *       check
-   *   - `BLOCK_SUBMISSION` When the the logic is checking and the form should be disabled for
+   *   - `BLOCK_SUBMISSION` When the logic is checking and the form should be disabled for
    *       submission.
    *   - `LOGIC_CHECK_DONE` When the logic check is done and the form should be enabled for
    *       submission.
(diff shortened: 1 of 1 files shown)


sergei-maertens (approved):
Let's try this against some of our complexer forms to see if this breaks any edge cases we haven't anticipated.

sergei-maertens:
Backports: b646d30e2bbd0d40e663fe198dbf8bdeda8869eb, 5d34139357b3d66b3f37378bccdabff3dfca7279 and 0f3462e3c963a433346fee044315d71a29fbc505

PR #586, opened 2023-11-02
🐛 Handle server side redirects

Viicos:
Fixes https://github.com/open-formulieren/open-forms/issues/3362
Linked to https://github.com/open-formulieren/open-forms/pull/3574

### DIFF
--- a/src/components/routingActions.js
+++ b/src/components/routingActions.js
@@ -1,0 +1,27 @@
+/**
+ * Get the correct redirect path for an action.
+ * @param {string} action The action to be performed.
+ * @param {Record<string, string>} actionParams The params linked to the action.
+ * @returns {{path: string, query?: URLSearchParams}} An object containing the pathname to be used,
+ * alongside with optional query parameters.
+ */
+export const getRedirectParams = (action, actionParams) => {
+  switch (action) {
+    case 'cosign':
+      return {
+        path: 'cosign/check',
+        query: new URLSearchParams(actionParams),
+      };
+    case 'afspraak-annuleren':
+      return {
+        path: 'afspraak-annuleren',
+        query: new URLSearchParams(actionParams),
+      };
+    case 'afspraak-maken':
+      return {path: 'afspraak-maken'};
+    case 'resume':
+      return {path: `stap/${actionParams.next_step}`};
+    default:
+      return {};
+  }
+};
--- a/src/sdk.js
+++ b/src/sdk.js
@@ -11,6 +11,7 @@ import {NonceProvider} from 'react-select';
 import {ConfigContext, FormContext} from 'Context';
 import {get} from 'api';
 import App, {routes as nestedRoutes} from 'components/App';
+import {getRedirectParams} from 'components/routingActions';
 import {AddFetchAuth} from 'formio/plugins';
 import {CSPNonce} from 'headers';
 import {I18NErrorBoundary, I18NManager} from 'i18n';
@@ -95,24 +96,65 @@ class OpenForm {
     CSPNonce.setValue(CSPNonceValue);
     initialiseSentry(sentryDSN, sentryEnv);
 
-    // ensure that the basename has no trailing slash (for react router)
-    let pathname = basePath || window.location.pathname;
+    let pathname = this.useHashRouting ? '' : basePath || window.location.pathname;
+
     if (pathname.endsWith('/')) {
+      // ensure that the pathname has no trailing slash (for react router)
       pathname = pathname.slice(0, pathname.length - 1);
     }
-    this.basePath = pathname;
+    this.routerBasePath = pathname;
+    this.browserBasePath = this.useHashRouting ? window.location.pathname : pathname;
+    this.makeRedirect();
     this.calculateClientBaseUrl();
+  }
+
+  makeRedirect() {
+    // Perform pre-redirect based on this action: this is decoupled from the backend
+    const query = new URLSearchParams(document.location.search);
+    const action = query.get('_of_action');
+    if (action) {
+      const actionParamsQuery = query.get('_of_action_params');
+      const actionParams = actionParamsQuery ? JSON.parse(actionParamsQuery) : {};
+      query.delete('_of_action');
+      query.delete('_of_action_params');
+
+      const {path: redirectPath, query: redirectQuery = new URLSearchParams()} = getRedirectParams(
+        action,
+        actionParams
+      );
+      const newUrl = new URL(this.browserBasePath, window.location.origin);
+      if (!this.useHashRouting) {
+        newUrl.pathname += `${!newUrl.pathname.endsWith('/') ? '/' : ''}${redirectPath}`;
+        // We first append query params from the redirect action
+        for (let [key, val] of redirectQuery.entries()) {
+          newUrl.searchParams.append(key, val);
+        }
+        // And extra unrelated query params
+        for (let [key, val] of query.entries()) {
+          newUrl.searchParams.append(key, val);
+        }
+      } else {
+        // First add extra unrelated query params, before hash (`#`)
+        for (let [key, val] of query.entries()) {
+          newUrl.searchParams.append(key, val);
+        }
+
+        // Then add our custom path as the hash part. Our query parameters are added here,
+        // but are only parsed as such by react-router, e.g. location.searchParams
+        // will not include them (as per RFC). This is why unrelated query params were added before hash.
+        // TODO use query.size once we have better browser support
+        newUrl.hash = `/${redirectPath}${[...redirectQuery].length ? '?' + redirectQuery : ''}`;
+      }
+
+      window.history.replaceState(null, '', newUrl);
+    }
   }
 
   calculateClientBaseUrl() {
     // calculate the client-side base URL, as this is recorded in backend calls for
     // submissions.
-    const clientBase = resolvePath(this.basePath).pathname; // has leading slash
-    const prefix = this.useHashRouting ? window.location.pathname : ''; // may have trailing slash
-    this.clientBaseUrl = new URL(
-      this.useHashRouting ? `${prefix}#${clientBase}` : clientBase,
-      window.location.origin
-    ).href;
+    const clientBase = resolvePath(this.browserBasePath).pathname; // has leading slash
+    this.clientBaseUrl = new URL(clientBase, window.location.origin).href;
   }
 
   async init() {
@@ -133,7 +175,7 @@ class OpenForm {
 
   render() {
     const createRouter = this.useHashRouting ? createHashRouter : createBrowserRouter;
-    const router = createRouter(routes, {basename: this.basePath});
+    const router = createRouter(routes, {basename: this.routerBasePath});
 
     // render the wrapping React component
     this.root.render(
@@ -143,7 +185,7 @@ class OpenForm {
             value={{
               baseUrl: this.baseUrl,
               clientBaseUrl: this.clientBaseUrl,
-              basePath: this.basePath,
+              basePath: this.routerBasePath,
               baseTitle: this.baseTitle,
               displayComponents: this.displayComponents,
               // XXX: deprecate and refactor usage to use useFormContext?
--- a/src/sdk.spec.js
+++ b/src/sdk.spec.js
@@ -125,17 +125,128 @@ describe('OpenForm', () => {
     expect(form.clientBaseUrl).toEqual('http://localhost/some-subpath');
   });
 
-  it('should correctly set the formUrl (hash fragment routing)', () => {
+  it("shouldn't take basepath into account (hash based routing)", () => {
     mswServer.use(...apiMocks);
-    window.history.pushState({}, 'Dummy title', '/some-server-side/path');
+    window.history.pushState({}, '', '/some-path');
     const formRoot = document.createElement('div');
     const form = new OpenForm(formRoot, {
       baseUrl: BASE_URL,
-      basePath: '/some-subpath/',
+      basePath: '/i-must-be-ignored',
       formId: '81a22589-abce-4147-a2a3-62e9a56685aa',
       useHashRouting: true,
     });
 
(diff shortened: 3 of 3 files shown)


src/index.js
```diff
@@ -6,7 +6,18 @@ import {OpenForm} from './sdk';
...
+  const searchParams = new URLSearchParams(document.location.search);
+  const redirectPath = searchParams.get('redirect_path');
+  if (redirectPath) {
+    searchParams.delete('redirect_path');
+    let basePathname = document.location.pathname;
+    if (!basePathname.endsWith('/')) basePathname = `${basePathname}/`;
+    const pathname = `${basePathname}${useHashRouting ? '#/' : ''}${redirectPath}?${searchParams}`;
+
+    window.location.replace(new URL(pathname, document.location.origin));
+  }
```
Viicos, line 19: This is dirty and should either be moved in a helper function, or (even better):
- have the router created on `OpenForm.render` set as an attribute (and created on `init` instead of `render` so that we don't have to render the form and then perform a redirect).
- Use this router attribute here to programmatically redirect to the correct path

I hope this is doable?

src/index.js
```diff
@@ -6,7 +6,18 @@ import {OpenForm} from './sdk';
 const {REACT_APP_BASE_API_URL, REACT_APP_FORM_ID, REACT_APP_USE_HASH_ROUTING} = process.env;
 
 window.onload = () => {
```
sergei-maertens, line 8: index.js is not used in the production bundle, it is only relevant for the local development. The actual entry point for production bundles is `sdk.js`, and that's where all this stuff should be handled.

I think it should also be possible to translate the real query string param into a react-router (query) param on the root route, and have the redirect handled there. We do something like that already for appointment forms: https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/App.js#L74

Viicos: Ok now it makes sense why the `CSPNonce` was hardcoded. And yes my goal here is to avoid handling things manually and let react router do the redirect

Viicos: I've spent some time experimenting and it is a real pain. With https://github.com/open-formulieren/open-forms/pull/3574, the backend is currently doing redirects to the following `Location`: `http://localhost:3000/?_action=afspraak-annuleren&time=...&submission_uuid=...`.

The issue is that as soon as the router is "initialized", hash based routing will not take into account anything before `#` (in this case it is implicit and added at the end of the URL, meaning nothing is accessible through the router API, e.g. `useLocation().search == ''`).

I can use `document.location.search` instead, and return `Navigate` after applying `_action` logic, but I end up with the following URL:

`http://localhost:3000/?_action=afspraak-annuleren&time=...&submission_uuid=...#/afspraak-annuleren?time=...&submission_uuid=...`
because hash based routing knows nothing before `#` and consider it to be the basepath.

So I could play manually with `history.pushState` but this is dirty and doesn't play well with `react-router`. Any suggestions welcomed.

sergei-maertens: What am I missing if I'm thinking of doing these steps?

1. In `src/components/App.js` get the global query params using `new URLSearchParams(window.location.search)`
2. Check if the "action" param is present. If there is, return the `<Navigate>` component to redirect
    1. Create a new (client side route) params struct by keeping all query params, dropping the `action` one
    2. Map the action to the client side route and use that for the `<Navigate>`

I'm expecting both hash based and normal URL based routes to present the same query string:

* `localhost:3000/subpath?action=resume&submission_id=123`
* `localhost:3000/subpath/?action=resume&submission_id=123#!/to-strip-of-in-the-backend`

Viicos:
Test passes, but coverage is completely off, so something is probably wrong

Viicos:
Maybe we should add an action for form steps as well: action: `form`, action_args: `<form_step>`

src/components/routingActions.js
```diff
@@ -0,0 +1,20 @@
+const validateActionArgs = func => {
+  return (...args) => {
+    if (args.some(arg => typeof arg !== 'string')) throw new Error('Invalid redirect arguments');
```
Viicos, line 3: I don't know which kind of errors we should raise if no action args are provided when required, maybe we should 404?

sergei-maertens (commented):
Found my unpublished comment :grimacing:

src/sdk.js
```diff
@@ -95,24 +96,55 @@ class OpenForm {
...
+      query.delete('_of_action');
+      query.delete('_of_action_params');
+
+      const redirectPath = getRedirectPathname(action, actionParams);
+
+      let newUrl;
+      if (!this.useHashRouting) {
+        newUrl = `${window.location.origin}${window.location.pathname}/${redirectPath}${
+          query.size ? '?' + query : ''
+        }`;
```
sergei-maertens, line 130: if `redirectPath` has query params *and* there's an uncontrolled query param like `foo=bar` from the CMS, you will end up with:

```
https://example.com/base-path/stap/step-1?submissionId=123?foo=bar
```
which is of course not the correct form. I think `getRedirectPathname` should return an object with the URL parts:

```
{
  "path": relativePath,
  "query": queryObj,
}
```
and then depending on the `useHashRouting` compose the final URL in this method.

Note that this (probably) works correctly with hash-based routing.

src/components/routingActions.js
```diff
@@ -0,0 +1,14 @@
+export const getRedirectPathname = (action, actionParams) => {
+  switch (action) {
+    case 'cosign':
+      return `cosign?submission_uuid=${actionParams.get('submission_uuid')}`;
+    case 'afspraak-annuleren':
+      return 'afspraak-annuleren';
+    case 'afspraak-maken':
+      return 'afspraak-maken';
+    case 'stap':
```
sergei-maertens, line 9: 'stap' is not a meaningful action, instead please use `resume` as the action, and the action param can be `stepSlug`

src/sdk.spec.js
```diff
@@ -125,17 +125,87 @@ describe('OpenForm', () => {
...
     const formRoot = document.createElement('div');
     const form = new OpenForm(formRoot, {
       baseUrl: BASE_URL,
-      basePath: '/some-subpath/',
       formId: '81a22589-abce-4147-a2a3-62e9a56685aa',
       useHashRouting: true,
     });
 
-    expect(form.clientBaseUrl).toEqual('http://localhost/some-server-side/path#/some-subpath');
+    expect(form.clientBaseUrl).toEqual('http://localhost/');
```
sergei-maertens, line 138: I think discarding `window.location.pathname` with hash based routing is wrong though, this should be included (but the variant is that we don't let `basePath` overwrite this value).

Imagine that a CMS has `https://example.com/forms/form1` where you start the form - the SDK will then initialize and point you to `https://example.com/forms/form1#/startpagina` and the submission will be created in the backend.

The `submission.formUrl` in this case needs to be `https://example.com/forms/form1` so that our backend can redirect you back to this (server-side) page and the SDK router can take over.

```suggestion
    expect(form.clientBaseUrl).toEqual('http://localhost/some-path');
```

src/sdk.spec.js
```diff
@@ -125,17 +125,87 @@ describe('OpenForm', () => {
...
+  ])('should handle action redirects correctly', async (initialUrl, expected) => {
+    mswServer.use(...apiMocks);
+    const formRoot = document.createElement('div');
+    window.history.pushState(null, '', initialUrl);
+    const form = new OpenForm(formRoot, {
+      baseUrl: BASE_URL,
+      basePath: '/some-subpath',
+      formId: '81a22589-abce-4147-a2a3-62e9a56685aa',
+      lang: 'nl',
+    });
```
sergei-maertens, line 167: perhaps a good idea to explicitly set `useHashRouting: false`

src/sdk.js
```diff
@@ -95,24 +96,55 @@ class OpenForm {
...
+      } else {
+        // No pathname with hash based routing
+        // TODO use query.size once we have better browser support
+        newUrl = `${window.location.origin}/${
+          [...query].length ? '?' + query : ''
+        }#/${redirectPath}`;
```
sergei-maertens, line 136: can we perhaps avoid the string concatenation and instead use the URL class?

```
const url = new URL(location);
url.searchParams.set("foo", "bar");
history.pushState({}, "", url);
```
from MDN: https://developer.mozilla.org/en-US/docs/Web/API/History/pushState#change_a_query_parameter

I was trying to figure out if we need to URL encode params for `pushState` or not and my question isn't really answered yet, but it seems that using the `new URL(...)` approach allows us to rely on the browser for the correct behaviour.

src/sdk.spec.js
```diff
@@ -125,17 +125,87 @@ describe('OpenForm', () => {
...
+    'should handle action redirects correctly (hash based routing)',
+    async (initialUrl, expected) => {
+      mswServer.use(...apiMocks);
+      const formRoot = document.createElement('div');
+      window.history.pushState(null, '', initialUrl);
+      const form = new OpenForm(formRoot, {
+        baseUrl: BASE_URL,
+        formId: '81a22589-abce-4147-a2a3-62e9a56685aa',
+        useHashRouting: true,
+        lang: 'nl',
```
sergei-maertens, line 202: maybe add `basePath: '/i-must-be-ignored'` for clarity?

src/sdk.spec.js
```diff
@@ -125,17 +125,87 @@ describe('OpenForm', () => {
...
+      'http://localhost/?unrelated_q=1#/afspraak-maken/producten', // SDK redirects to producten
+    ],
+    [
+      '/?_of_action=cosign&_of_action_params=submission_uuid%3Dabc&unrelated_q=1',
+      'http://localhost/?unrelated_q=1#/cosign?submission_uuid=abc',
+    ],
+    [
+      '/?_of_action=stap&_of_action_params=next_step%3Dstep-1&unrelated_q=1',
+      'http://localhost/?unrelated_q=1#/startpagina', // SDK redirects to start page
+    ],
```
sergei-maertens, line 191: please add a case with `/server-side/path?unrelated_q=1` -> `/server-side/path?unrelated_q=1#/startpagina` (I believe)

src/sdk.spec.js
```diff
@@ -125,17 +125,87 @@ describe('OpenForm', () => {
...
+      '/?_of_action=afspraak-maken&unrelated_q=1',
+      'http://localhost/?unrelated_q=1#/afspraak-maken/producten', // SDK redirects to producten
+    ],
+    [
+      '/?_of_action=cosign&_of_action_params=submission_uuid%3Dabc&unrelated_q=1',
+      'http://localhost/?unrelated_q=1#/cosign?submission_uuid=abc',
+    ],
+    [
+      '/?_of_action=stap&_of_action_params=next_step%3Dstep-1&unrelated_q=1',
+      'http://localhost/?unrelated_q=1#/startpagina', // SDK redirects to start page
```
sergei-maertens, line 190: that's odd, shouldn't it go to `...#/stap/step-1`?

sergei-maertens (changes requested):

src/sdk.js
Viicos: > if `redirectPath` has query params _and_ there's an uncontrolled query param like `foo=bar` from the CMS, you will end up with:

But shouldn't we keep unrelated query params? If the CMS needs `foo` to work correctly and we strip it off, it can break things, can it?

src/sdk.spec.js
Viicos: I suspect the api mocks to not cover an api call, so it goes back to the start page as it doesn't know about `step-1` but I'm not sure about that

Viicos: Edit, figured out while adding tests:
https://github.com/open-formulieren/open-forms-sdk/blob/c87f004fbaa2db868d30260aef648d0d3e5b3aeb/src/sdk.spec.js#L177-L212

The behavior is different depending on having a basepath or not with hash routing, something might be going wrong

src/sdk.js
sergei-maertens: yes we need to keep it, but now the implementation is naive in that there will only be before-hash query params or after-hash query params while both can be present at the same time - the problem is the double `?` in the resulting URL if you *don't* use hash based routing.

So you need to:

1. Server loads URL with query params, possibly with CMS-required params
2. SDK initializes
3. SDK does `pick<query, '_of_action' | '_of_action_args'>`
4. SDK processes action & action args into the (relative) path & SDK-specific query params
5. SDK builds a new URL:
    - Join basePath + relative path
    - Merge (original) params with SDK-specific query params from 4.
    - Delete `_of_action` and `_of_action_args params
6. SDK does `pushState` with newly constructed URL
    - it retains the "unrelated" query params
    - it drops the private _action routing params
    - it merges in any query params required by the SDK (like `submission_uuid` and `?productId=...` for appointments)

The current implementation fails at 5.2 - the merging is not done properly

src/sdk.spec.js
sergei-maertens: please investigate further what's causing this :grimacing:

src/sdk.js
Viicos: I think the new implementation should cover this

Viicos:
Note for later: `URLSearchParams` is treating `%2B` the wrong way: `%2B` is `+`, and `+` is being replaced by ` ` once decoded. It should stay as `+`

src/sdk.spec.js
sergei-maertens: so the resume flow relies on the `submission_id` (query) param, as that gets loaded from the backend first before the actual step is rendered. Many of these components are wrapped in a `<RequiredSubmission>` component/decorator.

Probably best to manually test the behaviour with a real submission, it doesn't make sense to redirect to a form/submission that's not in your session anyway.

Viicos: This ended up being a bit more complicated:
We have to differentiate between the react router base path and the "real" browser base path with hash based routing. In hash based routing, router base path will always be '', but we need the extra `browserBasePath` to be able to correctly set `this.clientBaseUrl`, as the CMS basepath needs to be taken into account

sergei-maertens (approved):

PR #589, opened 2023-11-07
[#36] Refactor Progress Indicator

### LINKED ISSUE #36: Clean up progress indicator implementation: JS & styling
The sass does weird things because it's based on the `card` component and on mobile spacing is different. It's also very contextual w/r to the font-size.

The combination of `position: sticky` + `flexbox` + having the expanded progress indicator _overlay_ the rest of the page is challenging.

Additional input:

* https://github.com/open-formulieren/open-forms/issues/2139
* https://github.com/open-formulieren/open-forms-sdk/pull/347#discussion_r1112925799

- [x] Refactor involved React components
- [x] Clean up SCSS & address known bugs

vaszig:
Fixes #36


src/components/ProgressIndicator/ProgressIndicatorItem.js
```diff
@@ -0,0 +1,75 @@
...
+  fixedText = null,
+}) => {
+  const getLinkModifiers = (isActive, isApplicable) => {
+    return [
+      'inherit',
+      'hover',
+      isActive ? 'active' : undefined,
+      isApplicable ? undefined : 'muted',
+    ].filter(mod => mod !== undefined);
+  };
```
sergei-maertens, line 27: I think this helper can be moved outside of the function body, it's a pure function.

src/components/ProgressIndicator/ProgressIndicatorItem.js
```diff
@@ -0,0 +1,75 @@
...
+  return (
+    <div className={getBEMClassName('progress-indicator-item')}>
+      <div className={getBEMClassName('progress-indicator-item__marker')}>
+        <CompletionMark completed={isCompleted} />
+      </div>
+      <div className={getBEMClassName('progress-indicator-item__label')}>
+        {isApplicable && canNavigateTo ? (
+          <Link
+            to={href}
+            placeholder={!canNavigateTo}
```
sergei-maertens, line 38: `placeholder` will never be `true` because when `canNavigateTo` is `false`, line 35 prevents this branch from rendering.

So this definitely requires more stories:

1. A story with `isApplicable: true, canNavigateTo: true`
2. A story with `isApplicable: true, canNavigateTo: false`
3. A story with `isApplicable: false, canNavigateTo: false`

I don't think that `isApplicable: false, canNavigateTo: true` makes sense, so that should probably `throw new Error(...)` if that's detected.

src/components/ProgressIndicator/ProgressIndicatorItem.js
```diff
@@ -0,0 +1,75 @@
...
+import CompletionMark from './CompletionMark';
+
+const ProgressIndicatorItem = ({
+  text,
+  href,
+  isActive,
+  isCompleted,
+  canNavigateTo,
+  isApplicable,
+  fixedText = null,
```
sergei-maertens, line 18: the way this is used only as `fixedText || text` makes me think that there's no need at all for this prop - instead I'd just always pass `text`?

```jsx
<>
  <ProgressIndicatorItem text={step.formDefinition} />
  <ProgressIndicatorItem text="Overview" />
  <ProgressIndicatorItem text={<FormattedMessage description="..." defaultMessage="Confirmation" />} />
<>
```
are all equivalent.

the proptype would then be `text: PropTypes.node` which is "anything that can be rendered".

I'd also consider renaming `text` to `label` or perhaps even use `children`, since `text` seems to imply only `string` is allowed.

If you go the `children` route, you can do things like:

```jsx
<>
  <ProgressIndicatorItem children={step.formDefinition} />
  <ProgressIndicatorItem>{step.formDefinition}</ProgressIndicatorItem>
</>
```
 - these both render the same.

src/components/ProgressIndicator/ProgressIndicatorItem.js
```diff
@@ -0,0 +1,75 @@
...
+              values={{
+                label: fixedText || text,
+                isApplicable: isApplicable,
+              }}
+            />
+          </Link>
+        ) : (
+          <Body component="span" modifiers={isCompleted ? ['big'] : ['big', 'muted']}>
+            {fixedText || text}
+          </Body>
```
sergei-maertens, line 58: my impression is that the `<Link placeholder />` variant replaces this - however this might be more suited for the styling updates PR and handle the CSS/design tokens for that (if needed) in that PR.

sergei-maertens (changes requested):

src/components/appointments/CreateAppointment/AppointmentProgress.js
```diff
@@ -47,26 +50,59 @@ const AppointmentProgress = ({title, currentStep}) => {
     };
   });
 
-  const ProgressIndicatorDisplayComponent =
-    config?.displayComponents?.progressIndicator ?? ProgressIndicatorDisplay;
+  // Add the fixed steps to the the original steps array
+  const stepsToRender = addFixedSteps(
```
sergei-maertens, line 54: appointments starts directly at the `producten`, there's no `startpagina` step.

I think re-using this utility is wrong - the step building is similar, but it's not the same so I'd prefer duplicating those step definitions and keeping it local. Imagine that you add another step between overview/confirmation, then the appointments would need to be updated for that too and it's not a guarantee that applies.

src/components/appointments/CreateAppointment/AppointmentProgress.js
```diff
@@ -47,26 +50,59 @@ const AppointmentProgress = ({title, currentStep}) => {
...
+  } else if (isSummary) {
+    activeStepTitle = intl.formatMessage({
+      description: 'Summary page title',
+      defaultMessage: 'Summary',
+    });
+  } else if (isConfirmation) {
+    activeStepTitle = intl.formatMessage({
+      description: 'Confirmation page title',
+      defaultMessage: 'Confirmation',
+    });
```
sergei-maertens, line 79: the translations/literals can be defined in a shared location using `react-intl`'s `defineMessage` :)

src/components/appointments/CreateAppointment/AppointmentProgress.js
```diff
@@ -47,26 +50,59 @@ const AppointmentProgress = ({title, currentStep}) => {
...
+    {title, activeStepTitle}
+  );
+
+  const ProgressIndicatorComponent =
+    config?.displayComponents?.progressIndicator ?? ProgressIndicator;
   return (
-    <ProgressIndicatorDisplayComponent
-      activeStepTitle={steps[currentStepIndex].formDefinition}
+    <ProgressIndicatorComponent
+      progressIndicatorTitle="Progress"
```
sergei-maertens, line 101: title should be translatable, now it isn't.

sergei-maertens (changes requested):

src/components/appointments/CreateAppointment/AppointmentProgress.js
vaszig: I didn't realize that. On the test env and in my local forms the start page was there (even in appointments), that's why I added it to this as well. I will remove it.

vaszig:
I have updated the code according to the suggestions-requests for change except for the styling issue, it will be added with the next commit.

sergei-maertens (approved):

PR #590, opened 2023-11-09
[#469] Add specific style for editgrid summary

### LINKED ISSUE #469: Parametrize styles for repeating groups
First: check if there are existing components we can use

- [x] Consider using paragraphs for the summaries instead of divs
- [x] Check that the buttons are properly themed (see #462)
- [x] Parametrize the CSS rules that are now using hardcoded values
- [x] Deal with spacing between the lines if the editgrid is collapsed or expanded (add token)

SilviaAmAm:
Fixes #469 

Needs: https://github.com/open-formulieren/design-tokens/pull/41

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "@floating-ui/react": "^0.24.2",
     "@formio/protected-eval": "^1.2.1",
     "@fortawesome/fontawesome-free": "^6.1.1",
-    "@open-formulieren/design-tokens": "^0.46.0",
+    "@open-formulieren/design-tokens": "^0.47.0",
     "@sentry/react": "^6.13.2",
     "@sentry/tracing": "^6.13.2",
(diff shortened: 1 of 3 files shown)


src/scss/components/_editgrid.scss
```diff
@@ -32,6 +32,16 @@
...
 
+  @include bem.element('summary') {
+    display: flex;
+    flex-direction: column;
+    gap: var(--of-editgrid-summary-gap, var(--of-editgrid-item-gap));
+
+    p {
+      margin: 0;
+    }
+  }
```
sergei-maertens, line 43: have you considered using https://nl-design-system.github.io/utrecht/storybook/?path=/docs/css_css-data-list--docs?

sergei-maertens (approved):

PR #591, opened 2023-11-09
[OF#3576] Fix empty label fields

SilviaAmAm:
Fixes https://github.com/open-formulieren/open-forms/issues/3576


src/formio/templates/editGridRow.ejs
```diff
@@ -1,5 +1,5 @@
   {% for (const key in ctx.flattenedComponents) { %}
-      {% if (ctx.isVisibleInRow(ctx.flattenedComponents[key])) { %}
+      {% if (ctx.isVisibleInRow(ctx.flattenedComponents[key]) && ctx.flattenedComponents[key].label !== "") { %}
```
sergei-maertens, line 2:
```suggestion
      {% if (ctx.isVisibleInRow(ctx.flattenedComponents[key]) && ctx.flattenedComponents[key].label) { %}
```
this would also account for `undefined` I think?

sergei-maertens (approved):

PR #594, opened 2023-11-16
:bug: Fix Leaflet styles

sergei-maertens:
The layout/container refactor (in the backend) broke the map styles in an 'interesting' way. We were loading the admin CSS for the form builder in the public frontend too, and that refactor PR realised this should not be needed and removed it. This admin CSS included the Leaflet CSS to be able to render the map in the form builder. It also provided the necessary CSS for the SDK maps, because it turns out the SDK never properly bundled the Leaflet CSS.

Cleaning up in the backend exposed this issue in the SDK, so now the import is moved to the SDK styles bundle to make it work reliably. This also means that the map component probably never worked properly with embedded forms on third party domains rather than served directly by our own backend.

The layout refactor is after Open Forms 2.4.0, so I don't expect any stable versions to be affected by this.

### DIFF
--- a/src/index.scss
+++ b/src/index.scss
@@ -1,5 +1,4 @@
 // Note: the styles here are only relevant in dev mode with `yarn start`
-@import '~leaflet/dist/leaflet';
 @import '~microscope-sass/lib/color';
 @import '~microscope-sass/lib/responsive';
 @import 'scss/settings';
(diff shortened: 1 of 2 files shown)


Viicos (approved):
I wonder if this would be possible to catch in a test/CI action

sergei-maertens:
> I wonder if this would be possible to catch in a test/CI action

Storybook snapshots should/would have caught this, but that also had a storybook-specific CSS import so that's quite unfortunate.

PR #596, opened 2023-11-20
[open-formulieren/open-forms#3611] Make time bounds validation inclusive

Viicos:
Fixes https://github.com/open-formulieren/open-forms/issues/3611

### DIFF
--- a/src/jstests/formio/components/time.spec.js
+++ b/src/jstests/formio/components/time.spec.js
@@ -2,6 +2,7 @@ import _ from 'lodash';
 import {Formio} from 'react-formio';
 
 import OpenFormsModule from 'formio/module';
+import {sleep} from 'utils';
 
 import {timeForm} from './fixtures/time';
 
@@ -28,14 +29,13 @@ describe('Time Component', () => {
     const component = form.getComponent('time');
     const changed = component.setValue(value);
     expect(changed).toBeTruthy();
-    setTimeout(() => {
-      if (valid) {
-        expect(!!component.error).toBeFalsy();
-      } else {
-        expect(!!component.error).toBeTruthy();
-        expect(component.error.message).toEqual('invalid_time');
-      }
-    }, 300);
+    await sleep(300);
+    if (valid) {
+      expect(!!component.error).toBeFalsy();
+    } else {
+      expect(!!component.error).toBeTruthy();
+      expect(component.error.message).toEqual('invalid_time');
+    }
   });
 
   it.each(['00:00:00', '23:59:59', '11:11:11'])(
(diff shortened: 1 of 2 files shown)


src/jstests/formio/components/time.spec.js
```diff
@@ -38,7 +37,7 @@ describe('Time Component', () => {
                 expect(component.error.message).toEqual('invalid_time');
               }
 
-              if (value === invalidValues[2]) {
+              if (value === invalidValues.at(-1)) {
                 done();
```
Viicos, line 41: Checking for the last array value to see if we are done is a weird pattern :thinking: I went for `at(-1)` so that you don't have to change the index value each time you add a test case.

Thinking about it, making these tests parametrized sounds better

sergei-maertens: yes, please do parametrize them

Viicos: I could go even further by parameterizing the max/min values, but it might hurt readability (but could remove a lot of duplicated code)

SilviaAmAm (approved):
Does it need to be rebased on main to fix failing tests?

Viicos:
> Does it need to be rebased on main to fix failing tests?

Weird, tests pass locally (before and after rebase). And the failing tests are from the hash based PR

sergei-maertens:
> > Does it need to be rebased on main to fix failing tests?
> 
> Weird, tests pass locally (before and after rebase). And the failing tests are from the hash based routing PR

I suspect a race condition in the redirect flow vs. test assertion time, I had similar suspicious flaky local failures too in another branch.

Viicos:
> vs. test assertion time

By "time" do you mean the tests related to the time validation? or in the literal sense

sergei-maertens:
> > vs. test assertion time
> 
> By "time" do you mean the tests related to the time validation? or in the literal sense

The point in physical time when the test assertion line executes - I suspect that the `expect(...)` call is invoked/scheduled before the redirects have fully completed and thus `window.location` is not updated yet or does not reflect the updated state yet.

sergei-maertens (changes requested):
Let's aim to fix the test race conditions/flakiness separately, however this test refactor needs to either call the `done` callback or use `await sleep(300)` for the delays and then make the assertions.

src/jstests/formio/components/time.spec.js
```diff
@@ -10,195 +9,133 @@ import {timeForm} from './fixtures/time';
...
+    const changed = component.setValue(value);
+    expect(changed).toBeTruthy();
+    setTimeout(() => {
+      if (valid) {
+        expect(!!component.error).toBeFalsy();
+      } else {
+        expect(!!component.error).toBeTruthy();
+        expect(component.error.message).toEqual('invalid_time');
+      }
+    }, 300);
```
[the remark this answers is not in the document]
Viicos: Uh, feels really weird from a UX perspective. I guess this is the current state of JS frontend tests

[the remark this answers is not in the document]
sergei-maertens: no, it has everything to do with Formio's event dispatching and debouncing. With a pure react implementation, we shouldn't need `setTimeout` anymore

src/jstests/formio/components/time.spec.js
```diff
@@ -1,204 +1,139 @@
...
+      const form = await Formio.createForm(element, formJSON);
+      form.setPristine(false);
+
+      const component = form.getComponent('time');
+      const changed = component.setValue(value);
+
+      expect(changed).toBeTruthy();
+      setTimeout(() => {
+        expect(!!component.error).toBeFalsy();
+      }, 300);
```
sergei-maertens, line 56: you have the same async/setTimeout issue here as in the other tests

sergei-maertens (changes requested):

sergei-maertens (approved):
Test flakiness was addressed in #599

sergei-maertens:
Backported in 41138f27d18b1678f588c6384b1a45f9c4b75925

PR #599, opened 2023-11-22
[OF#3626] Fix resume flow

SilviaAmAm:
Fixes https://github.com/open-formulieren/open-forms/issues/3626

### DIFF
--- a/src/components/routingActions.js
+++ b/src/components/routingActions.js
@@ -20,7 +20,10 @@ export const getRedirectParams = (action, actionParams) => {
     case 'afspraak-maken':
       return {path: 'afspraak-maken'};
     case 'resume':
-      return {path: `stap/${actionParams.next_step}`};
+      return {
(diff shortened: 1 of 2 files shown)


src/components/routingActions.js
```diff
@@ -20,7 +20,10 @@ export const getRedirectParams = (action, actionParams) => {
     case 'afspraak-maken':
       return {path: 'afspraak-maken'};
     case 'resume':
-      return {path: `stap/${actionParams.next_step}`};
+      return {
+        path: `stap/${actionParams.step_slug}`,
```
SilviaAmAm, line 24: A different parameter name was being used in the backend: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/submissions/views.py#L187

sergei-maertens (approved):

Viicos (commented):
Now I understand why I had the redirections :sweat_smile:

Issue #593, opened 2023-11-10
Refactor form routers


sergei-maertens:
These are still left over routes from react-router v5 that would benefit from being broken out into their own `routes` object structure,
properly leveraging `childRoutes`. This could simplify some progress indicator code.

Viicos:
@sergei-maertens Any place to look for in particular?

sergei-maertens:
https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/Form.js#L342

have fun :P

PR #600, opened 2023-11-22
Refactor routes for `ManageAppointment`

Viicos:
Part of #593

Spent quite some time trying to do the same with `Cosign` only to realize each route is relying on the same reducer and state and I have no idea how to handle this


sergei-maertens (approved):
For the shared state/reducer, you may want to render a skeleton/container component and make use of `<Outlet />` which renders the matching (sub)-route. The `CreateAppointment` component works like that :)

PR #604, opened 2023-12-01
:recycle: Refactor columns styling to use CSS grid

sergei-maertens:
... and design tokens.

This drops the usage of microscope-sass/bootstrap mixins/styling and aligns it with the appearance/behaviour of the new formio builder preview.

### DIFF
--- a/src/scss/components/_columns.scss
+++ b/src/scss/components/_columns.scss
@@ -1,73 +1,45 @@
 @use 'sass:math';
 
-@import '~microscope-sass/lib/grid';
-@import '~microscope-sass/lib/typography';
-
-@import '../mixins/bootstrap';
-@import '../mixins/prefix';
-
-$column-gap-size: $grid-margin-1;
-$column-half-gap-size: math.div($column-gap-size, 2);
+@import '~microscope-sass/lib/responsive';
 
 $col-sizes: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12;
 $col-sizes--mobile: 1, 2, 3, 4;
 
-.#{prefix(columns)} {
-  display: flex;
-  flex-direction: row;
+// A modern, NL DS compatible grid layout to lay out the columns.
+.openforms-columns {
+  display: grid;
+  grid-template-columns: repeat(12, 1fr);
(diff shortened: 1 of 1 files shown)


src/scss/components/_columns.scss
sergei-maertens, line 1: The class names themselves are not nice because they don't apply BEM convention. That will be done in another PR at some time, since it requires updating the templates too which may be a bit fragile.

Potentially we can then also refactor the file upload table templates to use these class names for the column layout.

CharString (changes requested):
Do I understand correctly this defines the grids for "desktop" and "mobile", and then each (layout-) component can/will define what goes where.

About changing the class names: wouldn't changing them potentially break styles currently in use?

sergei-maertens:
> Do I understand correctly this defines the grids for "desktop" and "mobile", and then each (layout-) component can/will define what goes where.

It's more like "anything bigger than mobile" and "mobile". It achieves the same visual result as the previous implementation, but instead of having to translate the `span-8` modifier to mean "apply a width of {8/12}%" we can now express directly that it should span 8 columns, because we can define our grid to consist of 12 columns. It *assumes* only a single row, but if the total number of columns exceeds 12, it will gracefully wrap onto another row so effectively multiple rows work (to catch "accidents" in the past by form designers).

The responsive styles then change this grid layout to consist of 4 columns rather than 12, and the `span-mobile-X` modifiers specify whether an element spans 1, 2, 3 or 4 columns in that grid (so 25%, 50%, 75% or 100%).

The `grid-column: span X` places the element after its sibling and tells it to span the specified number of columns, so with a markup like:

```html
<div class="columns">
  <div class="column column--span-3"></div>
  <div class="column column--span-6"></div>
</div>
```

the first element will span the grid columns 1, 2 and 3, while the second element will span 4-9 (inclusive), leaving 3 columns of free room.


> About changing the class names: wouldn't changing them potentially break styles currently in use?

No class names were changed, this PR should be 100% backwards compatible (and even then our CSS code is private API, only the design tokens is public API).


I fail to understand which changes you expect me to make :grimacing:

CharString (approved):

PR #608, opened 2023-12-07
FIx phone number validation regex bug

### LINKED ISSUE #607: Phone number field allows invalid inputs
Taiga DH 630

image

A leading dash should not be allowed.

sergei-maertens:
Fixes #607

### DIFF
--- a/src/formio/components/PhoneNumberField.js
+++ b/src/formio/components/PhoneNumberField.js
@@ -4,7 +4,7 @@ import enableValidationPlugins from '../validators/plugins';
 
 const PhoneNumber = Formio.Components.components.phoneNumber;
 
-const PHONE_NUMBER_REGEX = /^\+{0,1}[- 0-9]{0,}$/;
+const PHONE_NUMBER_REGEX = /^[+0-9][- 0-9]+$/;
 
 const PhoneNumberValidator = {
   key: 'validate.phoneNumber',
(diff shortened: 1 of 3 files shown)


src/formio/components/PhoneNumberField.js
```diff
@@ -4,7 +4,7 @@ import enableValidationPlugins from '../validators/plugins';
 
 const PhoneNumber = Formio.Components.components.phoneNumber;
 
-const PHONE_NUMBER_REGEX = /^\+{0,1}[- 0-9]{0,}$/;
+const PHONE_NUMBER_REGEX = /^[+0-9][- 0-9]+$/;
```
SilviaAmAm, line 7: Is the `- ` part here needed to accept phone numbers expressed as `06 123 123 12` or `06-123-123-12`? If yes, maybe good to add them to the tests?

sergei-maertens: yes, and you also get `06-123 456 78` etc.

I didn't add them to the tests because it was not the scope of this particular bug report, but I suppose I can add additional tests to lock in this behaviour :)

SilviaAmAm (approved):

sergei-maertens:
Backported to stable/2.0.x in 0cc53c8ba2658475c5e4f5ebe37a689b537ac866 for DH.

PR #616, opened 2023-12-18
[OF#3671] Fix max date validation when current included

vaszig:
Fixes open-formulieren/open-forms#3671


vaszig:
Tried to add a story-test for max date but same problem as for the min date. Can't find the way to make this pass automatically (it does show the error when we enter manually a wrong date).

sergei-maertens (approved):

sergei-maertens:
Backports: 67678ae84504b4cea77a0ba2169d448721e977fc

The validator code is not present in stable/1.5.x or older.

PR #623, opened 2023-12-29
Add address component

Viicos:
Related to open-formulieren/open-forms#3607

---

Frontend summary (and backend as well, currently the same) might need to be tweaked:
image

### DIFF
--- a/src/components/FormStepSummary/ComponentValueDisplay.js
+++ b/src/components/FormStepSummary/ComponentValueDisplay.js
@@ -209,6 +209,16 @@ const CoSignDisplay = ({component, value}) => {
   return <CoSignOld interactive={false} />;
 };
 
+const AddressNLDisplay = ({component, value}) => {
+  if (!value) {
+    return <EmptyDisplay />;
+  }
+
+  return `${value.postcode} ${value.houseNumber}${value.houseLetter || ''}${
+    value.houseNumberAddition || ''
+  }`;
+};
+
 const ComponentValueDisplay = ({value, component}) => {
   const {multiple = false, type} = component;
 
@@ -259,6 +269,7 @@ const TYPE_TO_COMPONENT = {
   map: MapDisplay,
   password: PasswordDisplay,
   coSign: CoSignDisplay,
+  addressNL: AddressNLDisplay,
 };
 
 export default ComponentValueDisplay;
--- a/src/components/FormStepSummary/ComponentValueDisplay.js
+++ b/src/components/FormStepSummary/ComponentValueDisplay.js
@@ -214,7 +214,7 @@ const AddressNLDisplay = ({component, value}) => {
     return <EmptyDisplay />;
   }
 
-  return `${value.postcode} ${value.houseNumber}${value.houseLetter || ''}${
+  return `${value.postcode} ${value.houseNumber}${value.houseLetter || ''} ${
     value.houseNumberAddition || ''
   }`;
 };
--- a/src/components/FormStepSummary/ComponentValueDisplay.js
+++ b/src/components/FormStepSummary/ComponentValueDisplay.js
@@ -210,7 +210,7 @@ const CoSignDisplay = ({component, value}) => {
 };
 
 const AddressNLDisplay = ({component, value}) => {
-  if (!value) {
+  if (!value || Object.values(value).every(v => v === '')) {
     return <EmptyDisplay />;
   }
 
--- a/src/components/Summary/Summary.stories.js
+++ b/src/components/Summary/Summary.stories.js
@@ -309,3 +309,31 @@ export const AddressNLSummary = {
     await canvas.findByText('1234AB 1');
   },
 };
+
+export const AddressNLSummaryEmpty = {
+  render,
+  args: {
+    summaryData: [
+      {
+        slug: 'address-nl',
+        name: 'Address NL',
+        data: [
+          {
+            name: 'Address NL',
+            value: {},
+            component: {
+              key: 'addressNL',
+              type: 'addressNL',
+              label: 'Adress NL',
+              hidden: false,
+            },
+          },
+        ],
+      },
+    ],
+  },
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+    await expect(await canvas.queryByText('1234AB 1')).toBeNull();
+  },
+};
--- a/src/components/Summary/Summary.stories.js
+++ b/src/components/Summary/Summary.stories.js
@@ -292,7 +292,7 @@ export const AddressNLSummary = {
         data: [
           {
             name: 'Address NL',
-            value: {postcode: '1234AB', houseNumber: '1'},
+            value: {postcode: '1234 AB', houseNumber: '1'},
             component: {
               key: 'addressNL',
               type: 'addressNL',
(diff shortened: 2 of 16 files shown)


src/components/FormStepSummary/ComponentValueDisplay.js
```diff
@@ -209,6 +209,16 @@ const CoSignDisplay = ({component, value}) => {
   return <CoSignOld interactive={false} />;
 };
 
+const AddressNLDisplay = ({component, value}) => {
+  if (!value) {
```
sergei-maertens, line 213: this won't catch a value like `{postcode: '', houseNumber: ''}` right?

Please add some storybook stories for this (display) component.

src/components/FormStepSummary/ComponentValueDisplay.js
```diff
@@ -209,6 +209,16 @@ const CoSignDisplay = ({component, value}) => {
...
+    return <EmptyDisplay />;
+  }
+
+  return `${value.postcode} ${value.houseNumber}${value.houseLetter || ''} ${
```
sergei-maertens, line 217: adding a story will take care of the coverage & protect us against visual regressions in the future

src/formio/components/AddressNL.mocks.js
sergei-maertens, line 1: I'm also expecting a mock/story where the validation call succeeds

src/formio/components/AddressNL.stories.js
```diff
@@ -0,0 +1,92 @@
...
+  },
+};
+
+export const Default = {
+  render: SingleFormioComponent,
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+
+    const postcodeInput = await canvas.findByRole('textbox', {name: 'Postcode'});
+    userEvent.type(postcodeInput, '1234AB');
```
sergei-maertens, line 43:
```suggestion
    await userEvent.type(postcodeInput, '1234AB');
```
`userEvent` is async, that's probably why you need the sleep calls here.

sergei-maertens (changes requested):

src/formio/components/AddressNL.stories.js
Viicos: Interaction still fails with `await` :/

src/formio/components/AddressNL.js
```diff
@@ -0,0 +1,301 @@
...
+  const {values} = useFormikContext();
+
+  useEffect(() => {
+    if (!isEqual(values, formioValues)) {
+      setFormioValues(values);
+    }
+  });
+
+  return (
+    <div className="formio-form">
```
sergei-maertens, line 246:
```suggestion
    <div className="openforms-form-field-container">
```
`.formio-form` is "reserved" for the top-level. They currently apply the same styling for the spacing, but that's not guaranteed to stay like that.

src/formio/components/AddressNL.js
```diff
@@ -0,0 +1,301 @@
...
+      <div className="openforms-columns">
+        <div className="column column--span-6 openforms-form-field-container">
```
sergei-maertens, line 248: The container class name in this case is not needed since there's only a single child element, but let's keep it for consistency sake with the generic `columns` component.

The generic columns component can have any number of nested components, so for those the spacing rule needs to be applied.

src/formio/components/AddressNL.js
```diff
@@ -0,0 +1,301 @@
...
+          <TextField
+            name="postcode"
+            label={
+              <FormattedMessage
+                description="Label for addressNL postcode input"
+                defaultMessage="Postcode"
+              />
+            }
+            placeholder="1234 AB"
```
sergei-maertens, line 257: did you create a ticket for the desired auto-formatting behaviour?

src/formio/components/AddressNL.stories.js
sergei-maertens: I'll see if I can figure out things tomorrow! Formio's debouncing does cause some issues here typically :/

src/formio/components/AddressNL.stories.js
```diff
@@ -0,0 +1,128 @@
...
+    const houseNumberInput = await canvas.findByRole('textbox', {name: 'Huis nummer'});
+    await userEvent.type(houseNumberInput, '1');
+
+    await sleep(300);
+    await userEvent.tab();
+    await sleep(300);
+
+    await waitFor(async () => {
+      expect(await canvas.queryByText('User is not a zaakgerechtigde for property.')).toBeNull();
+    });
```
sergei-maertens, line 92: Probably better to set up a form with a submit handler via `jest.fn()` and assert that is called, as it indicates that there are no validation errors. It takes only a small change in the mock error message returned and this test lost its value :grimacing: 

I think we have something like that elsewhere, I'll see if I can find it tomorrow.

src/formio/components/AddressNL.stories.js
Viicos: Yes this is what I was thinking as well

src/formio/components/AddressNL.js
```diff
@@ -0,0 +1,301 @@
...
+};
+
+const FormikAddress = ({required, formioValues, setFormioValues}) => {
+  const {values} = useFormikContext();
+
+  useEffect(() => {
+    if (!isEqual(values, formioValues)) {
+      setFormioValues(values);
+    }
+  });
```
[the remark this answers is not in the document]
Viicos: I think the check for `value.postcode` and `value.houseNumber` is useless, as `this.updateValue` is called (and should be as discussed on Slack) anyway and formio triggers the API validation.

src/components/Summary/Summary.stories.js
```diff
@@ -281,3 +281,59 @@ export const Loading = {
...
+              hidden: false,
+            },
+          },
+        ],
+      },
+    ],
+  },
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+    await canvas.findByText('1234AB 1');
```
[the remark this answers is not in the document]
sergei-maertens: this is now in main :)

src/components/Summary/Summary.stories.js
```diff
@@ -281,3 +281,59 @@ export const Loading = {
...
+              hidden: false,
+            },
+          },
+        ],
+      },
+    ],
+  },
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+    await expect(await canvas.queryByText('1234AB 1')).toBeNull();
```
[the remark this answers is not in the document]
sergei-maertens: this is now in main :)

src/formio/components/AddressNL.js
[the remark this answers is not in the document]
sergei-maertens: yes and no - it prevents some checks on empty data, but it doesn't prevent checks for data that will not be possibly considered valid by our backend.

What I mean to say is that while we are setting the formio values from the formik state, we can perhaps *also* extract the `isValid` state which is set by running the zod-schema/validators on the data, and pass that argument along with the values. So we always set the value, but only schedule a validation check if the client-side validation passes.

sergei-maertens (approved):
Let's ship it, test it and see what the stakeholder(s) think of it. I'm sure more tweaking will be needed.

PR #624, opened 2023-12-29
[#614] Use `leaflet-tools`

### LINKED ISSUE #614: Use leaflet-tools package
In favour of the copy-pasted rd.js/map.js code.

Viicos:
Fixes #614


Viicos:
Hitting a weird test failure. Might be related: https://github.com/Leaflet/Leaflet/issues/6416#issuecomment-1360369989

src/map/constants.js
sergei-maertens, line 1: I was hoping we could get rid of this file entirely and instead use only the exports from `@open-formulieren/leaflet-tools`

sergei-maertens:
> Hitting a weird test failure. Might be related: [Leaflet/Leaflet#6416 (comment)](https://github.com/Leaflet/Leaflet/issues/6416#issuecomment-1360369989)

Try updating leaflet to the latest version, this code works in the new form builder :thinking:

src/map/constants.js
Viicos: We could but there are still some constants specific to the sdk: `DEFAULT_LAT_LNG`, `DEFAULT_ZOOM`

Viicos:
> Try updating leaflet to the latest version, this code works in the new form builder 🤔

tried with no eval

SilviaAmAm (approved):

Viicos:
Regarding sergei's questions: he asked if it was possible to get rid of the `constants` module, but the current form builder does not set default values for the center. So did a bit of cleanup but still keeping the module

PR #631, opened 2024-01-12
Improve summary accessibility

sergei-maertens:
This was mentioned by @Robbert during our latest NL DS chat.

Rather than using `aria-describedby` I opted for `aria-label` to get good translatable messages - depending on the language, we can't anticipate how the form-designer provided `editStepText` properly fits into a sentence strung together from various DOM elements in a particular order.

### DIFF
--- a/src/scss/components/_summary.scss
+++ b/src/scss/components/_summary.scss
@@ -35,6 +35,11 @@
     border-block-end-color: var(--of-summary-header-border-block-end-color, var(--of-color-border));
   }
 
+  // accessible description
+  @include bem.element('link-description') {
+    display: none;
+  }
+
   // TODO: check if there's a modifier for utrecht-link with icon instead.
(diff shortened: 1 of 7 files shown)


src/scss/components/_summary.scss
```diff
@@ -35,6 +35,11 @@
     border-block-end-color: var(--of-summary-header-border-block-end-color, var(--of-color-border));
   }
 
+  // accessible description
+  @include bem.element('link-description') {
+    display: none;
```
vaszig, line 40: I thought that `display: None` is not considered as accessible. I tried this though with a screenreader as well.

vaszig (approved):

src/scss/components/_summary.scss
sergei-maertens: discussed in office - confirmed to work and I did find some article explaining this is okay because of the explicit `aria-describedby`

PR #634, opened 2024-01-12
[OF#180] Add GovMetric 

SilviaAmAm:
Fixes open-formulieren/open-forms#180

Backend: https://github.com/open-formulieren/open-forms/pull/3775


src/components/AbortionButton/AbortionButton.js
```diff
@@ -0,0 +1,71 @@
...
+
+    const govmetricUrl = buildGovMetricUrl(
+      analyticsToolsConfig.govmetricSourceId,
+      form.slug,
+      analyticsToolsConfig.govmetricSecureGuid
+    );
+
+    if (analyticsToolsConfig.enableGovmetricAnalytics) {
+      window.open(govmetricUrl);
+    }
```
Viicos, line 42:
```suggestion
    if (analyticsToolsConfig.enableGovmetricAnalytics) {
      const govmetricUrl = buildGovMetricUrl(
        analyticsToolsConfig.govmetricSourceId,
        form.slug,
        analyticsToolsConfig.govmetricSecureGuid
      );
      window.open(govmetricUrl);
    }
```

src/components/AbortionButton/AbortionButton.stories.js
```diff
@@ -0,0 +1,78 @@
...
+
+export const Authenticated = {
+  args: {
+    isAuthenticated: true,
+  },
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+
+    await waitFor(async () => {
+      const abortButton = await canvas.queryByRole('button', {name: 'Uitloggen'});
```
Viicos, line 22: Why not `getByRole`, if we know it will be there?

src/components/analytics/GovMetricSnippet.js
```diff
@@ -0,0 +1,43 @@
...
+  const form = useFormContext();
+  const intl = useIntl();
+
+  const govmetricUrl = buildGovMetricUrl(
+    analyticsToolsConfig.govmetricSourceId,
+    form.slug,
+    analyticsToolsConfig.govmetricSecureGuid
+  );
+
+  if (!analyticsToolsConfig.enableGovmetricAnalytics) return null;
```
Viicos, line 22:
```suggestion
  if (!analyticsToolsConfig.enableGovmetricAnalytics) return null;

  const govmetricUrl = buildGovMetricUrl(
    analyticsToolsConfig.govmetricSourceId,
    form.slug,
    analyticsToolsConfig.govmetricSecureGuid
  );
```

Viicos (commented):
The button seems to appear on the right in the stories:
image

Apart from that, if I understand correctly this also simplifies the current `onFormAbort` / `onLogout` logic and everything is centralized in one place?

src/components/AbortionButton/AbortionButton.stories.js
SilviaAmAm: I always hesitate when I want to test if a component is present on whether to use 'get' and then let it error immediately in case it can't find it, or use 'query' and let it error on the expect. I find the latter clearer in terms of 'what is this test testing' :thinking: what do you think?

SilviaAmAm:
> The button seems to appear on the right in the stories:

It is because the `AbortionButton` component returns a toolbar. Maybe this doesn't make much sense and it should return a button and then the component using it can add the toolbar :thinking: 

> Apart from that, if I understand correctly this also simplifies the current onFormAbort / onLogout logic and everything is centralized in one place?

Yes indeed! I tried combining those into one place. There is still the component taking care of saving the submission halfway through that also uses similar code, but I was worried of breaking it, so I didn't refactor it for now.

src/components/AbortionButton/AbortionButton.stories.js
Viicos: I don't have any preference, I find the test API not that user friendly anyway in the first place :sweat_smile:

SilviaAmAm:
Extra thought: maybe we should always have an abortion button, also  if govmetrics isn't enabled :thinking: Then we could also refactor `ExistingSubmissionOptions`

Viicos:
> Extra thought: maybe we should always have an abortion button, also if govmetrics isn't enabled 🤔 Then we could refactor `ExistingSubmissionOptions` further

I think it could be great, as a user I would expect an explicit way to abort, without having to close the browser and wonder about what actually happens when I do so

src/api-mocks/submissions.js
```diff
@@ -143,3 +143,15 @@ export const mockSubmissionPaymentStartGet = rest.post(
     return res(ctx.json({data: {method: 'get', action: 'https://example.com'}}));
   }
 );
+
+export const mockGetAnalyticsToolConfig = (overrides = {}) =>
```
sergei-maertens, line 147: this doesn't seem directly related to `submissions`, so I'd put it in a separate module `analytics-config.js` instead :)

src/components/AbortionButton/AbortionButton.mdx
```diff
@@ -0,0 +1,25 @@
...
+<Meta of={AbortionButtonStories} />
+
+# Abortion Button
+
+The abortion button can be used to abort a submission whether the user is authenticated or not.
+
+- If the user is authenticated, then the button will have the text 'Log out' to comply with Logius
+  directives that whenever a user is logged in with DigiD a 'Log out' option must be available.
+- If the analytics tool 'GovMetric' is enabled, aborting the submission will open a new window where
+  the user can give feedback about why they decided to not fill the form until the end.
```
sergei-maertens, line 14: if logout and govmetric is enabled -> probably also makes sense to send them to the feedback form

SilviaAmAm:
Follow up ticket: https://github.com/open-formulieren/open-forms/issues/3791

src/components/AbortionButton/AbortionButton.stories.js
```diff
@@ -0,0 +1,78 @@
...
+      await expect(abortButton).toBeVisible();
+    });
```
sergei-maertens, line 24: IMO the nicest is:

```js
const abortButton = await canvas.findByRole('button', {name: 'Uitloggen'});
await expect(abortButton).toBeVisible();
```
`findBy*` is really just a wrapper around `waitFor` + `getBy*`, but it makes the code flatter.

In general, you should use the async queries (`await findBy*`) when the component does stuff asynchronously, like `useEffect` or `useAsync` that may take some finite time to resolve. If the component is a "dumb" compnent and just instantly renders, typically the `getBy*` is preferred and can be done sync rather than async.

src/components/AbortionButton/AbortionButton.stories.js
```diff
@@ -0,0 +1,78 @@
...
+  },
+};
+
+export const GovMetricEnabled = {
+  args: {
+    isAuthenticated: false,
+    analyticsToolsParams: {
```
sergei-maertens, line 31: I've made some remarks in the decorator implementation about where to put these configuration params :)

src/components/App.stories.js
```diff
@@ -79,6 +79,7 @@ export default {
           {code: 'en', name: 'English'},
         ]),
         mockLanguageChoicePut,
+        mockGetAnalyticsToolConfig(),
```
sergei-maertens, line 82: looking at the other functions, this should be named `mockAnalyticsToolsConfigGet` where the HTTP method comes at the end :)

sergei-maertens (changes requested):

src/components/AbortionButton/AbortionButton.mdx
SilviaAmAm: This is indeed the case. I will reword so that it's clear

src/components/AbortionButton/AbortionButton.stories.js
```diff
@@ -0,0 +1,80 @@
...
+      govmetricSourceId: '1234',
+      govmetricSecureGuid: '',
+      enableGovmetricAnalytics: true,
+    },
+  },
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+
+    await waitFor(async () => {
```
sergei-maertens, line 40: perhaps update the remaining stories to use `await canvas.findBy*` queries?

src/components/FormStep/FormStep.stories.js
```diff
@@ -128,3 +130,47 @@ export const SuspensionDisallowed = {
...
+    analyticsToolsParams: {
+      govmetricSourceId: '1234',
+      govmetricSecureGuid: '',
+      enableGovmetricAnalytics: true,
+    },
+  },
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+
+    await waitFor(async () => {
```
sergei-maertens, line 171: would also replace this with a `findBy` query instead, to make the code flatter

src/components/PostCompletionViews/ConfirmationView.stories.js
```diff
@@ -81,3 +81,36 @@ export const WithFailedPayment = {
...
+    analyticsToolsParams: {
+      govmetricSourceId: '1234',
+      govmetricSecureGuid: '',
+      enableGovmetricAnalytics: true,
+    },
+  },
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+
+    await waitFor(async () => {
```
sergei-maertens, line 108: same as earlier - try to avoid `waitFor` :)

src/components/PostCompletionViews/ConfirmationView.stories.js
```diff
@@ -1,15 +1,15 @@
...
-import {withSubmissionPollInfo} from 'story-utils/decorators';
+import {AnalyticsToolsDecorator, withForm, withSubmissionPollInfo} from 'story-utils/decorators';
 
 import {ConfirmationViewDisplay} from './ConfirmationView';
 
 export default {
   title: 'Private API / Post completion views / Confirmation view',
   component: ConfirmationViewDisplay,
-  decorators: [withSubmissionPollInfo, withRouter],
+  decorators: [withSubmissionPollInfo, withRouter, AnalyticsToolsDecorator, withForm],
```
sergei-maertens, line 12: Can you try to keep the order of decorators similar to what is actually rendered in the app? I think the router is typically the outermost decorator.

Storybook decorators are applied from [left-to-right](https://storybook.js.org/docs/writing-stories/decorators#decorator-inheritance), so leftmost is the inner and rightmost the outer wrapper.

sergei-maertens (approved):

sergei-maertens:
Please see the comments on the Chromatic build - there's some empty blank space now that should ideally not be rendered if the abortion button is not relevant.

sergei-maertens (approved):

PR #636, opened 2024-01-18
[OF#3755] Datetime validation

SilviaAmAm:
Fixes https://github.com/open-formulieren/open-forms/issues/3755 (partially)

### DIFF
--- a/src/formio/validators/utils.js
+++ b/src/formio/validators/utils.js
@@ -8,7 +8,7 @@ export const validateBoundaries = (componentType, minBoundary, maxBoundary, valu
     return {isValid: true};
   }
 
-  const parsedValue = parseISO(value, 'yyyy-MM-dd', new Date());
+  const parsedValue = parseISO(value);
 
   let errorKeyMinValue, errorKeyMaxValue;
   if (componentType === 'date') {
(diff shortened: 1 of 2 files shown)


src/formio/validators/utils.js
```diff
@@ -0,0 +1,39 @@
...
+
+export const validateBoundaries = (componentType, minBoundary, maxBoundary, value) => {
+  const parsedMinBoundary = minBoundary ? new Date(minBoundary) : null;
+  const parsedMaxBoundary = maxBoundary ? new Date(maxBoundary) : null;
+
+  if (!parsedMinBoundary && !parsedMaxBoundary) {
+    return {isValid: true};
+  }
+
+  const parsedValue = parseISO(value, 'yyyy-MM-dd', new Date());
```
sergei-maertens, line 11: shouldn't the format be dependent on the `componentType` too?

src/jstests/formio/components/datetime.spec.js
```diff
@@ -0,0 +1,227 @@
...
+    const isValid = MinMaxDatetimeValidator.check(
+      componentInstance,
+      {},
+      '2020-01-01T10:00:00+01:00'
+    );
+
+    expect(isValid).toBeTruthy();
+
+    done();
+  });
```
sergei-maertens, line 35: this is all synchronous code, no? so we can leave out the `done` argument + callback I think?

src/formio/validators/utils.js
SilviaAmAm: This actually doesn't seem to be used :thinking: `parseIso` does not have a format argument :thinking: https://date-fns.org/v2.3.0/docs/parseISO#arguments
I will remove these extra args

sergei-maertens (approved):
Going to merge this *after* the 2.5.0 release, we can then release it as a bugfix for 2.5.1.

PR #639, opened 2024-01-26
[OF#3680] Handle multiple cosign auth methods for cosign

SilviaAmAm:
Fixes https://github.com/open-formulieren/open-forms/issues/3680


src/components/LoginOptions/LoginOptionsDisplay.js
```diff
@@ -36,7 +36,7 @@ const LoginOptionsDisplay = ({
         </>
       )}
 
-      {cosignLoginInfo && (
+      {cosignLoginOptions && cosignLoginOptions.length > 0 && (
```
vaszig, line 39: Is this the same with `cosignLoginOptions?.length > 0` ?

src/components/LoginOptions/LoginOptions.stories.js
```diff
@@ -76,18 +77,26 @@ export const Display = {
         url: '#',
       },
     ],
+    cosignLoginOptions: [
```
vaszig, line 80: Since this could be multiple options, maybe we could add one more to see the behavior?

src/components/LoginOptions/LoginOptions.stories.js
```diff
@@ -76,18 +77,26 @@ export const Display = {
...
+      {
+        identifier: 'digid',
+        label: 'DigiD',
+        logo: {
+          appearance: 'dark',
```
vaszig, line 85: I don't see `appearance`  in the `FormattedLoginOption` definition.

src/components/LoginOptions/index.js
```diff
@@ -12,8 +12,9 @@ import LoginOptionsDisplay from './LoginOptionsDisplay';
 const LoginOptions = ({form, onFormStart}) => {
   const config = useContext(ConfigContext);
 
-  const loginAsYourselfOptions = [];
-  const loginAsGemachtigdeOptions = [];
+  let loginAsYourselfOptions = [];
```
vaszig, line 15: Why do we change this to `let` ?I thought that `const` is ok with arrays even if we modify the content (the refernece is still to `loginAsYourselfOptions`)

vaszig (approved):
Approving this but I have some small remarks-questions.

src/types/FormattedLoginOption.js
```diff
@@ -8,6 +8,7 @@ const FormattedLoginOption = PropTypes.shape({
     title: PropTypes.string.isRequired,
     imageSrc: PropTypes.string.isRequired,
     href: PropTypes.string,
+    appearance: PropTypes.bool,
```
sergei-maertens, line 11:
```suggestion
    appearance: PropTypes.oneOf(LOGO_APPEARANCES).isRequired,
```
you can <kbd>CTRL</kbd> + <kbd>F</kbd> on this line and you'll find where this is defined. Probably it would be useful to create `src/types/LoginOptionLogo.js` to put this struct, and then do a search on `imageSrc` in the codebase since this shape has been duplicated in a number of places - let's re-use it as a custom proptype instead.

sergei-maertens (changes requested):

src/types/FormattedLoginOption.js
SilviaAmAm: No idea what happened in my head when I wrote `bool` :thinking: will change this!

sergei-maertens (approved):

PR #640, opened 2024-02-01
Make required asterisk theme-independant

vjawala:
In _formio_component.scss the field-required asterisk is removed because it is handled with the "utrecht-form-label--openforms-required" class. But this class was scoped on the "openforms-theme", meaning it did not work with other themes.

As the showing/hiding of the asterisk is a core functionality I think this css should be scoped on all themes. Styling can be further tweaked in the specific design system.

### DIFF
--- a/src/scss/components/_label.scss
+++ b/src/scss/components/_label.scss
@@ -13,21 +13,19 @@
   color: var(--of-color-fg-muted);
 }
 
-.openforms-theme {
-  .utrecht-form-label {
-    display: block;
-    line-height: 1.333;
-    overflow-wrap: break-word; // see open-formulieren/open-forms#576
-    @include anchor.extend-utrecht-link;
-    // TODO: remove font-weight rule on 2.0, it's only here for backwards compatibility
-    font-weight: var(--utrecht-form-label-font-weight, var(--of-label-font-weight));
+.utrecht-form-label {
+  display: block;
+  line-height: 1.333;
+  overflow-wrap: break-word; // see open-formulieren/open-forms#576
+  @include anchor.extend-utrecht-link;
+  // TODO: remove font-weight rule on 2.0, it's only here for backwards compatibility
+  font-weight: var(--utrecht-form-label-font-weight, var(--of-label-font-weight));
 
-    @include bem.modifier('openforms-required') {
-      &:after {
-        content: ' *';
-        color: var(--of-color-danger);
-        z-index: 0; // otherwise this is above the progress indicator in some browsers
-      }
+  @include bem.modifier('openforms-required') {
+    &:after {
+      content: ' *';
+      color: var(--of-color-danger);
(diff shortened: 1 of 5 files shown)


src/scss/components/_label.scss
```diff
@@ -13,21 +13,19 @@
...
 
-.openforms-theme {
-  .utrecht-form-label {
-    display: block;
-    line-height: 1.333;
-    overflow-wrap: break-word; // see open-formulieren/open-forms#576
-    @include anchor.extend-utrecht-link;
-    // TODO: remove font-weight rule on 2.0, it's only here for backwards compatibility
-    font-weight: var(--utrecht-form-label-font-weight, var(--of-label-font-weight));
+.utrecht-form-label {
```
sergei-maertens, line 16: This will also affect `utrecht-form-label` components outside of our SDK root, so if a CMS solution has other forms on the page (e.g. contact), our CSS overrides leak outside of our own "zone" when embedding forms in pages.

~~To avoid this, I think we need to gate the CSS overrides like `display`, `line-height` etc. behind design tokens and define these design token values in our own `openforms-theme`.~~

OR we scope these CSS overrides to `.utrecht-form-label.utrecht-form-label--openforms` instead - that's probably the neater solution:

```scss
.utrecht-form-label {
    @include bem.modifier('openforms') {
        display: block;
        line-height: 1.333;
        overflow-wrap: break-word;
        ...
    }
    
    @include bem.modifier('openforms-required') { ... }
}
```

sergei-maertens (changes requested):

sergei-maertens:
Thanks for the contribution!

src/scss/components/_label.scss
vjawala: Ah good one. 

When the field is required the label does not get the `.utrecht-form-label--openforms` class, only '.utrecht-form-label--openforms-required'. What if we scope it to the generic openforms container?

```scss
.openforms-container {
  .utrecht-form-label {
    ...
```

sergei-maertens: `.openforms-container` is not a class/component we have :thinking: 

> When the field is required the label does not get the `.utrecht-form-label--openforms class`

Yeah I think we should patch that extra `utrecht-form-label--openforms` classname in (to always have it):

* https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/forms/Label.js#L15
* https://github.com/open-formulieren/open-forms-sdk/blob/main/src/formio/templates/label.ejs#L3

Not sure if there are any extra places where this would apply, but I think our visual regression testing will catch them if I missed anything.

vjawala: ah yes it's the class that formio wraps around the form I think. I like approach you mentioned (instead of catching all scenario's with just css). Found a few other files where I added the "base" classname. Updated the PR!

sergei-maertens (approved):
Very nice!

PR #644, opened 2024-02-08
[OF#3835] Hide non applicable steps from Progress Indicator

vaszig:
Fixes open-formulieren/open-forms#3835


src/components/ProgressIndicator/ProgressIndicator.stories.js
```diff
@@ -24,10 +24,10 @@ export default {
       {
         to: 'first-step',
         label: 'Stap 1',
-        isCompleted: true,
-        isApplicable: true,
+        isCompleted: false,
```
vaszig, line 27: This was modified in order to show that a non-applicable step is not navigable

src/components/ProgressIndicator/ProgressIndicator.stories.js
```diff
@@ -24,10 +24,10 @@ export default {
...
+        isApplicable: false,
         isCurrent: false,
-        canNavigateTo: true,
+        canNavigateTo: false,
```
vaszig, line 30: These were modified in order to show that a non-applicable step is not navigable

src/components/Form.js
```diff
@@ -292,13 +292,28 @@ const Form = () => {
...
+  let applicableSteps = form.hideNonApplicableSteps
+    ? (updatedSubmissionSteps || steps).filter(step => step.isApplicable)
+    : [];
 
   const updatedSteps = getStepsInfo(
-    applicableSteps.length > 0 ? applicableSteps : form.steps,
+    applicableSteps.length > 0 ? applicableSteps : updatedSubmissionSteps || form.steps,
     submission,
     currentPathname
   );
```
sergei-maertens, line 319: untested code, but I think the following is logically equivalent and doesn't do data transformations:

```jsx
  // process the form/submission steps information into step data that can be passed
  // to the progress indicator.
  // If the form is marked to not display non-applicable steps at all, filter them out.
  const showNonApplicableSteps = !form.hideNonApplicableSteps;
  const updatedSteps =
    // first, process all the form steps in a format suitable for the PI
    getStepsInfo(steps, submission, currentPathname)
      // then, filter out the non-applicable steps iff they should not be displayed
      .filter(step => showNonApplicableSteps || step.isApplicable);
```

Looking at `getStepsInfo`, it already takes the submisson steps into account if a submission is provided, e.g.:

```js
isApplicable: submission ? submission.steps[index].isApplicable : step.isApplicable ?? true,
```
the `.map` expression in there (and almost all of the code in the SDK!) is built on the premise that `formsSteps[index]` and `submission.steps[index]` point to the same step. i.e. `formSteps.length === submission.steps.length`.

So, that should already take the dynamic nature of `isApplicable` from the logic rules into account, which makes it possible to do the `.filter` call based on `isApplicable`. And instead of looking only at `step.isApplicable`, we also take `form.hideNonApplicableSteps` into account *inside* the filter, so that if we display them with the `n/a` suffix, they are not actually filtered out.

sergei-maertens (approved):

sergei-maertens:
Backport to 2.1.x: 20d75ea240eca43ff85b4a038078dad3053500ee (following feedback from Laurens)

PR #646, opened 2024-02-19
:arrow_up: Bump storybook to 7.0.27

sergei-maertens:
Had to just entirely nuke `yarn.lock` to get this to work since apparently multi-version dependency resolution is not properly fixed/implemented in yarn :unamused: 

Some dependencies are now pinned more specifically in `package.json` because they were causing visual regressions.


sergei-maertens:
@nikkiysendoorn1 can you check out this branch and try if it resolves your storybook issues for you?

nikkiysendoorn1:
@sergei-maertens It is not working. I get the same error as yesterday :(

PR #647, opened 2024-02-19
[OF#3855] Catch API errors on start submission

SilviaAmAm:
Fixes open-formulieren/open-forms#3855


sergei-maertens (commented):
Would an error boundary not be better? :thinking:

PR #649, opened 2024-02-20
Move session expiry page to own route

### LINKED ISSUE #645: SDK: session expiry should be its own route
Now we have clunky state management and container routes that perform side effects (such as creating a new submission).

Instead, when the user session expires, we should (client-side) redirect to a dedicated page showing the message, and the side effect should be that any session data and state (`sessionStorage` and the like) are cleared.

nikkiysendoorn1:
Closes #645

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -4,6 +4,7 @@ import {Navigate, Outlet, useMatch} from 'react-router-dom';
 import {Cosign} from 'components/CoSign';
 import ErrorBoundary from 'components/Errors/ErrorBoundary';
 import Form from 'components/Form';
+import SessionExpired from 'components/Sessions/SessionExpired';
 import {
   CreateAppointment,
   appointmentRoutes,
@@ -27,6 +28,14 @@ export const routes = [
     path: 'cosign/*',
     element: <Cosign />,
   },
+  {
+    path: 'sessie-verlopen',
+    element: (
+      <ErrorBoundary useCard>
+        <SessionExpired />
+      </ErrorBoundary>
+    ),
+  },
   // All the rest goes to the formio-based form flow
   {
     path: '*',
@@ -46,12 +55,13 @@ const App = () => {
   const query = useQuery();
   const appointmentMatch = useMatch('afspraak-maken/*');
   const appointmentCancelMatch = useMatch('afspraak-annuleren/*');
+  const isSessionExpiryMatch = useMatch('sessie-verlopen');
 
   // register localized error messages in the default zod error map
   useZodErrorMap();
 
   const isAppointment = form.appointmentOptions?.isAppointment ?? false;
-  if (isAppointment && !appointmentMatch && !appointmentCancelMatch) {
+  if (isAppointment && !appointmentMatch && !appointmentCancelMatch && !isSessionExpiryMatch) {
     return (
       <Navigate
         replace
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -14,7 +14,7 @@ import Loader from 'components/Loader';
 import {ConfirmationView, StartPaymentView} from 'components/PostCompletionViews';
 import ProgressIndicator from 'components/ProgressIndicator';
 import RequireSubmission from 'components/RequireSubmission';
-import {RequireSession} from 'components/Sessions';
+import {SessionTrackerModal} from 'components/Sessions';
 import SubmissionSummary from 'components/Summary';
 import {START_FORM_QUERY_PARAM} from 'components/constants';
 import {findNextApplicableStep} from 'components/utils';
@@ -122,7 +122,6 @@ const Form = () => {
   const [state, dispatch] = useImmerReducer(reducer, initialStateFromProps);
 
   const onSubmissionLoaded = (submission, next = '') => {
-    if (sessionExpired) return;
     dispatch({
       type: 'SUBMISSION_LOADED',
       payload: submission,
@@ -140,11 +139,7 @@ const Form = () => {
     onSubmissionLoaded
   );
 
-  const [sessionExpired, expiryDate, resetSession] = useSessionTimeout(() => {
-    removeSubmissionId();
-    dispatch({type: 'DESTROY_SUBMISSION'});
-    flagNoActiveSubmission();
-  });
+  const [, expiryDate, resetSession] = useSessionTimeout();
 
   const {value: analyticsToolsConfigInfo, loading: loadingAnalyticsConfig} = useAsync(async () => {
     return await get(`${config.baseUrl}analytics/analytics_tools_config_info`);
@@ -362,7 +357,7 @@ const Form = () => {
         path="overzicht"
         element={
           <ErrorBoundary useCard>
-            <RequireSession expired={sessionExpired} expiryDate={expiryDate}>
+            <SessionTrackerModal expiryDate={expiryDate}>
               <RequireSubmission
                 submission={state.submission}
                 form={form}
@@ -372,7 +367,7 @@ const Form = () => {
                 onClearProcessingErrors={() => dispatch({type: 'CLEAR_PROCESSING_ERROR'})}
                 onDestroySession={onDestroySession}
               />
-            </RequireSession>
+            </SessionTrackerModal>
           </ErrorBoundary>
         }
       />
@@ -411,7 +406,7 @@ const Form = () => {
         path="stap/:step"
         element={
           <ErrorBoundary useCard>
-            <RequireSession expired={sessionExpired} expiryDate={expiryDate}>
+            <SessionTrackerModal expiryDate={expiryDate}>
               <RequireSubmission
                 form={form}
                 submission={state.submission}
@@ -422,7 +417,7 @@ const Form = () => {
                 component={FormStep}
                 onDestroySession={onDestroySession}
               />
-            </RequireSession>
+            </SessionTrackerModal>
           </ErrorBoundary>
         }
       />
--- a/src/components/Sessions/SessionExpired.js
+++ b/src/components/Sessions/SessionExpired.js
@@ -1,0 +1,49 @@
+import React, {useContext, useEffect} from 'react';
+import {FormattedMessage} from 'react-intl';
+
+import {ConfigContext} from 'Context';
+import Card from 'components/Card';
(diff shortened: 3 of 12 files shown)


src/components/App.js
```diff
@@ -12,6 +13,8 @@ import useFormContext from 'hooks/useFormContext';
 import useQuery from 'hooks/useQuery';
 import useZodErrorMap from 'hooks/useZodErrorMap';
 
+import SessionExpired from './SessionExpired';
```
sergei-maertens, line 16:
```suggestion
import SessionExpired from 'components/SessionExpired';
```

src/components/Form.js
```diff
@@ -14,7 +14,7 @@ import Loader from 'components/Loader';
 import {ConfirmationView, StartPaymentView} from 'components/PostCompletionViews';
 import ProgressIndicator from 'components/ProgressIndicator';
 import RequireSubmission from 'components/RequireSubmission';
-import {RequireSession} from 'components/Sessions';
+import {SessionModalTracker} from 'components/Sessions';
```
sergei-maertens, line 17:
```suggestion
import {SessionModalTracker} from 'components/Sessions';
```
maybe `SessionTrackerModal`?

src/components/SessionExpired.js
```diff
@@ -0,0 +1,46 @@
+import React, {useContext} from 'react';
+import {FormattedMessage} from 'react-intl';
+
+import Card from 'components/Card';
+import ErrorMessage from 'components/Errors/ErrorMessage';
+import Link from 'components/Link';
+
+import {ConfigContext} from '../Context';
+import {flagNoActiveSubmission} from '../data/submissions';
+import useSessionTimeout from '../hooks/useSessionTimeout';
```
sergei-maertens, line 10:
```suggestion
import useSessionTimeout from 'hooks/useSessionTimeout';
```

src/components/SessionExpired.js
```diff
@@ -0,0 +1,46 @@
...
+
+const SessionExpired = () => {
+  const config = useContext(ConfigContext);
+  const [expired] = useSessionTimeout();
+
+  if (!config.debug && !expired) {
+    throw new Error("The session isn't expired");
```
sergei-maertens, line 17: there should be some kind of marker you can use to ignore coverage for this line

src/components/SessionExpired.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  flagNoActiveSubmission();
+  window.localStorage.clear();
+  window.sessionStorage.clear();
```
sergei-maertens, line 22: This is imperative code, while React is based on a declarative model. For side effects, you should wrap this code in a `useEffect` hook to perform the side effects in an imperative way.

src/components/SessionExpired.js
```diff
@@ -0,0 +1,46 @@
...
+
+  return (
+    <Card
+      title={
+        <FormattedMessage
+          description="Session expired card title"
+          defaultMessage="Your session has expired"
+        />
```
sergei-maertens, line 30: To make sure all translations are known, please also run `yarn makemessages` and `yarn compilemessages`

If these exact messages already existed, this will be a no-op so it's safe to always run.

src/hooks/useSessionTimeout.js
```diff
@@ -28,7 +30,10 @@ const useSessionTimeout = onTimeout => {
 
     // don't schedule the expiry-setter (which leads to state updates and re-renders)
     // if the session is already expired.
-    if (expired) return;
+    if (expired) {
+      navigate('/sessie-verlopen');
```
sergei-maertens, line 34: self-reminder: test if this works with appointments too

src/components/Summary/CosignSummary.js
```diff
@@ -73,7 +73,7 @@ const CosignSummary = ({
     await destroySession();
   };
 
-  const [sessionExpired, expiryDate] = useSessionTimeout(async () => {
+  const [expiryDate] = useSessionTimeout(async () => {
```
sergei-maertens, line 76: the signature of `useSessionTimeout` is now not used consistently - the first argument in the return tuple is the `expired` boolean.

In `src/components/Form.js` you do account for this with `[, arg2, arg3]`, but this one has a type issue.

If the `expired` value is not relevant (anymore), I prefer updating the `useSessionTimeout` return value to return a two-tuple rather than the three-tuple it has now.

src/components/appointments/CreateAppointment/CreateAppointment.js
```diff
@@ -66,11 +66,15 @@ const CreateAppointment = () => {
               {isLoading ? (
                 <Loader modifiers={['centered']} />
               ) : (
-                <RequireSession expired={sessionExpired} expiryDate={expiryDate} onNavigate={reset}>
+                <SessionModalTracker
+                  expired={sessionExpired}
+                  expiryDate={expiryDate}
+                  onNavigate={reset}
+                >
```
sergei-maertens, line 73:
```suggestion
                <SessionModalTracker expiryDate={expiryDate}>
```
the `expired` and `onNavigate` props were obsoleted - you'll likely be able to do some more cleanups with unused references after this change :)

src/components/SessionExpired.js
```diff
@@ -0,0 +1,46 @@
...
+import ErrorMessage from 'components/Errors/ErrorMessage';
+import Link from 'components/Link';
+
+import {ConfigContext} from '../Context';
+import {flagNoActiveSubmission} from '../data/submissions';
+import useSessionTimeout from '../hooks/useSessionTimeout';
+
+const SessionExpired = () => {
+  const config = useContext(ConfigContext);
+  const [expired] = useSessionTimeout();
```
sergei-maertens, line 14: we should check if the `reset` callback returned by `useSessionTimeout` should be called here or not - likely you can test this by using the appointment flow:

1. start an appointment
2. let the session timeout
3. verify that we get redirected to the session expiry
4. click the "return" link going back to the start page
5. it should be possible to start a new appointment submission (and complete this)

there may be some internal state about expiry date/timers that now prevents this.

sergei-maertens (changes requested):
There should be tests for these changes too - we use testing-library (with user-event) that should be relatively straight forward to test the behaviour of this new component.

Checks for when this component is rendered:
* local storage is cleared
* ...?

src/components/Sessions/index.js
```diff
@@ -1 +1,4 @@
-export {default as RequireSession} from './RequireSession';
+import SessionExpired from './SessionExpired.js';
```
sergei-maertens, line 1:
```suggestion
import SessionExpired from './SessionExpired';
```

src/components/Sessions/SessionExpired.stories.js
```diff
@@ -0,0 +1,13 @@
...
+
+export default {
+  title: 'Views / Session Expired',
+  component: SessionExpired,
+  decorators: [withRouter],
+};
+
+export const Default = {
+  name: 'Session Expired',
+};
```
sergei-maertens, line 13: I did some experimenting locally to get this story to work in non-debug mode too (which is happening in Chromatic, and the following code gets us there:

```suggestion
import {useEffect} from 'react';
import {withRouter} from 'storybook-addon-react-router-v6';

import {sessionExpiresAt} from 'api';
import useSessionTimeout from 'hooks/useSessionTimeout';
import {ConfigDecorator} from 'story-utils/decorators';

import SessionExpired from './SessionExpired';

const render = args => {
  const [expired] = useSessionTimeout();
  useEffect(() => {
    const now = new Date();
    sessionExpiresAt.setValue({expiry: now});
    return () => {
      sessionExpiresAt.setValue({expiry: null});
    };
  });
  if (!expired) {
    return 'Waiting for expiry...';
  }

  return <SessionExpired {...args} />;
};

export default {
  title: 'Views / Session Expired',
  component: SessionExpired,
  decorators: [ConfigDecorator, withRouter],
  parameters: {
    config: {
      debug: false, // force false in local dev-mode
    },
  },
};

export const Default = {
  name: 'Session Expired',
  render,
};

```

However, this will require some changes to `useSessionTimeout` too which I'll leave for you to figure out :wink:

sergei-maertens (changes requested):

src/components/Sessions/SessionExpired.js
```diff
@@ -0,0 +1,50 @@
+import React, {useContext, useEffect} from 'react';
+import {FormattedMessage} from 'react-intl';
+
+import Card from 'components/Card';
+import ErrorMessage from 'components/Errors/ErrorMessage';
+import Link from 'components/Link';
+import useSessionTimeout from 'hooks/useSessionTimeout';
+
+import {ConfigContext} from '../../Context';
+import {flagNoActiveSubmission} from '../../data/submissions';
```
sergei-maertens, line 10:
```suggestion
import {ConfigContext} from 'Context';
import {flagNoActiveSubmission} from 'data/submissions';
```
we try to avoid "path hell" :smile:

src/components/Sessions/SessionExpired.js
```diff
@@ -0,0 +1,50 @@
...
+
+const SessionExpired = () => {
+  const config = useContext(ConfigContext);
+  const [expired] = useSessionTimeout(null, true);
+
+  // codecov:ignore:start
```
sergei-maertens, line 16: that didn't seem to work :thinking: 

You can try:

```suggestion
/* istanbul ignore next */
```

sergei-maertens (changes requested):
As discussed - please also investigate the visual regression :)

src/components/appointments/CreateAppointment/CreateAppointment.spec.js
```diff
@@ -110,9 +113,7 @@ describe('Create appointment session expiration', () => {
 
     // now finally let the session timeout in 1s
     act(() => updateSessionExpiry(1));
-    await waitFor(async () => {
-      await screen.findByText('Your session has expired');
-    });
+    await screen.findByText('Your session has expired', undefined, {timeout: 2000});
```
sergei-maertens, line 116: failed because of the internal timeout being too low, so the session-will-expire-over-1000ms-and-then-render-a-different-page cycle exceeded that. Giving 1s for the rendering etc. is sufficient here.

PR #652, opened 2024-02-22
[OF#3791] Abort button for anonymous user

SilviaAmAm:
Fixes https://github.com/open-formulieren/open-forms/issues/3791

### DIFF
--- a/src/components/FormStart/index.js
+++ b/src/components/FormStart/index.js
@@ -41,7 +41,9 @@ const FormStartMessage = ({form}) => {
  * This is shown when the form is initially loaded and provides the explicit user
  * action to start the form, or present the login button (DigiD, eHerkenning...)
  */
-const FormStart = ({form, hasActiveSubmission, onFormStart, onDestroySession}) => {
+const FormStart = ({form, submission, onFormStart, onDestroySession}) => {
+  const hasActiveSubmission = !!submission;
+  const isAuthenticated = hasActiveSubmission && submission.isAuthenticated;
   const doStart = useStartSubmission();
   const outagePluginId = useDetectAuthenticationOutage();
   const authErrors = useDetectAuthErrorMessages();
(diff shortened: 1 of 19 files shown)


src/components/FormStart/index.js
```diff
@@ -124,7 +130,7 @@ const FormStart = ({form, hasActiveSubmission, onFormStart, onDestroySession}) =
 
 FormStart.propTypes = {
   form: Types.Form.isRequired,
-  hasActiveSubmission: PropTypes.bool.isRequired,
+  submission: Types.Submission,
```
vaszig, line 133: Why do we use both `Types` and `propTypes`?What is the difference?

vaszig (approved):
Looks good to me, only a couple of small remarks.
Only the story for the `AbortButton` has still the old "Abortion" title.

src/components/FormStart/index.js
SilviaAmAm: `Types` is defined by us to reuse types like submission/form: https://github.com/open-formulieren/open-forms-sdk/blob/main/src/types/index.js

src/components/FormStart/tests.spec.js
```diff
@@ -112,12 +106,12 @@ it('Form start page does not show login buttons if an active submission is prese
         form={testForm}
         onFormStart={onFormStart}
         onDestroySession={onDestroySession}
-        hasActiveSubmission={true}
+        submission={{id: 'tralala', isAuthenticated: false}}
```
sergei-maertens, line 109: This will cause prop type check warnings, please use the mock facilities to build a submission object that conforms to the prop types: https://github.com/open-formulieren/open-forms-sdk/blob/main/src/api-mocks/index.js#L3

`submission={buildSubmission({isAuthenticated: false})}` should be sufficient

src/components/Summary/test.spec.js
```diff
@@ -112,3 +113,27 @@ it('Summary does not display logout button if loginRequired is false', () => {
...
+          onConfirm={onConfirm}
+          onDestroySession={onDestroySession}
+          onClearProcessingErrors={() => {}}
+        />
+      </Wrap>
+    );
+  });
+
+  expect(container.textContent).toContain('Afbreken');
+});
```
sergei-maertens, line 139: can you use testing-library instead for new tests?

src/components/appointments/SubmitRow.js
```diff
@@ -20,10 +21,11 @@ const SubmitRow = ({canSubmit, nextText, previousText = '', navigateBackTo = ''}
       isLastStep={false}
       isCheckingLogic={false}
       loginRequired={false}
+      hideAbortButton={true}
```
sergei-maertens, line 24:
```suggestion
      hideAbortButton
```
same effect, but is more idiomatic React

sergei-maertens (approved):
Seems fine to me :+1: Couple small remarks

sergei-maertens (approved):
let's ship it!

PR #653, opened 2024-02-23
Correctly render `content` component in summary

Viicos:
Fixes https://github.com/open-formulieren/open-forms/issues/3845

Not sure this is the right solution

### DIFF
--- a/src/components/FormStepSummary/ComponentValueDisplay.js
+++ b/src/components/FormStepSummary/ComponentValueDisplay.js
@@ -219,6 +219,10 @@ const AddressNLDisplay = ({component, value}) => {
   }`;
 };
 
+const ContentDisplay = ({component, value}) => {
+  return <span dangerouslySetInnerHTML={{__html: component.html}} />;
+};
+
 const ComponentValueDisplay = ({value, component}) => {
   const {multiple = false, type} = component;
 
@@ -270,6 +274,7 @@ const TYPE_TO_COMPONENT = {
   password: PasswordDisplay,
   coSign: CoSignDisplay,
   addressNL: AddressNLDisplay,
+  content: ContentDisplay,
 };
 
 export default ComponentValueDisplay;
--- a/src/components/Summary/Summary.stories.js
+++ b/src/components/Summary/Summary.stories.js
@@ -32,6 +32,7 @@ export default {
               type: 'content',
               label: 'Content',
               key: 'content',
+              html: '<i>In this section you can enter your personal details.</i>',
             },
           },
           {


src/components/FormStepSummary/ComponentValueDisplay.js
```diff
@@ -219,6 +219,10 @@ const AddressNLDisplay = ({component, value}) => {
   }`;
 };
 
+const ContentDisplay = ({component, value}) => {
+  return <span dangerouslySetInnerHTML={{__html: component.html}} />;
```
Viicos, line 223: or `__html: value`, but this would require a change in the backend: `Submission.render_summary_page` currently uses `Renderer(..., as_html=False)`, hence `value` is just the string value without html and styling

SilviaAmAm: I think it makes sense to use `component.html` :thinking:

SilviaAmAm (approved):

SilviaAmAm:
Euhh not sure what I did wrong when reviewing chromatic, but maybe update the stories to show how the content component with style are rendered?

src/components/FormStepSummary/ComponentValueDisplay.js
sergei-maertens: One thing that I'm not sure about is if that HTML is CSP post-processed to convert inline styles (like text color) to CSP-nonced `<style>` tags. I can't find any reference in the `submissions` code that would hint this is the case.

I'm okay with leaving this out, as most of the formatting will still work since they're applied through `<strong>` tags etc. rather than inline styles.

sergei-maertens (changes requested):
Additional/updated stories seems appropriate indeed :)

src/components/FormStepSummary/ComponentValueDisplay.js
Viicos: It's a bit weird, the summary endpoint returns the `html` without `style` tags, e.g.:
`<p>Some <i>pretty </i><span style="color:#4c99e6;"><i>component</i></span></p>`

But the submission step endpoint uses `style` elements:

`<style nonce="RqgbALvp8D5b3+8NuhfuKg==">#nonce-a2ef23e2ee96a66e5e16b1c2c7aec411-139719785812288 {    color:#4c99e6;} </style><p>Some <i>pretty </i><span id="nonce-a2ef23e2ee96a66e5e16b1c2c7aec411-139719785812288"><i>component</i></span></p>`

sergei-maertens (approved):

sergei-maertens:
Backports:

* stable/2.1.x: 24b59bf3871a47f7ed11e0140bd1f59d2723d6e8
* stable/2.0.x: 674eb66386c94e5f937a1b3caa79d33fd8433a82
* stable/1.5.x: 841c54050e06db784765cde5dd1a55d3b7ee10bc

PR #664, opened 2024-03-18
:sparkles: [#483] add description to selectboxes and radio option

### LINKED ISSUE #483: Use description per option in selectboxes and radio components
DH Taiga issue 540

This should happen both in the native components and the formio templates.

image

nikkiysendoorn1:
Closes #483

Renamed the branch, see here the previous PR: https://github.com/open-formulieren/open-forms-sdk/pull/658

### DIFF
--- a/src/formio/components/Checkbox.scss
+++ b/src/formio/components/Checkbox.scss
@@ -27,7 +27,8 @@
         }
       }
 
-      .utrecht-form-field__label {
+      .utrecht-form-field__label,
+      .utrecht-form-description--openforms {
         background-color: var(--_of-utrecht-form-field-checkbox-background-color);
         --_of-checkbox-label-shift: calc(
           var(--of-utrecht-form-label-checkbox-padding-block-start) + var(--utrecht-checkbox-size)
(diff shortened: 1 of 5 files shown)


nikkiysendoorn1:
Renamed the branch, see here the previous PR: https://github.com/open-formulieren/open-forms-sdk/pull/658

src/formio/components/Checkbox.scss
```diff
@@ -48,10 +49,24 @@
...
     }
   }
 }
 
+.openforms-theme {
+  .utrecht-custom-checkbox--openforms:not(:checked) {
+    background-color: var(--utrecht-checkbox-background-color);
+  }
+}
+
```
sergei-maertens, line 69: this appears to duplicate the lines below, no?

src/formio/templates/radio.ejs
```diff
@@ -13,6 +13,7 @@
     ref="input"
     value="{{item.value}}"
     id="{{ctx.id}}{{ctx.row}}-{{item.value}}"
+    aria-describedby="{{ctx.id}}{{ctx.row}}-{{item.value}}-description"
```
sergei-maertens, line 16: if there's no item description, there's an aria-describedby pointer to an ID that doesn't exist. That doesn't seem valid - probably best to only include the attribute if there is a description:

```suggestion
    {% if (item.description) %}aria-describedby="{{ctx.id}}{{ctx.row}}-{{item.value}}-description"{% } %}
```

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #666, opened 2024-03-20
Release 2.2.0

sergei-maertens:
* Checked translations
* Updated changelog
* Bumped version

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,7 +1,7 @@
 [bumpversion]
 commit = False
 tag = False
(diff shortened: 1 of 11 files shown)


vaszig (approved):
It would be nicer to have the styling fix in another PR but I guess it's because of the beta version, right?

sergei-maertens:
> It would be nicer to have the styling fix in another PR but I guess it's because of the beta version, right?

Yeah it was supposed to have been fixed already, but I noticed while writing the changelog that we missed a case. It's a separate commit at least :)

PR #671, opened 2024-04-15
[open-formulieren/open-forms#4158] Fix date validation and translations

Viicos:
Fixes open-formulieren/open-forms#4158

This fixes two things:
- ~~A `date` component with `minDate` set would not validate properly (e.g. with `minDate=2024-01-01`, `2024-01-01` would fail to validate). This is fixed by switching from `new Date(minBoundary)` to `parseISO(minBoundary)`.~~ done in https://github.com/open-formulieren/open-forms-sdk/pull/673
- Translations were not being picked up properly. I simplified the code of `validateBoundaries` to only return one possible error key, although we might want to follow what is done in the `time` component at some point (see [`MinMaxTimeValidator.js`](https://github.com/open-formulieren/open-forms-sdk/blob/e8fa867b105426b37e0cba16c6829b7f43dee99e/src/formio/validators/MinMaxTimeValidator.js); `invalid_time` is a valid error key, and it will fallback to the provided message of `invalid_time` if `minTime` doesn't have a custom message).

Lots of test failing so I'll wait for a review before tackling these.


src/formio/validators/utils.js
```diff
@@ -1,39 +1,29 @@
...
   const parsedValue = parseISO(value);
 
-  let errorKeyMinValue, errorKeyMaxValue;
-  if (componentType === 'date') {
-    errorKeyMinValue = 'minDate';
-    errorKeyMaxValue = 'maxDate';
-  } else if (componentType === 'datetime') {
-    errorKeyMinValue = 'minDatetime';
-    errorKeyMaxValue = 'maxDatetime';
-  }
```
sergei-maertens, line 20: In the context of my own ticket I'll need this fix, but I'm also trying to track down which error keys were actually supposed to be used.

Viicos: For the `datetime` component, this part of the code would use `minDatetime` and `maxDatetime` as error keys, but we saw in the actual `MinMaxDatetimeValidator` that we expect `minDate` and `maxDate` to be used. This is the main reason the custom translations were not being picked up

sergei-maertens: Okay, I had to do some serious digging, but I found:

* for the `date` component, we exposed the `minDate` and `maxDate` error keys
* for the `datetime` component, we never deliberately exposed those error keys - it came accidentally with the new form builder. The `datetime` component is the built in component from formio, and uses the `minDate` and `maxDate` error keys
* the time component is out of scope here now.

The `minDatetime` and `maxDatetime` error keys never seem to have been a thing. I found commit 99d984275e27283aa460bf84670b89c1b421320e where the datetime min/max validation behaviour was added, which refers to ticket open-formulieren/open-forms#3755 The associated backend PR never specified those errors keys either, so I think it's okay to just stick to `minDate` and `maxDate` for both component types.

src/formio/validators/utils.js
sergei-maertens, line 1: I think this is functionally equivalent? I'd like to see @SilviaAmAm 's opininion though.

src/formio/validators/minMaxDatetimeValidator.js
```diff
@@ -9,9 +9,9 @@ const MinMaxDatetimeValidator = {
...
 
-    const errorKeys =
-      component?.openForms?.validationErrorContext?.minMaxDatetimeValidatorErrorKeys;
-    const errorMessage = errorKeys ? errorKeys[0] : 'invalidDatetime';
+    const errorKey = component?.openForms?.validationErrorContext?.minMaxDatetimeValidatorErrorKey;
+    const componentErrorMessages = component.component.errors;
+    const errorMessage = componentErrorMessages[errorKey] || errorKey;
 
     return component.t(errorMessage, {
       minDatetime: minDatetime,
```
sergei-maertens, line 17: same pattern as above applies, however you need to change the context from `minDatetime` to `minDate` etc. If you use the same message keys, then the message templates will be the same and expect the same context.

src/formio/validators/minMaxDatetimeValidator.js
sergei-maertens, line 1: I don't see a reason why there should still be two validator "classes", they functionally and programmatically do the same at this point, no? So, shouldn't we just merge these and re-use it for both date and datetime components?

src/formio/validators/minMaxDateValidator.js
```diff
@@ -8,8 +8,9 @@ const MinMaxDateValidator = {
     const minDate = new Date(component.component.minDate);
     const maxDate = new Date(component.component.maxDate);
 
-    const errorKeys = component?.openForms?.validationErrorContext?.minMaxDateValidatorErrorKeys;
-    const errorMessage = errorKeys ? errorKeys[0] : 'invalidDate';
```
sergei-maertens, line 12: formio actually documents the `invalid_date` error key, so there may be a camel case conversion thing in play.

sergei-maertens:
The `new Date` -> `parseIso` patch needs to be backported

src/formio/validators/minMaxDatetimeValidator.js
Viicos: Yes I think they can be merged

src/formio/validators/minMaxDateValidator.js
```diff
@@ -8,8 +8,9 @@ const MinMaxDateValidator = {
...
+    const errorKey = component?.openForms?.validationErrorContext?.minMaxDateValidatorErrorKey;
+    const componentErrorMessages = component.component.errors;
+    const errorMessage = componentErrorMessages[errorKey] || errorKey;
```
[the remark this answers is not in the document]
sergei-maertens: > Or perhaps what is wanted is that if there's no minDate custom error message, we should try to look if a custom error message is set for invalidDate?

Bingo :)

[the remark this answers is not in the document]
Viicos: Ok so this behavior will be similar to the time component iirc

src/formio/validators/minMaxDateValidator.js
sergei-maertens: I checked this with our custom `invalid_time` thing and I don't see any usage of `invalidDate` (camelcase) in Formio, so looks like it's just a bug and we should fall back to `invalid_date` like Formio itself does

src/formio/validators/minMaxDateAndDatetimeValidator.js
```diff
@@ -0,0 +1,48 @@
...
+const createValidator = (key, invalidKey) => {
+  return {
+    key,
+    message(component) {
+      const minDate = new Date(component.component.minDate);
+      const maxDate = new Date(component.component.maxDate);
+
+      const errorKey =
+        component.openForms.validationErrorContext.minMaxDateAndDatetimeValidatorErrorKey;
+      const componentErrorMessages = component.component.errors;
```
sergei-maertens, line 14: I'm confused, we want to show the most appropriate error message, so `errorKey` is for example `minDate` - why can't we use `component.errorMessage(errorKey)`?

e.g. with `minDate`, this will:

1. Look up the key in the custom error messages and use it if it's present
2. Look up the global message/translation if it's not present in the custom keys

If we don't have a specific error key, the error key itself should become `invalid_date` or `invalid_datetime`, which can in turn again be looked up with the custom error messages or the global translation.

We always want to use the most specific error keys, so `minDate` global translation is favored over `invalid_date` custom error message. I think I misunderstood your [comment](https://github.com/open-formulieren/open-forms-sdk/pull/671#discussion_r1567504978)

Viicos: The issue with `component.errorMessage(errorKey)` is that it will fallback to `errorKey`:

```js
  errorMessage(type) {
    return (this.component.errors && this.component.errors[type]) ? this.component.errors[type] :  type;
  }
```

So instead of doing this currently:

```js
const errorMessage = componentErrorMessages[errorKey] || componentErrorMessages[invalidKey] || errorKey;
```

I'd have to do:

```js
const errorMessage = component.errorMessage(errorKey) === errorKey ? component.errorMessage(invalidKey) === invalidKey ? errorKey : invalidKey;
```
 (not sure if this is a 100% correct), but it is hard to understand

sergei-maertens: but the fallback to `errorKey` is fine, because then it'll use the translation from https://github.com/open-formulieren/open-forms/blob/master/src/openforms/js/lang/formio/nl.json

Viicos: > We always want to use the most specific error keys, so `minDate` global translation is favored over `invalid_date` custom error message. I think I misunderstood your [comment](https://github.com/open-formulieren/open-forms-sdk/pull/671#discussion_r1567504978)

Reading this again, I'm not sure I understand what would be the use case for `invalid_date` in this case? If we fallback to the `minDate` global translation, `invalid_date` will never be used?

It might be that you want the `invalid_date` error key to be used when the input is not a valid date (e.g. `2020-ab-cd`), but you can't input non-numeric characters in the SDK, and when giving an "incomplete" date (e.g. `2020-__-__`), formio will autocomplete the rest.

sergei-maertens: `invalid_date` is used when the input cannot be parsed as a date: https://github.com/formio/formio.js/blob/bebc2ad73cad138a6de0a8247df47f0085a314cc/src/validator/Validator.js#L526

Viicos: :+1: 
The time component would also have to follow this path as well (in another PR)

Viicos: I'll note that trying to set the date to an invalid input via backend logic (and I assume via other ways too) crashes the SDK, before [this validator](https://github.com/formio/formio.js/blob/bebc2ad73cad138a6de0a8247df47f0085a314cc/src/validator/Validator.js#L526) is being called, so the `invalid_date` key is currently not being used

sergei-maertens: hrm, concerning, but let's just stick to the formio patterns here. There were some flags with allowing/not allowing invalid dates so this is probably relevant for old components still.

src/formio/validators/minMaxDateAndDatetimeValidator.js
```diff
@@ -0,0 +1,43 @@
+import set from 'lodash/set';
+
+import {validateBoundaries} from './utils';
+
+const createValidator = key => {
+  return {
+    key,
+    message(component) {
+      const minDate = new Date(component.component.minDate);
+      const maxDate = new Date(component.component.maxDate);
```
sergei-maertens, line 10: shouldn't this also use `parseISO` for dates, otherwise we end up with the same timezone issues again?

Viicos: These are only used for the error message formatting, but I think switching to `parseISO` is fine. I also see we are using different attributes from the `check` call, I'll look into that.

sergei-maertens: I'd rather have `parseISO` be applied consistently, otherwise next time someone will be tempted to use `new Date` on a date (rather than datetime) string again and then we have the same timezone bug again.

Viicos:
This is how it looks with the format options:

image

sergei-maertens (approved):

PR #672, opened 2024-04-15
[open-formulieren/open-forms#4130] Do not show main url when asking for payment

Viicos:
Fixes open-formulieren/open-forms#4130


SilviaAmAm:
I think that these stories still need updating: src/components/PostCompletionViews/viewsWithPolling.stories.js ?

SilviaAmAm:
> I think that these stories still need updating: src/components/PostCompletionViews/viewsWithPolling.stories.js ?

No, nothing needs to be done for these: these are the confirmation views for *after* payment

PR #674, opened 2024-04-18
[open-formulieren/open-forms#4201] Fix crash with hidden map component

Viicos:
Fixes open-formulieren/open-forms#4201


src/formio/components/Map.js
```diff
@@ -78,8 +78,10 @@ export default class Map extends Field {
       mapContainer: 'single',
     });
     return super.attach(element).then(() => {
-      this.reactRoot = createRoot(this.refs.mapContainer);
-      this.renderReact();
+      if (this.refs.mapContainer) {
+        this.reactRoot = createRoot(this.refs.mapContainer);
+        this.renderReact();
```
sergei-maertens, line 83: I don't understand how this line is not covered when there's a test?

sergei-maertens:
I don't understand how this test replicates the crash? The issue specifies that the crash does not occur when adding frontend logic, so it doesn't seem logical to me to add a story of a map component with frontend logic.

> Adding FE logic to the map component solves the issue. back-end logic does not.

Can't we just write a simple jest regression test (not a story) here for a map component that's hidden and then validate the fix against that?

src/formio/components/Map.js
Viicos: Yes I don't understand as well, and adding a `console.log` in this code path confirms it is covered by the story by running it in the browser

Viicos:
> Can't we just write a simple jest regression test (not a story) here for a map component that's hidden and then validate the fix against that?

Hum I think what I did first was to define the story without any `play` defined, or at least a `play` without any body, but from what I remember, while it did crash without my fix, it still passed the story test. This is why I used the conditional logic to have a meaningful story test. However, testing now even with an empty `play` body seems to be caught by SB. Anyway, this is confusing and I can't remember exactly what I did previously

src/formio/components/Map.js
sergei-maertens: hrm, I suspect JS race conditions and instrumentation not playing along well then

src/jstests/formio/components/map.spec.js
```diff
@@ -0,0 +1,38 @@
...
+
+    const element = document.createElement('div');
+
+    Formio.createForm(element, formJSON)
+      .then(form => {
+        form.setPristine(false);
+        done();
+      })
+      .catch(done);
+  });
```
sergei-maertens, line 37: IMO this is nicer:

```suggestion
  // smoke/regression test for map components, which used to crash
  test('Hidden map component', async () => {
    let formJSON = _.cloneDeep(mapForm);

    const element = document.createElement('div');

    const form = await Formio.createForm(element, formJSON);
    form.setPristine(false);
  });
```

sergei-maertens (approved):

PR #676, opened 2024-04-19
Fix imports after SBv8 upgrade

Viicos:


sergei-maertens (approved):
Did you use a tool for this or go through them by hand?

Viicos:
> Did you use a tool for this or go through them by hand?
By hand, there's a chance I missed some. But I think setting up eslint or https://biomejs.dev/ would help and would pick up the missing ones if we ever use it

sergei-maertens:
no I was asking because if there's a tool that I didn't find, I want it :) there's an open issue in the import sort plugin for prettier to apply the grouping, but it hasn't been implemented yet.

I've gone through the same in the form-builder repo and it wasn't much fun

PR #679, opened 2024-04-22
[open-formulieren/open-forms#4207] Fix styling overflow for select dropdown

Viicos:
Fixes open-formulieren/open-forms#4207


sergei-maertens:
where is the storybook story for this? I want visual regression tests for this.

sergei-maertens (changes requested):
On mobile (and probably on desktop too in some cases) the options themselves are overflowing, we should probably fix that too.

Viicos:
Ok finally this should be ok

PR #680, opened 2024-04-25
 [open-formulieren/open-forms#4222] Fix max files upload limit 

Viicos:
Partly fixes open-formulieren/open-forms#4222

I can't make the test to pass, it seems to not update the data when only calling `.upload` :/

### DIFF
--- a/src/formio/components/FileField.js
+++ b/src/formio/components/FileField.js
@@ -184,12 +184,7 @@ class FileField extends Formio.Components.components.file {
 
   upload(files) {
     if (this.component.multiple && this.component.maxNumberOfFiles) {
-      // this.data.file contains files that may have already been uploaded, while the argument 'files' gives the
-      // files that are being uploaded in this call
-      if (
-        files.length > this.component.maxNumberOfFiles ||
-        this.data?.file?.length >= this.component.maxNumberOfFiles
-      ) {
+      if (this.dataValue.length + files.length > this.component.maxNumberOfFiles) {
         const messages = [
           {
             message: this.t(
--- a/src/jstests/formio/components/file.spec.js
+++ b/src/jstests/formio/components/file.spec.js
@@ -66,4 +66,30 @@ describe('Multiple File Component', () => {
       })
       .catch(done);
   });
+
+  // GH-4222
+  test('Uploading 1 file then 2 files gives an error', done => {
+    let formJSON = _.cloneDeep(maxNFilesForm);
+
+    const element = document.createElement('div');
+
+    Formio.createForm(element, formJSON)
+      .then(form => {
+        form.setPristine(false);
+        const component = form.getComponent('multipleFiles');
+        component.dataValue.push({name: 'File 1', size: '50'});
+
+        component.upload([
+          {name: 'File 2', size: '50'},
+          {name: 'File 3', size: '50'},
+        ]);
+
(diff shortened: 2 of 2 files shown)


src/formio/components/FileField.js
```diff
@@ -184,12 +185,8 @@ class FileField extends Formio.Components.components.file {
 
   upload(files) {
     if (this.component.multiple && this.component.maxNumberOfFiles) {
-      // this.data.file contains files that may have already been uploaded, while the argument 'files' gives the
```
Viicos, line 187: `data` is a mapping of component keys -> values, so `file` was only relevant if it was the actual key of the file component

src/formio/components/FileField.js
```diff
@@ -184,12 +185,8 @@ class FileField extends Formio.Components.components.file {
...
-      // files that are being uploaded in this call
-      if (
-        files.length > this.component.maxNumberOfFiles ||
-        this.data?.file?.length >= this.component.maxNumberOfFiles
-      ) {
+      const existingFiles = get(this.data, this.component.key);
```
Viicos, line 188: There might be a formio util function to do this without using the lodash implementation detail, but couldn't find one

sergei-maertens: `this.dataValue` is a getter that looks up the value using lodash in the base `Component` class, so I'd use that.

sergei-maertens (changes requested):
I don't know about the tests and which/how many promises are involved, but you may need to convert the tests to use `async`/`await` for more reliable results.

src/formio/components/FileField.js
```diff
@@ -184,12 +184,7 @@ class FileField extends Formio.Components.components.file {
 
   upload(files) {
     if (this.component.multiple && this.component.maxNumberOfFiles) {
-      // this.data.file contains files that may have already been uploaded, while the argument 'files' gives the
-      // files that are being uploaded in this call
-      if (
-        files.length > this.component.maxNumberOfFiles ||
-        this.data?.file?.length >= this.component.maxNumberOfFiles
-      ) {
+      if (this.component.dataValue + files.length > this.component.maxNumberOfFiles) {
```
sergei-maertens, line 187: `this.dataValue`, not `this.component.dataValue`

sergei-maertens (changes requested):

src/jstests/formio/components/file.spec.js
```diff
@@ -66,4 +66,30 @@ describe('Multiple File Component', () => {
...
+    Formio.createForm(element, formJSON)
+      .then(form => {
+        form.setPristine(false);
+        const component = form.getComponent('multipleFiles');
+        component.upload([{name: 'File 1', size: '50'}]);
+
+        component.upload([
+          {name: 'File 2', size: '50'},
+          {name: 'File 3', size: '50'},
+        ]);
```
sergei-maertens, line 85: internally this is called `redraw()`, which returns a promise but nowhere do they actually await this promise. So I'm expecting some race conditions here - what happens if you put some `sleep` calls before and after the `upload` calls?

Viicos: I can make it work by pushing an initial file in `component.dataValue`

sergei-maertens (approved):
I give up with Formio, this is good enough for me

PR #688, opened 2024-05-13
:sparkles: [open-formulieren/open-forms#4199] Explicitly start anonymous submission

stevenbal:
Issue: https://github.com/open-formulieren/open-forms/issues/4199

Related PR in backend: https://github.com/open-formulieren/open-forms/pull/4271

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -164,7 +164,7 @@ const Form = () => {
    * @param  {Event} event The DOM event, could be a button click or a custom event.
    * @return {Void}
    */
-  const onFormStart = async event => {
+  const onFormStart = async (event, anonymous = false) => {
     event && event.preventDefault();
 
     // required to get rid of the error message saying the session is expired - once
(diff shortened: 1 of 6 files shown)


src/components/Form.js
```diff
@@ -178,7 +178,14 @@ const Form = () => {
 
     let submission;
     try {
-      submission = await createSubmission(config.baseUrl, form, config.clientBaseUrl);
+      const anonymous = !!event;
```
stevenbal, line 181: I'm not sure if there is another way for me to figure out what login option the user selected at this point

sergei-maertens: yeah, in `src/components/LoginOptions/index.js` on line 66 you can provide an explicit argument to indicate that it's an anonymous login:

```jsx
  const containerProps = form.loginRequired
    ? {}
    : {
        onSubmit: e => {
          e.preventDefault();
          onFormStart(e, {anonymous: true});
        },
      };
```
so then just update the `onFormStart` signature to take `(event, {anonymous = false}) {...}`, that makes it explicit :)

src/components/Form.js
```diff
@@ -178,7 +178,13 @@ const Form = () => {
 
     let submission;
     try {
-      submission = await createSubmission(config.baseUrl, form, config.clientBaseUrl);
+      submission = await createSubmission(
```
stevenbal, line 181: @sergei-maertens I don't think this line was covered before either, and I'm not sure where to start with writing a test for this `onFormStart` :sweat_smile:

sergei-maertens (approved):
It was painful, but I managed a test for the non-anonymous flow too.

PR #690, opened 2024-06-10
[OF#3993] Derive address functionality in addressNL component

vaszig:
Partly fixes open-formulieren/open-forms#3993

### DIFF
--- a/src/components/FormStepSummary/ComponentValueDisplay.js
+++ b/src/components/FormStepSummary/ComponentValueDisplay.js
@@ -214,9 +214,14 @@ const AddressNLDisplay = ({component, value}) => {
     return <EmptyDisplay />;
   }
 
-  return `${value.postcode} ${value.houseNumber}${value.houseLetter || ''} ${
-    value.houseNumberAddition || ''
-  }`;
+  return (
+    <>
+      {value.streetName} {value.houseNumber}
+      {value.houseLetter} {value.houseNumberAddition}
+      <br />
+      {value.postcode} {value.city}
+    </>
+  );
 };
 
 const ContentDisplay = ({component, value}) => {
--- a/src/components/FormStepSummary/ComponentValueDisplay.js
+++ b/src/components/FormStepSummary/ComponentValueDisplay.js
@@ -214,13 +214,26 @@ const AddressNLDisplay = ({component, value}) => {
     return <EmptyDisplay />;
   }
 
+  if (component.deriveAddress) {
+    return (
+      <address>
+        {value.streetName} {value.houseNumber}
+        {value.houseLetter} {value.houseNumberAddition}
+        {value.postcode && value.city && (
+          <>
+            <br />
+            {value.postcode} {value.city}
+          </>
+        )}
+      </address>
+    );
+  }
+
   return (
-    <>
-      {value.streetName} {value.houseNumber}
+    <address>
+      {value.postcode} {value.houseNumber}
       {value.houseLetter} {value.houseNumberAddition}
-      <br />
-      {value.postcode} {value.city}
-    </>
+    </address>
   );
 };
 
--- a/src/formio/components/AddressNL.js
+++ b/src/formio/components/AddressNL.js
@@ -3,7 +3,7 @@
  */
 import {Formik, useFormikContext} from 'formik';
 import debounce from 'lodash/debounce';
-import React, {useEffect} from 'react';
+import React, {useContext, useEffect} from 'react';
 import {createRoot} from 'react-dom/client';
 import {Formio} from 'react-formio';
 import {FormattedMessage, IntlProvider, defineMessages, useIntl} from 'react-intl';
@@ -11,6 +11,7 @@ import {z} from 'zod';
 import {toFormikValidationSchema} from 'zod-formik-adapter';
 
 import {ConfigContext} from 'Context';
+import {get} from 'api';
 import {TextField} from 'components/forms';
 import enableValidationPlugins from 'formio/validators/plugins';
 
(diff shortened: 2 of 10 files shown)


vaszig:
This should be merged after the PRs in types and formio builder have been approved and merged
https://github.com/open-formulieren/types/pull/48
https://github.com/open-formulieren/formio-builder/pull/155

src/components/FormStepSummary/ComponentValueDisplay.js
```diff
@@ -216,7 +216,7 @@ const AddressNLDisplay = ({component, value}) => {
 
   return `${value.postcode} ${value.houseNumber}${value.houseLetter || ''} ${
     value.houseNumberAddition || ''
-  }`;
+  } ${value.city || ''} ${value.streetName || ''}`;
 };
```
sergei-maertens, line 220: can we sort out the order to be more natural here?

Modified preview in storybook should look like this:

image


So that also makes that this should probably be: `$streetName, $houseNumber$houseLetter $houseNumberAddition, $postcode $city`

src/formio/components/AddressNL.js
```diff
@@ -319,23 +324,70 @@ const FormikAddress = ({required, setFormioValues}) => {
...
+                name="streetName"
+                label={
+                  <FormattedMessage
+                    description="Label for addressNL streetName read only result"
+                    defaultMessage="Street name"
+                  />
+                }
+                disabled
+              />
+            </div>
```
sergei-maertens, line 352: I think it's more natural to swap the order here - first street name, then city

src/formio/components/AddressNL.js
```diff
@@ -349,3 +401,42 @@ const PostCodeField = ({required}) => {
...
 };
+
+const HouseNumberField = ({required, allowDeriveAddress}) => {
+  const {values, setFieldValue, getFieldProps} = useFormikContext();
+  const {onBlur: onBlurFormik} = getFieldProps('houseNumber');
+  const {baseUrl} = useContext(ConfigContext);
+  const postcode = values['postcode'];
+  const houseNumber = values['houseNumber'];
+
+  const fetchStreetNameAndCity = async () => {
```
sergei-maertens, line 412: this code is now duplicated, you can extract it into a single function used by both fields, and you might even just pass it in as a prop from the parent component:

```jsx
const {values, setFieldValue} = useFormikContext();

const autofillAddress = () => {
    if (!allowDeriveAddress) return;
    if (!values.postcode) || (!values.houseNumber)} return;
    
    const data = await get(`${baseUrl}geo/address-autocomplete`, params);
    setFieldValue('city', data['city']);
    setFieldValue('streetName', data['streetName']);
    setFieldValue('secretStreetCity', data['secretStreetCity']);
};

// ...

return (
  <>
    <PostCodeField required={required} autoFillAddress={autofillAddress} />
  </>
);
```

and then your onBlur becomes:

```jsx
const onBlur = (event) => {
  onBlurFormik(event);
  autoFillAddress();
};
```

src/formio/templates/editGridRow.ejs
```diff
@@ -3,9 +3,19 @@
...
+          </dd>
+        {% } else { %}
+          {% for (const prop in ctx.row[key]) { %}
+            {% if (ctx.row[key][prop] && prop !== 'secretStreetCity') { %}
+              <dd class="utrecht-data-list__item-value utrecht-data-list__item-value--html-dd">
+                {{ ctx.row[key][prop] }}
+              </dd>
+            {% } %}
+          {% } %}
+        {% } %}
```
sergei-maertens, line 18: Objects in JS do not formally have any ordering - and we'd like to have a stable presentation order like this:

image

I think the `AddressNL` formio component should instead implement the `getView` method: https://github.com/formio/formio.js/blob/4.13.x/src/components/_classes/component/Component.js#L1315 and it looks like as long as you implement `https://github.com/formio/formio.js/blob/4.13.x/src/components/_classes/component/Component.js#L1315` that it should be pretty straight forward. Then you won't need to touch this template at all.

sergei-maertens (changes requested):

src/components/FormStepSummary/ComponentValueDisplay.js
```diff
@@ -214,9 +214,14 @@ const AddressNLDisplay = ({component, value}) => {
     return <EmptyDisplay />;
   }
 
-  return `${value.postcode} ${value.houseNumber}${value.houseLetter || ''} ${
-    value.houseNumberAddition || ''
-  }`;
+  return (
+    <>
+      {value.streetName} {value.houseNumber}
+      {value.houseLetter} {value.houseNumberAddition}
```
sergei-maertens, line 220: I'd consider using the `<address>` tag as it has semantic value

src/components/FormStepSummary/ComponentValueDisplay.js
```diff
@@ -214,9 +214,14 @@ const AddressNLDisplay = ({component, value}) => {
...
+      <br />
+      {value.postcode} {value.city}
```
sergei-maertens, line 222: these should probably only be displayed if `deriveAddress` is enabeld, and if their values are empty maybe don't display them?

src/formio/components/AddressNL.js
```diff
@@ -162,6 +171,19 @@ export default class AddressNL extends Field {
...
+
+  // Called in the representation of an edit grid
+  getValueAsString(value, options) {
+    if (!value || !Object.keys(value).length) return '';
+    const {postcode, houseNumber, houseLetter, houseNumberAddition, streetName, city} = value;
+    return `
+        <address>
+          ${postcode} ${houseNumber}${houseLetter} ${houseNumberAddition}
+          <br>
+          ${streetName || ''} ${city || ''}
```
sergei-maertens, line 183: I don't think the linebreak should be there if there is no streetname and no city

sergei-maertens (changes requested):

PR #692, opened 2024-06-20
[OF#4382] Missing client validation for email in the save and resume modal

vaszig:
Fixes open-formulieren/open-forms#4382

### DIFF
--- a/src/components/modals/FormStepSaveModal.js
+++ b/src/components/modals/FormStepSaveModal.js
@@ -29,8 +29,8 @@ const emailValidationSchema = z.object({
 
 const FIELD_LABELS = defineMessages({
   email: {
-    description: "Label for 'email' field",
-    defaultMessage: 'Email address',
+    description: 'Form save modal email field label',
+    defaultMessage: 'Your email address',
   },
 });
 
@@ -74,19 +74,23 @@ const FormStepSaveModal = ({
   const [{errorMessage, isSaving}, dispatch] = useImmerReducer(reducer, initialState);
 
   useEffect(() => {
-    if (!isOpen) {
+    if (!isOpen && errorMessage) {
       dispatch({
         type: 'REMOVE_ERROR_MESSAGE',
       });
     }
-  });
+  }, [dispatch, isOpen, errorMessage]);
 
   const errorMap = (issue, ctx) => {
+    const fieldLabelDefinition = FIELD_LABELS[issue.path.join('.')];
+    if (!fieldLabelDefinition) {
+      return {message: ctx.defaultError}; // use global schema as fallback
+    }
+    const fieldLabel = intl.formatMessage(fieldLabelDefinition);
+
     switch (issue.code) {
       case z.ZodIssueCode.invalid_type: {
         if (issue.received === z.ZodParsedType.undefined) {
-          const fieldName = issue.path.join('.');
-          const fieldLabel = intl.formatMessage(FIELD_LABELS[fieldName]);
           const message = intl.formatMessage(
             {
               description: 'Required field error message',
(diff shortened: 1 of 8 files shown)


sergei-maertens (changes requested):
While the fix is indeed adding validation, I'm missing some bits in this PR:

* Story updates/interaction tests -> `src/components/modals/FormStepSaveModal.stories.js`
* Translations - now the validation errors are displayed always in English. See https://open-formulieren.github.io/open-forms-sdk/?path=/docs/developers-documentation--docs#handling-validation-errors-with-zod for instructions

vaszig:
> While the fix is indeed adding validation, I'm missing some bits in this PR:
> 
>     * Story updates/interaction tests -> `src/components/modals/FormStepSaveModal.stories.js`
> 
>     * Translations - now the validation errors are displayed always in English. See https://open-formulieren.github.io/open-forms-sdk/?path=/docs/developers-documentation--docs#handling-validation-errors-with-zod for instructions

Yes, sorry I forgot to do that as a draft. I wanted to add a story but again I would have forgotten the translations.

src/components/modals/FormStepSaveModal.js
```diff
@@ -21,6 +23,17 @@ const initialState = {
   isSaving: false,
 };
 
+const emailValidationSchema = z.object({
+  email: z.string().email(),
+});
+
+const FIELD_LABELS = defineMessages({
```
sergei-maertens, line 30: typically you do a construct like this so that the field label and the label mentioned in the error message are the same, so you should also update https://github.com/open-formulieren/open-forms-sdk/pull/692/files#diff-505e57d8cf61bc5d12e7aca0402abc93e96f76e868e894629d8cb3b5fe7c4f7cR186 to read:

```jsx
            <EmailField
              name="email"
              isRequired
              label={intl.formatMessage(FIELD_LABELS.email)}
              description={
                <FormattedMessage
                  description="Form save modal email field help text"
                  defaultMessage="The email address where you will receive the resume link."
                />
              }
            />
```

src/components/modals/FormStepSaveModal.js
```diff
@@ -50,11 +63,52 @@ const FormStepSaveModal = ({
...
+        const fieldLabel = intl.formatMessage(FIELD_LABELS[fieldName]);
+        const message = intl.formatMessage(
+          {
+            description: 'Invalid field value message',
+            defaultMessage: 'Invalid {field}.',
+          },
+          {
+            field: fieldLabel,
+          }
+        );
```
sergei-maertens, line 102: I think you can specifically find problems with email fields here to give a better error message. See https://github.com/open-formulieren/formio-builder/blob/main/src/registry/email/edit-validation.ts#L20 for example

src/components/modals/FormStepSaveModal.stories.js
```diff
@@ -20,7 +21,7 @@ const mockDestroySessionDELETE = rest.delete(
 export default {
   title: 'Private API / FormStepSaveModal',
   component: FormStepSaveModalComponent,
-  decorators: [ConfigDecorator], // DeprecatedRouterDecorator],
+  decorators: [FormikDecorator, ConfigDecorator], // DeprecatedRouterDecorator],
```
sergei-maertens, line 24: I don't think we need `FormikDecorator`, because the component itself already uses `<Formik>` inside, no?

src/components/modals/FormStepSaveModal.stories.js
```diff
@@ -37,45 +38,83 @@ export default {
...
+    const openModalButton = canvas.getByRole('button', {name: 'Open Modal'});
+    await expect(openModalButton).toBeVisible();
+    await userEvent.click(openModalButton);
+
+    const modal = await canvas.findByRole('dialog');
+    const emailInput = await within(modal).getByText('Uw e-mailadres');
+
+    await step('Test empty value', async () => {
+      await userEvent.click(emailInput);
+      await userEvent.tab();
```
sergei-maertens, line 109: better option would be to click the submit button, that should also trigger validation and it tests that we can submit after correcting the problem (i.e. the button is clickable). That is what is originally reported as being the issue.

I also believe there is still an unfixed bug that leaves the submit button disabled if backend validation errors arrive, so I think having a separate story (with a mock returning an HTTP 400 error) + closing and re-opening the modal should assert that the submit button is in clickable state (again).

sergei-maertens (changes requested):

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #695, opened 2024-07-01
[OF#4423] Update addressNL component to render single/double column

vaszig:
Partly fixes open-formulieren/open-forms#4423

### DIFF
--- a/src/formio/components/AddressNL.js
+++ b/src/formio/components/AddressNL.js
@@ -15,6 +15,8 @@ import {get} from 'api';
 import {TextField} from 'components/forms';
 import enableValidationPlugins from 'formio/validators/plugins';
 
+import './AddressNL.scss';
+
 const Field = Formio.Components.components.field;
 
 export default class AddressNL extends Field {
@@ -36,6 +38,7 @@ export default class AddressNL extends Field {
         defaultValue: {},
         validateOn: 'blur',
         deriveAddress: false,
+        layout: 'doubleColumn',
         openForms: {
           checkIsEmptyBeforePluginValidate: true,
         },
@@ -158,6 +161,7 @@ export default class AddressNL extends Field {
             initialValues={initialValues}
             required={required}
             deriveAddress={this.component.deriveAddress}
+            layout={this.component.layout}
             setFormioValues={this.onFormikChange.bind(this)}
           />
         </ConfigContext.Provider>
@@ -246,7 +250,7 @@ const addressNLSchema = (required, intl) => {
     });
 };
 
-const AddressNLForm = ({initialValues, required, deriveAddress, setFormioValues}) => {
+const AddressNLForm = ({initialValues, required, deriveAddress, layout, setFormioValues}) => {
   const intl = useIntl();
 
   const errorMap = (issue, ctx) => {
@@ -288,14 +292,16 @@ const AddressNLForm = ({initialValues, required, deriveAddress, setFormioValues}
         required={required}
         setFormioValues={setFormioValues}
         deriveAddress={deriveAddress}
+        layout={layout}
       />
     </Formik>
   );
 };
 
-const FormikAddress = ({required, setFormioValues, deriveAddress}) => {
+const FormikAddress = ({required, setFormioValues, deriveAddress, layout}) => {
   const {values, isValid, setFieldValue} = useFormikContext();
   const {baseUrl} = useContext(ConfigContext);
+  const useColumns = layout === 'doubleColumn';
 
   useEffect(() => {
     // *always* synchronize the state up, since:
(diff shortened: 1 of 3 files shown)


src/formio/components/AddressNL.js
```diff
@@ -36,6 +36,7 @@ export default class AddressNL extends Field {
         defaultValue: {},
         validateOn: 'blur',
         deriveAddress: false,
+        layout: {show: 'doubleColumn'},
```
sergei-maertens, line 39:
```suggestion
        layout:  'doubleColumn',
```
there is no `show` (or need for it) in the types :)

src/formio/components/AddressNL.js
```diff
@@ -158,6 +159,7 @@ export default class AddressNL extends Field {
             initialValues={initialValues}
             required={required}
             deriveAddress={this.component.deriveAddress}
+            layout={this.component.layout.show}
```
sergei-maertens, line 162:
```suggestion
            layout={this.component.layout}
```

src/formio/components/AddressNL.js
```diff
@@ -323,69 +326,78 @@ const FormikAddress = ({required, setFormioValues, deriveAddress}) => {
...
+      }
+
+      return field;
+    });
+    return layout === 'doubleColumn' ? (
+      <div className="openforms-columns">{wrappedFields}</div>
+    ) : (
+      wrappedFields
+    );
+  };
```
sergei-maertens, line 342: I'm wondering if we can clean this code up a bit by defining helper components instead of inlining everything.

Having something like this would be nice I suppose:

```jsx
const useColumns = layout === 'doubleColumn';

<div className="openforms-form-field-container">
  <Row useColumns={useColumns}>
    <Field useColumns={useColumns}>
      <PostCodeField required={required} autoFillAddress={autofillAddress} />
    </Field>
    <Field useColumns={useColumns}>
      <HouseNumberField required={required} autoFillAddress={autofillAddress} />
    </Field>
  </Row>
</div>
```

etc.

I find this reads much easier and you can manage the exact markup in the (new) `Row` and `Field` components.

I'd also instead of doing columns consider using a flat structure and do the layout with CSS grid, so you can toggle it into single/double column mode:

```css
.openforms-addressnl {
  display: grid;
  gap: 20px; // should be design token of course
}

.openforms-addressnl.openforms-addressnl--single-column {
  grid-template-columns: 1fr;  // make it single column
}

.openforms-addressnl.openforms-addressnl--double-column {
  grid-template-columns: 1fr 1fr; // make it double column
}
```

this has the benefit that the markup is much nicer

src/formio/components/AddressNL.stories.js
```diff
@@ -261,3 +261,39 @@ export const WithDeriveCityStreetNameNoData = {
...
+      handlers: [mockBAGDataGet],
+    },
+  },
+  args: {
+    type: 'addressNL',
+    key: 'addressNL',
+    label: 'Address NL',
+    extraComponentProperties: {
+      deriveAddress: true,
+      layout: {show: 'singleColumn'},
```
sergei-maertens, line 278:
```suggestion
      layout: 'singleColumn',
```

src/formio/components/AddressNL.stories.js
```diff
@@ -261,3 +261,39 @@ export const WithDeriveCityStreetNameNoData = {
...
+      handlers: [mockBAGDataGet],
+    },
+  },
+  args: {
+    type: 'addressNL',
+    key: 'addressNL',
+    label: 'Address NL',
+    extraComponentProperties: {
+      deriveAddress: true,
+      layout: {show: 'doubleColumn'},
```
sergei-maertens, line 296:
```suggestion
      layout: 'doubleColumn',
```

sergei-maertens (changes requested):

src/formio/components/AddressNL.js
```diff
@@ -324,67 +330,53 @@ const FormikAddress = ({required, setFormioValues, deriveAddress}) => {
...
+      <TextField
+        name="streetName"
+        label={
+          <FormattedMessage
+            description="Label for addressNL streetName read only result"
+            defaultMessage="Street name"
+          />
+        }
+        disabled
+      />
```
sergei-maertens, line 379: I think it makes more sense to swap the order of these two fields? So first street, then city.

src/formio/components/AddressNL.scss
```diff
@@ -0,0 +1,19 @@
+@import '~microscope-sass/lib/responsive';
```
sergei-maertens, line 1:
```suggestion
@use 'microscope-sass/lib/bem';

// @import instead of @use because breakpoints are defined globally
@import 'microscope-sass/lib/responsive';
```

src/formio/components/AddressNL.scss
```diff
@@ -0,0 +1,19 @@
...
+  grid-template-columns: 1fr;
+}
+
+.openforms-addressnl.openforms-addressnl--double-column {
+  grid-template-columns: 1fr 1fr;
+
+  @include mobile-only {
+    grid-template-columns: 1fr;
+  }
+}
```
sergei-maertens, line 19:
```suggestion
.openforms-addressnl {
  display: grid;
  grid-template-columns: var(--_of-addressnl-grid-template-columns, 1fr);
  grid-column-gap: var(--of-addressnl-column-gap, 8px);
  grid-row-gap: var(--of-addressnl-row-gap, var(--of-form-field-container-gap, 24px));
   
  @include bem.modifier('double-column') {
    --_of-addressnl-grid-template-columns: 1fr 1fr;
    
      @include mobile-only {
        grid-template-columns: 1fr;
      }
  }
}
```

* the `bem.modifier` makes sure to output the right class names
* the `grid-template-columns` variable approach is untested but I think it will work - the idea is that just override the variable for the more specific selector (and the default is single-column).

sergei-maertens (approved):

PR #708, opened 2024-09-05
Add introduction page before start page

sergei-maertens:
Closes open-formulieren/open-forms#4545


src/components/IntroductionPage/index.js
```diff
@@ -0,0 +1,36 @@
...
+      <Body
+        modifiers={['wysiwyg']}
+        component="div"
+        dangerouslySetInnerHTML={{__html: introductionPageContent}}
+      />
+
+      <Link to="/startpagina" component={ButtonLink} appearance="primary-action-button">
+        <FormattedMessage
+          description="Button text for link to continue from introduction page to start page"
+          defaultMessage="Continue ➜"
```
vaszig, line 27: Should this be inside the text and not an icon?

src/components/IntroductionPage/IntroductionPage.stories.js
```diff
@@ -0,0 +1,50 @@
...
+export default {
+  title: 'Views / IntroductionPage',
+  component: IntroductionPage,
+  decorators: [withForm, withRouter],
+  args: {
+    form: buildForm({introductionPageContent: DEFAULT_CONTENT}),
+  },
+};
+
+export const Default = {
```
vaszig, line 48: I am not sure if another one is needed that shows the navigation to the start page and back by using the progress indicator (just thinking..).

vaszig (approved):
Why do we need `IntroductionPage.js` file?For structure purposes?

sergei-maertens:
> Why do we need IntroductionPage.js file?For structure purposes?

no, it's an oversight! I'll remove it :)

src/components/IntroductionPage/index.js
sergei-maertens: I wanted to avoid discussion since the design includes it - I think at some point someone will complain about it and then we will need a mechanism to optionally set an icon, which is a complicated thing so went for the simple route for now.

src/components/IntroductionPage/IntroductionPage.stories.js
sergei-maertens: I didn't find it worth the effort tbh :)

PR #711, opened 2024-09-12
Fix step submission being blocked by backend validation error

sergei-maertens:
Closes open-formulieren/open-forms#4643


vaszig (approved):
One minor comment concerning the order of the stories..I guess it can be fixed in another PR

image

PR #715, opened 2024-09-25
[open-formulieren/open-forms#4600] Language switch now triggers an page reload

robinmolen:
This is to make sure that all language elements will be translated to the newly selected language

### DIFF
--- a/src/sdk.js
+++ b/src/sdk.js
@@ -70,6 +70,7 @@ class OpenForm {
       languageSelectorTarget,
       displayComponents = {}, // TODO: document as unstable API
       useHashRouting = false,
+      onLanguageChange,
     } = opts;
 
     this.targetNode = targetNode;
@@ -79,6 +80,7 @@ class OpenForm {
     this.lang = lang;
     this.displayComponents = {...defaultDisplayComponents, ...displayComponents};
     this.useHashRouting = useHashRouting;
+    this.onLanguageChange = typeof onLanguageChange === 'function' ? onLanguageChange : undefined;
 
     switch (typeof languageSelectorTarget) {
       case 'string': {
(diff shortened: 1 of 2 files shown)


src/sdk.js
```diff
@@ -173,8 +173,7 @@ class OpenForm {
   }
 
   async onLanguageChangeDone(newLanguagecode) {
-    this.formObject = await get(this.url);
-    this.render();
+    window.location.reload();
```
sergei-maertens, line 176: unfortunately this approach is too naive - we can only trigger this in the backend project where we know what the full page looks like, but on pages (like a CMS) outside of our control where the form is embedded, we must play nice and not take over control of the whole page.

I think adding an option to accept a callback to invoke when the language change is done/requested is the best approach, so that anyone who wishes to can hook into this to do the necessary thing.

sergei-maertens (changes requested):

robinmolen:
TODO in SDK:
- Add an property `onLanguageChange` that can be set by the implementation.

TODO in open-forms:
- Add onLanguageChange page reload to /src/openforms/forms/templates/forms/sdk_snippet.html

src/sdk.spec.js
```diff
@@ -112,7 +112,11 @@ describe('OpenForm', () => {
       await form.onLanguageChangeDone('en');
     });
 
-    expect(within(formRoot).getAllByText('English version').length).toBeGreaterThan(0);
+    // The language switch triggers a page reload
+    // After the reload, the english translations should be used
+    document.onload = () => {
+      expect(within(formRoot).getAllByText('English version').length).toBeGreaterThan(0);
+    };
```
sergei-maertens, line 119: this can be reverted now, no?

sergei-maertens (approved):

sergei-maertens:
Backports: 

* stable/2.4.x: 62a15ea1c9d2c1776fd1502a1f9be1dd0eb999e0
* stable/2.3.x: eb71900b861390b42f45b96d7c273a151cd2aaea

PR #716, opened 2024-10-01
:recycle: [open-formulieren/open-forms#4398] Cache initialDataReference to pass it with submissionCreate

stevenbal:
required for https://github.com/open-formulieren/open-forms/pull/4696

previously this was not passed properly, resulting in the initialDataReference not being sent on submissionCreate

### DIFF
--- a/src/components/Form.js
+++ b/src/components/Form.js
@@ -350,6 +350,8 @@ const Form = () => {
     ) : null;
 
   if (state.startingError) throw state.startingError;
+  const extraStartUrlParams = {};
+  if (initialDataReference) extraStartUrlParams.initial_data_reference = initialDataReference;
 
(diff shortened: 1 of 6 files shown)


sergei-maertens (commented):
Discussed in the office:

* caching in session storage seems risky because it may re-use this reference in a completely different form
* ideal approach: baking in this parameter in the login link, so that it's sent to the backend and when the backend redirects again to the SDK, it should pass along the parameter again
* if we can not reliably obtain this parameter from the URL, we may need to grab it in `App.js` and store it in state (in-memory), and expose it via React.createContext
* make sure this works when the introduction page is enabled, which is a page *before* `FormStart` - it should not swallow the initial data reference query param

Good luck :slightly_smiling_face:

PR #717, opened 2024-10-02
Accessibility improvements - form fields

robinmolen:
Closes open-formulieren/open-forms#4716

These are some accessibility changes regarding chapters 1.3.1 and 1.3.5 of the accessibility rapport

### DIFF
--- a/src/components/forms/Checkbox/Checkbox.js
+++ b/src/components/forms/Checkbox/Checkbox.js
@@ -40,10 +40,11 @@ const Checkbox = ({
         invalid={invalid}
         required={isRequired}
         appearance="custom"
+        aria-describedby={invalid ? `${id}-error-message` : undefined}
         {...inputProps}
       />
       <HelpText>{description}</HelpText>
-      {touched && <ValidationErrors error={error} />}
+      {touched && <ValidationErrors error={error} inputId={id} />}
     </FormField>
   );
 };
--- a/src/components/forms/Checkbox/Checkbox.js
+++ b/src/components/forms/Checkbox/Checkbox.js
@@ -22,6 +22,7 @@ const Checkbox = ({
 
   const {error, touched} = getFieldMeta(name);
   const invalid = touched && !!error;
+  const errorMessageId = invalid ? `${id}-error-message` : undefined;
 
   return (
     <FormField type="checkbox" invalid={invalid} className="utrecht-form-field--openforms">
@@ -40,11 +41,11 @@ const Checkbox = ({
         invalid={invalid}
         required={isRequired}
         appearance="custom"
-        aria-describedby={invalid ? `${id}-error-message` : undefined}
+        aria-describedby={errorMessageId}
         {...inputProps}
       />
       <HelpText>{description}</HelpText>
-      {touched && <ValidationErrors error={error} inputId={id} />}
+      {touched && <ValidationErrors error={error} id={errorMessageId} />}
     </FormField>
   );
 };
--- a/src/components/forms/EmailField/EmailField.js
+++ b/src/components/forms/EmailField/EmailField.js
@@ -11,7 +11,7 @@ EmailField.propTypes = {
   description: PropTypes.node,
   id: PropTypes.string,
   disabled: PropTypes.bool,
-  autocomplete: PropTypes.string,
+  autocomplete: PropTypes.oneOf(['email']),
 };
 
 export default EmailField;
--- a/src/components/forms/EmailField/EmailField.js
+++ b/src/components/forms/EmailField/EmailField.js
@@ -11,6 +11,7 @@ EmailField.propTypes = {
   description: PropTypes.node,
   id: PropTypes.string,
   disabled: PropTypes.bool,
+  autocomplete: PropTypes.string,
 };
 
 export default EmailField;
--- a/src/components/forms/RadioField/RadioField.js
+++ b/src/components/forms/RadioField/RadioField.js
@@ -35,6 +35,7 @@ export const RadioField = ({
   const {error, touched} = getFieldMeta(name);
   const invalid = touched && !!error;
   const descriptionid = `${id}-description`;
+  // @TODO check aria descriptions with screen reader... Is it still usable, with the description and error messages?
   return (
     <Fieldset
       className="utrecht-form-fieldset--openforms"
@@ -58,6 +59,7 @@ export const RadioField = ({
             id={`${id}-opt-${index}`}
             name={name}
             value={optionValue}
+            aria-describedby={invalid ? `${id}-error-message` : undefined}
             {...inputProps}
           />
           <Paragraph className="utrecht-form-field__label utrecht-form-field__label--radio">
@@ -74,7 +76,7 @@ export const RadioField = ({
       ))}
 
       <HelpText id={descriptionid}>{description}</HelpText>
-      {touched && <ValidationErrors error={error} />}
+      {touched && <ValidationErrors error={error} inputId={id} />}
     </Fieldset>
   );
 };
--- a/src/components/forms/RadioField/RadioField.js
+++ b/src/components/forms/RadioField/RadioField.js
@@ -35,7 +35,7 @@ export const RadioField = ({
   const {error, touched} = getFieldMeta(name);
   const invalid = touched && !!error;
   const descriptionid = `${id}-description`;
-  // @TODO check aria descriptions with screen reader... Is it still usable, with the description and error messages?
+  const errorMessageId = invalid ? `${id}-error-message` : undefined;
   return (
     <Fieldset
       className="utrecht-form-fieldset--openforms"
@@ -59,7 +59,7 @@ export const RadioField = ({
             id={`${id}-opt-${index}`}
             name={name}
             value={optionValue}
-            aria-describedby={invalid ? `${id}-error-message` : undefined}
+            aria-describedby={errorMessageId}
             {...inputProps}
           />
           <Paragraph className="utrecht-form-field__label utrecht-form-field__label--radio">
@@ -76,7 +76,7 @@ export const RadioField = ({
       ))}
 
       <HelpText id={descriptionid}>{description}</HelpText>
-      {touched && <ValidationErrors error={error} inputId={id} />}
+      {touched && <ValidationErrors error={error} id={errorMessageId} />}
     </Fieldset>
   );
 };
--- a/src/components/forms/ValidationErrors.js
+++ b/src/components/forms/ValidationErrors.js
@@ -2,7 +2,7 @@ import {FormFieldDescription} from '@utrecht/component-library-react';
 import PropTypes from 'prop-types';
 import React from 'react';
 
-const ValidationErrors = ({error = '', id = ''}) => {
+const ValidationErrors = ({error = '', id}) => {
   if (!error) return null;
   return (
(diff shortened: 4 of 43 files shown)


src/components/forms/Checkbox/Checkbox.js
```diff
@@ -40,10 +40,11 @@ const Checkbox = ({
         invalid={invalid}
         required={isRequired}
         appearance="custom"
+        aria-describedby={invalid ? `${id}-error-message` : undefined}
         {...inputProps}
       />
       <HelpText>{description}</HelpText>
-      {touched && <ValidationErrors error={error} />}
+      {touched && <ValidationErrors error={error} inputId={id} />}
```
sergei-maertens, line 47: isn't it more foo-proof to compute the error message ID once and pass it down to the component entirely, instead of only passing the `inputId`?

```jsx
const errorMessageId = invalid ? `${id}-error-message` : undefined;

<Component
  aria-describedby={errorMessageId}
>
    {touched && <ValidationErrors error={error} id={errorMessageId} />}
</Component>
```

src/components/forms/EmailField/EmailField.js
```diff
@@ -11,6 +11,7 @@ EmailField.propTypes = {
   description: PropTypes.node,
   id: PropTypes.string,
   disabled: PropTypes.bool,
+  autocomplete: PropTypes.string,
```
sergei-maertens, line 14:
```suggestion
  autocomplete: PropTypes.oneOf(['email']),
```
it doesn't make sense to allow other autocomplete strings, does it?

src/components/forms/RadioField/RadioField.js
```diff
@@ -35,6 +35,7 @@ export const RadioField = ({
   const {error, touched} = getFieldMeta(name);
   const invalid = touched && !!error;
   const descriptionid = `${id}-description`;
+  // @TODO check aria descriptions with screen reader... Is it still usable, with the description and error messages?
```
sergei-maertens, line 38: left over TODO

sergei-maertens (commented):
more review for the formio components will follow!

src/components/forms/Checkbox/Checkbox.js
robinmolen: Ahh.. yeah that would be better... Will fix

src/components/forms/EmailField/EmailField.js
robinmolen: Ah check, yeah i guess so. Or at least, for now that is the correct solution. I've changed it 👍

src/components/forms/ValidationErrors.js
```diff
@@ -2,17 +2,22 @@ import {FormFieldDescription} from '@utrecht/component-library-react';
 import PropTypes from 'prop-types';
 import React from 'react';
 
-const ValidationErrors = ({error = ''}) => {
+const ValidationErrors = ({error = '', id = ''}) => {
```
sergei-maertens, line 5: I think you want to leave this `undefined` instead of defaulting to an empty string so that if no `id` prop is passed, it doesn't get set. Otherwise this could possibly result in multiple divs with `<div id="">...</div>` and then IDs are not unique in the page.

The HTML 5 spec also says:

> When specified on [HTML elements](https://html.spec.whatwg.org/multipage/infrastructure.html#html-elements), the [id](https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute) attribute value must be unique amongst all the [IDs](https://dom.spec.whatwg.org/#concept-id) in the element's [tree](https://dom.spec.whatwg.org/#concept-tree) and must contain at least one character. The value must not contain any [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).

So unless React does something clever here, I'd stick to an explicit `id = undefined`

src/formio/utils.js
```diff
@@ -14,4 +14,38 @@ const escapeHtml = source => {
   return pre.innerHTML.replace(/"/g, '&quot;').replace(/'/g, '&apos;').replace(/&/g, '&amp;');
 };
 
-export {applyPrefix, escapeHtml};
+const setErrorAttributes = (elements, hasErrors, hasMessages, parentElement) => {
+  // Update the attributes 'aria-invalid' and 'aria-describedby' using hasErrors
+  elements.forEach(element => {
+    const errorMessageElementId = parentElement.querySelector('[ref="messageContainer"]')?.id;
```
sergei-maertens, line 20: Instead of passing `parentElement`, I think it makes more sense to pass the `this` object (which is the component) instance and then you can use `component.refs.messageContainer` - that's how Formio's ref mechanism is supposed to be used.

src/formio/components/FileField.js
```diff
@@ -291,6 +291,13 @@ class FileField extends Formio.Components.components.file {
     return super.validatePattern(file, val);
   }
 
+  setErrorClasses(elements, dirty, hasErrors, hasMessages) {
+    const input = this.element.querySelector('[ref="fileBrowse"]');
```
sergei-maertens, line 295: this should also use `this.refs.fileBrowse`

src/formio/templates/file.ejs
sergei-maertens, line 1: This looks odd in the Chromatic visual regression tests, especially on mobile there are now table headers but the rows/files themselves don't use this same layout.

I'll also need some more explanation on why the change to table is required :thinking:

sergei-maertens (commented):
I'd like to see some (Jest) tests that assert the aria-describedby behaves as expected. See all the annotations about missing test coverage. I'm also very interested in how `multiple: true` components behave!

src/components/forms/ValidationErrors.js
robinmolen: Ah yeah fair point.. Yeah you're right, that shouldn't be an empty string 👍

src/formio/components/FileField.js
robinmolen: ah okay, yeah if we can reach it via `this` then that would be better. I'll change it 👍

robinmolen:
Formio doesn't validate inputs inside a component with the `multiple` property individually. Meaning: if we have a textfield component with the `multiple` property, where the first input is valid and the second invalid, we cannot determine in the code which input is the valid/invalid one. This means we cannot add things like `aria-invalid="true"` to a specific input inside the component (because, again, we don't know which input is invalid).

So to prevent some weird issues for screenreader users, we cannot make components with the `multiple` property accessible. The accessibility change to link the error message to the input (with `aria-describedby`) and to add `aria-invalid="true"` won't happen for components with the `multiple` property

src/formio/templates/file.ejs
robinmolen: As we discussed, the `aria-labelledby` also provides the needed information to screenreaders. I've changed the table back to a list

src/formio/templates/file.ejs
```diff
@@ -6,10 +6,10 @@
         {% if (!ctx.disabled) { %}
           <div class="col-md-1"></div>
         {% } %}
-        <div class="col-md-{% if (ctx.self.hasTypes) { %}7{% } else { %}9{% } %}"><strong>{{ctx.t('File Name')}}</strong></div>
-        <div class="col-md-2"><strong>{{ctx.t('Size')}}</strong></div>
+        <div id="{{ctx.instance.id}}-{{ctx.component.key}}__file-list-header--file-name" class="col-md-{% if (ctx.self.hasTypes) { %}7{% } else { %}9{% } %}"><b>{{ctx.t('File Name')}}</b></div>
```
sergei-maertens, line 9: `strong` is typically used for its semantic meaning, and that has pretty much made `b` obsolete since that's only meant for visual presentation, which you can (and should) do with CSS.

So if the `strong` tag is an a11y problem, then we should remove it and apply the relevant/proper CSS instead of replacing it with `b` through a class name.

sergei-maertens (changes requested):

src/formio/templates/file.ejs
robinmolen: Ah shit, i thought i removed those...

Yep, will replace them

src/formio/utils.js
robinmolen: I've changed it to:

```
setErrorAttributes(..., ... ,..., this.refs.messageContainer.id);
```

Seeing how this function only needs the id of the message container

src/formio/templates/file.ejs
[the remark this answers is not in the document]
robinmolen: Ah yeah, fair enough

src/formio/templates/file.ejs
```diff
@@ -18,31 +18,34 @@
         <div class="row">
           {% if (!ctx.disabled) { %}
             <div class="col-md-1">
-              <button type="button" class="utrecht-button utrecht-button--subtle utrecht-button--danger" ref="removeLink" title="{{ctx.t('Remove')}}">
-                <i class="{{ctx.iconClass('trash-can')}}"></i>
+              <button type="button" class="utrecht-button utrecht-button--subtle utrecht-button--danger" ref="removeLink" title="{{ctx.t('Remove')}} {{file.originalName || file.name}}">
```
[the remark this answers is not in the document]
robinmolen: Mhh the apple screenreader didn't have any issues with the `title`, but its indeed better to be save 👍 

And if found what you described for the translations with variables. I hadn't considered different language structures...

src/formio/components/Checkbox.spec.js
```diff
@@ -0,0 +1,76 @@
...
+    const {form} = await renderForm();
+
+    const checkbox = screen.getByLabelText('Checkbox');
+
+    // Check and uncheck the checkbox to trigger the validation
+    await user.click(checkbox);
+    await user.click(checkbox);
+
+    // All selectboxes are marked as invalid and have aria-describedby and aria-invalid
+    expect(checkbox.className.includes('is-invalid')).toBeTruthy();
```
sergei-maertens, line 62: you can use [`toHaveClass`](https://github.com/testing-library/jest-dom?tab=readme-ov-file#tohaveclass)

src/formio/components/jest-util.js
```diff
@@ -0,0 +1,5 @@
+const getComponent = element => element.closest('[ref="component"]');
```
sergei-maertens, line 1: I think this will be mega-confusing with getting a formio component instance from the `WebForm`. Maybe rename it?

```suggestion
const getComponentNode = element => element.closest('[ref="component"]');
```

src/formio/components/Checkbox.spec.js
```diff
@@ -0,0 +1,76 @@
...
+    expect(checkbox.hasAttribute('aria-describedby')).toBeTruthy();
+    expect(checkbox.hasAttribute('aria-invalid')).toBeTruthy();
+    expect(checkbox.getAttribute('aria-invalid')).toBeTruthy();
```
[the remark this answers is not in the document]
robinmolen: Ah yeah thats a lot nicer (and probably better...). I'll change it

src/formio/components/jest-util.js
robinmolen: Mhh yeay, you might be right.. getComponentNode sounds a lot better.. 👍

sergei-maertens (approved):
Can be merged once #722 is merged!

sergei-maertens:
The test failures are known issues and will be fixed via #723

PR #718, opened 2024-10-08
:sparkles: [open-formulieren/open-forms#4544] Configurable form 'previous' link

robinmolen:
Closes open-formulieren/open-forms#4544

This PR allows you to configure (using the design tokens) if the previous link should be at the top, the bottom or both. By default the previous link is shown at the bottom (same as the current functionality).

This PR also adds the option to show an icon in the previous link (this is also configurable using design tokens)

To keep everything uniform, the arrow icon on the introduction page "start" button is changed to a Font Awesome icon.

### DIFF
--- a/src/components/ButtonsToolbar/index.js
+++ b/src/components/ButtonsToolbar/index.js
@@ -3,9 +3,9 @@ import React from 'react';
 
 import AbortButton from 'components/AbortButton';
 import {OFButton} from 'components/Button';
-import Link from 'components/Link';
 import {Literal, LiteralsProvider} from 'components/Literal';
 import Loader from 'components/Loader';
+import PreviousLink from 'components/PreviousLink';
 import {Toolbar, ToolbarList} from 'components/Toolbar';
 import {SUBMISSION_ALLOWED} from 'components/constants';
 
@@ -31,9 +31,7 @@ const ButtonsToolbar = ({
         <Toolbar modifiers={['mobile-reverse-order', 'bottom']}>
           <ToolbarList>
             {previousPage && (
-              <Link to={previousPage} onClick={onNavigatePrevPage}>
-                <Literal name="previousText" />
-              </Link>
+              <PreviousLink to={previousPage} onClick={onNavigatePrevPage} position="end" />
             )}
           </ToolbarList>
           <ToolbarList>
--- a/src/components/ButtonsToolbar/index.js
+++ b/src/components/ButtonsToolbar/index.js
@@ -3,14 +3,13 @@ import React from 'react';
 
 import AbortButton from 'components/AbortButton';
 import {OFButton} from 'components/Button';
-import {Literal, LiteralsProvider} from 'components/Literal';
+import {Literal} from 'components/Literal';
 import Loader from 'components/Loader';
 import PreviousLink from 'components/PreviousLink';
 import {Toolbar, ToolbarList} from 'components/Toolbar';
 import {SUBMISSION_ALLOWED} from 'components/constants';
 
 const ButtonsToolbar = ({
-  literals,
   canSubmitStep,
   canSubmitForm,
   canSuspendForm,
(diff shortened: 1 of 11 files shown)


src/components/ButtonsToolbar/index.js
sergei-maertens, line 1: I think we no longer need the `LiteralsProvider` in the `ButtonsToolbar` component, since it's now provided in the appointment components itself and the `FormStep` component. So this can likely be cleaned up!

That would obsolete the `literals` prop from the component too :)

src/components/IntroductionPage/index.js
```diff
@@ -21,12 +22,15 @@ const IntroductionPage = () => {
...
-          defaultMessage="Continue ➜"
-        />
-      </Link>
+      <span className="openforms-start-link">
+        <Link to="/startpagina" component={ButtonLink} appearance="primary-action-button">
+          <FormattedMessage
+            description="Button text for link to continue from introduction page to start page"
+            defaultMessage="Continue"
+          />
+          <FAIcon icon="arrow-right-long" extraClassName="openforms-start-link--icon" />
```
sergei-maertens, line 31:
```suggestion
          <FAIcon icon="arrow-right-long" extraClassName="openforms-start-link__icon" />
```
`__` signifies a child/element inside the block/component, `--` is for modifiers

src/components/PreviousLink.js
```diff
@@ -0,0 +1,31 @@
...
+      <Link to={to} onClick={onClick}>
+        <FAIcon icon="arrow-left-long" extraClassName="openforms-previous-link__icon" />
+        <Literal name="previousText" />
+      </Link>
+    </span>
+  );
+};
+
+PreviousLink.propTypes = {
+  to: PropTypes.string,
```
sergei-maertens, line 26:
```suggestion
  to: PropTypes.string.isRequired,
```

src/scss/components/_previous-link.scss
```diff
@@ -0,0 +1,17 @@
+@use 'microscope-sass/lib/bem';
+
+.openforms-previous-link {
+  @include bem.modifier('start') {
+    margin-block-end: 20px;
```
sergei-maertens, line 5: hrm, block-level elements should never define margins, since this breaks how you use a component inside other components (it should be limited to paddings). Can we do a different approach based on flexbox/grid where the row-gaps are used for spacing between blocks?

src/scss/components/_previous-link.scss
```diff
@@ -0,0 +1,17 @@
...
+    display: var(--of-previous-link-display-start, none);
+  }
+
+  @include bem.modifier('end') {
+    display: var(--of-previous-link-display-end, inline);
+  }
+
+  @include bem.element('icon') {
+    display: var(--of-previous-link-icon-display, none);
+    margin-inline-end: 10px;
```
sergei-maertens, line 15: preferably apply a `display: flex; flex-direction: row; gap: 8px;` for horizontal spacing instead of relying on margins, it's more consistent with the rest of the code.

For (default) spacings, we also use multiples of `4px` (historical reasons)

src/scss/components/_start-link.scss
```diff
@@ -0,0 +1,7 @@
+@use 'microscope-sass/lib/bem';
+
+.openforms-start-link {
+  @include bem.element('icon') {
+    margin-inline-start: 10px;
```
sergei-maertens, line 5: consider the flexbox approach here too :) `display: inline-flex` can be useful.

For a complicated example how we apply spacings, see `_app.scss`

sergei-maertens (changes requested):

src/scss/components/_start-link.scss
robinmolen: Mhh yeah that is an option... So that we have the flexibility to display the icon on either side?

I assume a `display: flex` (or `inline-flex` 😉 ) with a `row-gap: 10px;` would be enough, but i'll take a look

src/scss/components/_previous-link.scss
robinmolen: Mhh I think this would introduce a lot of changes for the form elements, but i'll take a look (that is precisely why is added the margin 😄 )

src/components/PreviousLink.js
robinmolen: 👍  I'll add it

src/components/IntroductionPage/index.js
robinmolen: ah shit, my bad 🙈 . I'll change it

robinmolen: With the flex-box change, this class will be removed

src/components/ButtonsToolbar/index.js
robinmolen: We still need it for the `SubmitRow` (component that uses the `ButtonsToolbar`). But we can move it to `SubmitRow`, and indeed remove the `literals` prop 👍

robinmolen: (and change the tests a bit, but thats also nothing special)

src/scss/components/_previous-link.scss
robinmolen: Lets discuss this tomorrow, this will limit some customization (for example the `--of-summary-progress-padding-block-end` design token will be obsolete. And if we would make the `row-gap` on the form step elements customizable, it would change the spacing between all elements)

src/scss/components/_previous-link.scss
robinmolen: I've changed it 👍

src/scss/components/_previous-link.scss
sergei-maertens: let's be pragmatic and go with the margin then for now, but please add a comment like:

```js
// NOTE: usually block elements should not define margins, but this is done deliberately
// See https://github.com/open-formulieren/open-forms-sdk/pull/718#discussion_r1791617060 for
// more discussion context.
```

sergei-maertens (approved):

PR #719, opened 2024-10-10
Improve a11y with aria attributes

robinmolen:
Closes open-formulieren/open-forms#4717

Some accessibility improvements involving aria tags, `role`'s and textual alternatives:

- The `Loader` element now has a textual alternative, which will be translated.
- The modals now use html buttons as buttons (instead of icons with a button role).
- The delete button/link for an file now specifies which file will be removed.
- And finally alerts now have `role="alert"`.

### DIFF
--- a/src/components/Errors/ErrorMessage.js
+++ b/src/components/Errors/ErrorMessage.js
@@ -17,7 +17,7 @@ const ErrorMessage = ({children, modifier = 'error'}) => {
   const errorRef = useScrollIntoView();
   if (!children) return null;
   return (
-    <Alert type={modifier} icon={ICONS[modifier]} ref={errorRef}>
+    <Alert type={modifier} icon={ICONS[modifier]} ref={errorRef} role="alert">
       {children}
     </Alert>
   );
--- a/src/components/Errors/ErrorMessage.js
+++ b/src/components/Errors/ErrorMessage.js
@@ -11,13 +11,20 @@ const ICONS = {
   ok: <i className="fa fas fa-check-circle" />,
 };
 
+const ARIA_TAGS = {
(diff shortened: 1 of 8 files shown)


src/components/Errors/ErrorMessage.js
```diff
@@ -17,7 +17,7 @@ const ErrorMessage = ({children, modifier = 'error'}) => {
   const errorRef = useScrollIntoView();
   if (!children) return null;
   return (
-    <Alert type={modifier} icon={ICONS[modifier]} ref={errorRef}>
+    <Alert type={modifier} icon={ICONS[modifier]} ref={errorRef} role="alert">
```
sergei-maertens, line 20: shouldn't we limit this to the warning/error modifiers? I'm not convinced that info/success require the users immediate attention.

Quoting from https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role

> Dynamic changes that are less urgent should use a less aggressive method, such as including `aria-live="polite"` or using an other live region role like [status](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/status_role)

robinmolen: Ah check, yeah good one. The role status sounds like a good option for info/success

robinmolen: If changed the role to `status` and added a `aria-live="polite"` to the info and ok ErrorMessages, as suggested by https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions#roles_with_implicit_live_region_attributes

sergei-maertens (approved):

PR #720, opened 2024-10-17
Textual and viewport accessibility improvements

robinmolen:
Partially closes open-formulieren/open-forms#4718

Changes regarding 1.4.10 Reflow

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
         "@floating-ui/react": "^0.26.9",
         "@formio/protected-eval": "^1.2.1",
         "@fortawesome/fontawesome-free": "6.4.0",
-        "@open-formulieren/design-tokens": "^0.54.0",
+        "@open-formulieren/design-tokens": "^0.55.0",
         "@open-formulieren/formiojs": "^4.13.14",
         "@open-formulieren/leaflet-tools": "^1.0.0",
         "@sentry/react": "^6.13.2",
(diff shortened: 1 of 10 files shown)


src/scss/components/_progress-indicator.scss
```diff
@@ -24,6 +24,8 @@
  */
 .openforms-progress-indicator {
   @include card.card('progress-indicator');
+  overflow: auto;
+  max-height: 100vh;
```
sergei-maertens, line 28:
```suggestion
  overflow: auto;
  max-height: 100vh;
```
this breaks the story http://localhost:6006/?path=/story/private-api-progressindicator--default on "Small mobile" viewport, presumably because the screen height is less than 100vh - but this can happen if there's no introduction or very little content.

Additionally, instead of `max-height`, use `max-block-size` for a logical property (`max-width` is accordingly replaced with `max-inline-size`). This ensures we're prepared for other languages that don't read horizontal, but vertical.

Finally, especially here you want to use `dvh` instead `vh` because mobile browsers are weird.

sergei-maertens: discussion in office - not sure if these are actually needed at all, make sure to test on different devices with many variations in amount of content etc. :grimacing:

robinmolen: Yeah we don't need it..

I'll drop this change/commit

robinmolen: I've made a fix for the underlying issue, as discussed in the office.

In the js a `verticalSpaceUsed` calculation is done for mobile views. This calculates the height of the topbar, which is passed to the css as a variable so that the correct `max-block-size` can be set. The problem is, that this calculation is only done on initial paint. Meaning, if you open the page and *THEN* zoom-in, the `max-block-size` will be set to none (because of the missing css variable)

I've added a window resize listener, that triggers the js calculation and sets/updates the css variable.

sergei-maertens (approved):

PR #726, opened 2024-10-28
Implement soft required validation

sergei-maertens:
Closes open-formulieren/open-forms#4546

This went better than I had anticipated.

* Implement 'content' component variant for the soft required validation errors
* Styled it as a common/regular warning alert
* Added escape hatches for background color/font color/font-weight in case it needs to look different from a regular alert
* Ensured accessibility (linked list header, form fields refer back to content)
* Added stories and unit tests
* Checked the template interpolation/rendering for XSS, seems to be okay.

Final tasks to wrap this up:

* Update formio translations in the backend
* Review documentation if anything extra is needed

### DIFF
--- a/src/formio/components/SoftRequiredErrors.js
+++ b/src/formio/components/SoftRequiredErrors.js
@@ -39,9 +39,11 @@ class SoftRequiredErrors extends FormioContentField {
     if (!missingFieldLabels.length) return '';
 
     const missingFieldsMarkup = this.renderTemplate('missingFields', {labels: missingFieldLabels});
-    return this.interpolate(this.component.html, {
+    const content = this.interpolate(this.component.html, {
       missingFields: missingFieldsMarkup,
     });
+
+    return `<div id="${this.id}-content">${content}</div>`;
   }
 }
 
--- a/src/formio/components/SoftRequiredErrors.js
+++ b/src/formio/components/SoftRequiredErrors.js
@@ -1,0 +1,34 @@
+import {Formio} from 'react-formio';
+
+const FormioContentField = Formio.Components.components.content;
+
+class SoftRequiredErrors extends FormioContentField {
+  static schema(...extend) {
+    return FormioContentField.schema(
+      {
+        type: 'softRequiredErrors',
+        key: 'softRequiredErrors',
+        label: '', // not displayed anyway
+      },
+      ...extend
+    );
+  }
+
+  constructor(component, options, data) {
+    component.refreshOnChange = true;
+    super(component, options, data);
+  }
+
+  get content() {
+    const missingFieldLabels = ['Foo', 'Bar'];
+
+    if (!missingFieldLabels.length) return '';
+
+    const missingFieldsMarkup = this.renderTemplate('missingFields', {labels: missingFieldLabels});
+    return this.interpolate(this.component.html, {
+      missingFields: missingFieldsMarkup,
+    });
+  }
+}
+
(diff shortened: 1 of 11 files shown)


src/formio/components/SoftRequiredErrors.stories.js
```diff
@@ -0,0 +1,111 @@
...
+        url: UPLOAD_URL,
+        label: 'Soft required file',
+        multiple: false,
+        openForms: {softRequired: true},
+      },
+      {
+        type: 'textfield',
+        key: 'textfield',
+        label: 'Soft required text',
+        openForms: {softRequired: true},
```
robinmolen, line 28: Textfield cannot be `softRequired`, right? I thought this was only for file upload fields :)

sergei-maertens: officially they can't (yet), but the implementation was meant to be generic so that it also works for text fields :wink:

src/formio/components/SoftRequiredErrors.js
```diff
@@ -0,0 +1,59 @@
...
+
+    if (!missingFieldLabels.length) return '';
+
+    const missingFieldsMarkup = this.renderTemplate('missingFields', {labels: missingFieldLabels});
+    const content = this.interpolate(this.component.html, {
+      missingFields: missingFieldsMarkup,
+    });
+
+    return `
+      <div id="${this.id}-content" class="utrecht-alert utrecht-alert--warning openforms-soft-required-errors">
```
robinmolen, line 47: To add some context for users with a visual disability:

```suggestion
      <div id="${this.id}-content" class="utrecht-alert utrecht-alert--warning openforms-soft-required-errors" role="alert">
```

src/formio/components/SoftRequiredErrors.stories.js
robinmolen: Ah nevermind.. I'm seeing your comment in the Textfield.spec.js 👍

src/formio/components/SoftRequiredErrors.js
sergei-maertens: From https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role:

> The alert role is for important, and usually time-sensitive, information.

I don't find this time sensitive, especially because it's also displayed at the beginning of a form step that is yet to be filled out.

> The alert role should only be used for information that requires the user's immediate attention, for example:

This is not the case - it becomes important once you are about to submit the form.

I think the `role="status"` is more appropriate: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/status_role

src/formio/utils.js
```diff
@@ -45,4 +48,39 @@ const setErrorAttributes = (elements, hasErrors, hasMessages, messageContainerId
...
+  const softRequiredIds = [];
+  FormioUtils.eachComponent(component.root.components, component => {
+    if (component.type === 'softRequiredErrors') {
+      const id = `${component.id}-content`;
+      softRequiredIds.push(id);
+    }
+  });
+
+  // Update the attribute 'aria-describedby' based on whether the component is empty
+  elements.forEach(element => {
```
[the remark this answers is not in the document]
sergei-maertens: yeah wasn't sure if this warranted a refactor or not since your helper also touches aria-invalid and mine doesn't so there is going to be some duplication, but I'll take a look.

PR #728, opened 2024-11-11
:bug: [open-formulieren/open-forms#4772] Fix select component with integer values

stevenbal:
Closes https://github.com/open-formulieren/open-forms/issues/4772


stevenbal:
closing in favor of other PRs to ensure the dataType is correctly set when creating the component in the formbuilder

PR #729, opened 2024-11-11
Cosign improvements

sergei-maertens:
Closes open-formulieren/open-forms#4320 (partially), required for backend PR open-formulieren/open-forms#4808


vaszig (approved):
Should we change `CosignStart.js` to `CosignStart.jsx`?

PR #731, opened 2024-11-12
[Backport 2.3.x] 🐛 [open-formulieren/open-forms#4772] Fix select component with integer values

stevenbal:
Fixes open-formulieren/open-forms#4772 partially

This was fixed in the formio-builder, but in order to fix it for OF 2.8.x and 2.7.x, this SDK workaround is applied

### DIFF
--- a/src/formio/components/Select.js
+++ b/src/formio/components/Select.js
@@ -7,6 +7,14 @@ import {applyPrefix} from '../utils';
  * Extend the default select field to modify it to our needs.
  */
 class Select extends Formio.Components.components.select {
+  constructor(component, options, data) {
+    super(component, options, data);
+    // Fix for https://github.com/open-formulieren/open-forms/issues/4772
+    // ensure the datatype is always set to string to avoid formio casting it to other
+    // types (such as integer)
+    component.dataType = 'string';
+  }
+
   get inputInfo() {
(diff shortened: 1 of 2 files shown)


src/formio/components/Select.stories.js
```diff
@@ -1,3 +1,4 @@
+import {expect, jest} from '@storybook/jest';
 import {userEvent, within} from '@storybook/test';
```
sergei-maertens, line 2: you should import those from `@storybook/test`, not `@storybook/jest`

src/formio/components/Select.stories.js
```diff
@@ -131,3 +132,57 @@ export const MultilineOptions = {
     },
   },
 };
+
+export const WithIntegerValues = {
```
sergei-maertens, line 136: can't we write a unit test for this instead like in `src/jstests/formio/components/textfield.spec.js`? Chromatic snapshots cost money, and checking the type of values is really not an interaction test.

sergei-maertens (changes requested):

src/formio/components/Select.stories.js
stevenbal: Replaced it with a regular test

sergei-maertens (approved):

Issue #724, opened 2024-10-24
Update toolchain (Jest/Vitest/ViteJS/TS)


sergei-maertens:
Umbrella ticket for tasks and their order to update our toolchain to modern standards.

Short summary - CRA is dead and for us ViteJS looks very viable. We need to upgrade because at some point it won't be possible to upgrade Storybook etc. anymore or we'll be forced to work with deprecated APIs/tooling, which hurts development speed.

- [x] Upgrade Storybook (#723)
- [x] Upgrade MSW from 1.x to 2.x (#723)
- [x] Upgrade Jest / unit tests tooling (#723)
- [x] Set up Codecov flags for more accurate codecov checks (?)
- [x] Switch to vitest
    - [x] Ensure we still have coverage reporting
- [x] Switch Storybook to vite-builder
- [x] Replace CRA setup with ViteJS (keep ESLint etc., tough one!)
    - [x] Build NPM package ESM with Vite
    - [x] Build UMD library with Vite
    - [x] Swap out `dist/` for `dist-vite/`
- [ ] #445 
- [x] Set up ESLint in CI

sergei-maertens:
The Typescript task is tracked separately, moving to implemented.

PR #741, opened 2024-11-26
Migrate jest tests to vitest

sergei-maertens:
Part of #724

Migrating to the ViteJS ecosystem step by step - sorting out the test runner before the bundler allows us to prepare and solve the hard problems. Storybook is also going the Vite direction, and this allows us to work around some weirdness from Jest.


src/vitest.setup.mjs
```diff
@@ -0,0 +1,15 @@
+// jest-dom adds custom jest matchers for asserting on DOM nodes.
+// allows you to do things like:
+// expect(element).toHaveTextContent(/react/i)
+// learn more: https://github.com/testing-library/jest-dom
+import '@testing-library/jest-dom';
```
robinmolen: I think this needs to be updated :) https://github.com/testing-library/jest-dom?tab=readme-ov-file#with-vitest

```suggestion
import '@testing-library/jest-dom/vitest';
```

sergei-maertens: it's super confusing, as I understood this should not be necessary when configuring vitest to inject the globals, which I did for compatibility with jest: https://vitest.dev/guide/migration.html#globals-as-a-default

This should install the automatic cleanup and looking at the code it does extend the global expect (if not, tests would fail too).

Looking at the source: https://github.com/testing-library/jest-dom/blob/main/src/vitest.js it's extending the expect from vitest rather than the global (which would have the same effect), but because of the globals option I don't think it's necessary.

robinmolen (approved):
Nice to see you've also re-written some of the tests!

Looks good to me 👍

src/vitest.setup.mjs
sergei-maertens: right, I didn't take into account a possible jest-dom upgrade yet. Our version (v5) doesn't have the `vitest` module yet to import, and I'd like to hold of on upgrading jest-dom for now since it's also used for Storybook tests.

PR #747, opened 2024-12-03
Custom map backgrounds

robinmolen:
Part of open-formulieren/open-forms#2173

Adds a new property to the LeaftletMap component, named `tileUrl`. This allows the use of custom map backgrounds.

### DIFF
--- a/src/formio/components/Map.jsx
+++ b/src/formio/components/Map.jsx
@@ -115,6 +115,7 @@ export default class Map extends Field {
             onMarkerSet={this.onMarkerSet.bind(this)}
             defaultCenter={defaultCenter}
             defaultZoomLevel={zoom || DEFAULT_ZOOM}
+            tileLayerUrl={this.component.tileLayerUrl}
           />
         </ConfigContext.Provider>
       </IntlProvider>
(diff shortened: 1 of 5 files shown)


src/formio/components/Map.jsx
```diff
@@ -115,6 +115,7 @@ export default class Map extends Field {
             onMarkerSet={this.onMarkerSet.bind(this)}
             defaultCenter={defaultCenter}
             defaultZoomLevel={zoom || DEFAULT_ZOOM}
+            tileUrl={this.component.url}
```
sergei-maertens, line 118: I think it would be wise to already define the type in the `types` repository for this, when we have the custom renderer we can build on top of that.

Additionally, I think a property `tileLayerUrl` is much better than just `url` - the latter is too generic and will be confusing when you add URLs/configurations for the feature layers.

robinmolen: With the types repository you mean https://github.com/open-formulieren/types, right?

Check, yeah it would be best to keep that naming consistent.. I'll change it

sergei-maertens: indeed

sergei-maertens (approved):

PR #752, opened 2024-12-12
Avoid displaying layout components in editgrid summary

sergei-maertens:
Closes open-formulieren/open-forms#4809

Updated the Formio templates for edit grid rows to use the utility which skips over layout components.

### DIFF
--- a/src/formio/components/EditGrid.stories.jsx
+++ b/src/formio/components/EditGrid.stories.jsx
@@ -268,3 +268,93 @@ export const AddressNLWithNoData = {
     await userEvent.click(saveButton);
   },
 };
+
+export const WithLayoutComponents = {
+  render,
+  args: {
+    key: 'editgrid',
+    label: 'Nested layout components',
+    extraComponentProperties: {
+      hideLabel: false,
+      components: [
+        {
+          type: 'textfield',
+          key: 'text',
+          label: 'A text field',
+        },
+        {
+          type: 'content',
+          key: 'test1',
+          input: false,
+          label: 'WYSIWYG content',
+          html: '<p>Bonjour</p>',
+        },
+        {
+          type: 'fieldset',
+          key: 'level2Group',
+          label: 'Nested fieldset',
+          hideHeader: true,
+          components: [
+            {
+              type: 'textfield',
+              key: 'nestedText',
+              label: 'Nested text',
+            },
+          ],
+        },
+        {
+          type: 'columns',
+          key: 'columns',
+          label: 'Nested columns',
+          columns: [
+            {
+              size: 6,
+              components: [{type: 'number', key: 'number1', label: 'Number 1'}],
+            },
+            {
+              size: 6,
+              components: [{type: 'number', key: 'number2', label: 'Number 2'}],
+            },
+          ],
+        },
+      ],
+      inlineEdit: false,
+      disableAddingRemovingRows: false,
+      addAnother: 'Nog één toevoegen',
+      saveRow: 'Bevestigen',
+      removeRow: 'Verwijderen',
+    },
+
+    data: [
+      {
+        text: 'Some text field value',
+        nestedText: 'Nested text field value',
+        number1: 42,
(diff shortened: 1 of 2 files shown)


src/formio/components/EditGrid.stories.jsx
```diff
@@ -268,3 +268,93 @@ export const AddressNLWithNoData = {
...
+    expect(await canvas.findByText('Nested text field value')).toBeVisible();
+
+    expect(await canvas.findByText('Number 1')).toBeVisible();
+    expect(await canvas.findByText('42', {exact: true})).toBeVisible();
+
+    expect(await canvas.findByText('Number 2')).toBeVisible();
+    expect(await canvas.findByText('420', {exact: true})).toBeVisible();
+
+    await step('Layout component labels not shown', () => {
+      expect(canvas.queryByText('WYSIWYG content')).not.toBeInTheDocument();
```
viktorvanwijk, line 355: If I understand the problem correctly, a content field in a repeating group showed the label instead of the actual content. I see you added a check that the label 'WYSIWYG content' is no longer visible. Would it be useful to also add a check that the content ('Bonjour!') is actually displayed now?

src/formio/templates/editGridRow.ejs
```diff
@@ -1,14 +1,14 @@
 <dl class="utrecht-data-list utrecht-data-list--html-dl utrecht-data-list--rows">
-  {% for (const key in ctx.flattenedComponents) { %}
-    {% if (ctx.isVisibleInRow(ctx.flattenedComponents[key]) && ctx.flattenedComponents[key].label) { %}
```
viktorvanwijk, line 3: You removed the check that a label exists, right? Won't this change the behaviour of this if statement, or does `isVisibleInRow(component)` automatically mean it has a label as well?

src/formio/components/EditGrid.stories.jsx
sergei-maertens: Not really, but ultimately it's a design choice to not display layout components so I'm sure someone at some point will ask for this :laughing: 

For some more context - Formio components can be put roughly in two categories: input components, and layout components. Input components are actual form fields, like a textfield, number, dropdown... Layout components are used to organize the structure of the form, like a fieldset for a set of related inputs, or columns layouts. We support three layout component types: fieldsets, columns and content (WYSIWYG). Layout components never have a value associated with them.

The default behaviour of Formio is to only display input components in this summary - when you edit the item again, layout components are also displayed since they are related to UX. However, the default template from Formio shows data in a tabular format, and that is something we had to override. When we did that override, we didn't take into account that layout components aren't/shouldn't be displayed, and they were treated like input components: display label followed by the value.

Displaying columns/fieldsets here doesn't make any sense at all, for content you could argue that displaying the WYSIWYG could be desired, but I went for the simple approach of just skipping all layout components alltogether.

src/formio/templates/editGridRow.ejs
sergei-maertens: Yeah, because label is one of the required properties, and if it's missing, it'll just end up with an empty string but shouldn't break the layout. `key`, `type` and `label` are properties for which we enforce in our https://github.com/open-formulieren/formio-builder that a value must be provided/set, and this is also expressed in our Typescript types. So, we can assume that there will always be a label value in each component, even if it's not relevant (like for content component).

The `ctx.isVisibleInRow` is internal Formio stuff related to the logic. We don't expose all this logic to end-users (and we have a long-standing backend issue related to this), but in short - logic on/in repeating group items is not supported.

This template code is actually closer to the default template code of Formio which also loops over `ctx.eachComponent` rather than `ctx.flattenedComponents`, as the former skips over layout components while the latter doesn't.

src/formio/components/EditGrid.stories.jsx
viktorvanwijk: OK, clear

viktorvanwijk (approved):

PR #766, opened 2025-01-06
Drop Create-React-App

### LINKED ISSUE #724: Update toolchain (Jest/Vitest/ViteJS/TS)

sergei-maertens:
Partly closes #724

Good bye and thanks for all the fish.


robinmolen:
Another question about the `vite.config.mts`:

Im seeing a TS error about the test config. I took a 'quick' look and came across this stack overflow discussion (yes, its a bit older..) https://stackoverflow.com/questions/72146352/vitest-defineconfig-test-does-not-exist-in-type-userconfigexport.

One of the suggestions was about misaligned vite/vitest versions, so i took a look at that.. The vitest version that we use, "^2.1.8", is dependent on vite "^5.0.0". So i've changes the vite version locally to "^5.0.0", which solves the issue..

Is there a reason why we use version "^6.0.5" for vite?


Another solution might would be to define the type ourselfs, or to move the test config to a separate file?

sergei-maertens:
> `vite.config.mts` has a type import from 'rollup', but this isn't defined as a dependency in the project.. Is `rollup` missing from the package.json, or am i missing something? :)

Vite depends on rollup, so it's a transitive dependency :) The production build is made using rollup, in dev mode Vite uses native browser modules.

sergei-maertens:
> Another question about the `vite.config.mts`:
> 
> Im seeing a TS error about the test config. I took a 'quick' look and came across this stack overflow discussion (yes, its a bit older..) https://stackoverflow.com/questions/72146352/vitest-defineconfig-test-does-not-exist-in-type-userconfigexport.

Yes - it's a known issue - it's because Vitest officially only supports Vite 5 and support for Vite 6 is currently in beta (or RC) - I'm waiting for Vitest 3 to be stable and then we can upgrade and it should resolve the type error! There's a github issue in vitest for it. There are some dirty workarounds, but ultimately it's something that we just have to wait for.

> Is there a reason why we use version "^6.0.5" for vite?

It released just after doing the initial configuration and everything appears to work properly with Vite 6, and I don't want to "start" a new toolchain already on an outdated major version.

IMO we can just ignore the type error for the time being because it doesn't break anything except cause some red lines in your editor, but my expectation and hope is that we won't have to touch this config file for a while now.

PR #769, opened 2025-01-07
Convert cosign v5 react-router routes to data router format

sergei-maertens:
Closes open-formulieren/open-forms#4929 partially.

* Added tests for Cosign flow including the router integration
* Refactored/re-organized the cosign components
* Extracted the `<Routes />` nested inside the component into data router, replaced with `<Outlet />`

The `check` path is a nice candidate for a loader at the route level, but that's out of scope for this PR.


src/components/CoSign/Context.js
```diff
@@ -0,0 +1,28 @@
...
+      onCosignComplete,
+    }}
+  >
+    {children}
+  </CosignContext.Provider>
+);
+
+CosignProvider.propTypes = {
+  reportDownloadUrl: PropTypes.string,
+};
```
sergei-maertens, line 24: Add prop type for onCosignComplete

robinmolen (approved):

PR #771, opened 2025-01-10
React Router upgrade: main app (part 1)

sergei-maertens:
Depends on #769 
Partially fixes open-formulieren/open-forms#4929

Extracted some of the routes from the `Form` component and refactored some implementation details.

I've run the backend e2e tests against a (production) build of this branch and they all pass.


src/hooks/useQuery.js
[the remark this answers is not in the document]
sergei-maertens: yeah that's the plan - do this afterwards!

src/components/App.jsx
```diff
@@ -9,6 +9,7 @@ import {
   appointmentRoutes,
   manageAppointmentRoutes,
 } from 'components/appointments';
+import formRoutes from 'components/formRoutes';
```
sergei-maertens, line 12: once we're around to this refactoring, I want to move the route definition files too, since this is not a component :)

src/components/Form.jsx
```diff
@@ -354,39 +323,9 @@ const Form = () => {
...
-              submission={state.submission}
-              onFormStart={onFormStart}
-              onDestroySession={onDestroySession}
-              initialDataReference={initialDataReference}
-            />
-          </ErrorBoundary>
-        }
-      />
-
       <Route
```
sergei-maertens, line 329: rest will be done in a different PR, this one is already quite big

src/components/Form.spec.jsx
```diff
@@ -51,69 +49,6 @@ const Wrapper = ({form = buildForm(), initialEntry = '/startpagina'}) => {
   );
 };
 
-test('Start form anonymously', async () => {
```
sergei-maertens, line 54: tests are moved/merged into the FormStart tests

src/components/FormLandingPage.jsx
```diff
@@ -0,0 +1,15 @@
+import {Navigate} from 'react-router-dom';
+
+import useFormContext from 'hooks/useFormContext';
+import useInitialDataReference from 'hooks/useInitialDataReference';
+
+const FormLandingPage = () => {
+  const {introductionPageContent = ''} = useFormContext();
+  const {addInitialDataReference} = useInitialDataReference();
+  const startPageUrl = introductionPageContent ? 'introductie' : 'startpagina';
```
robinmolen, line 9: Im wondering if we can replace the route strings with constants/enums/something.

If you agree, this is probably better suited as a stand-alone improvement :)

src/components/Form.jsx
```diff
@@ -454,14 +401,45 @@ const Form = () => {
...
 
 Form.propTypes = {};
 
+const SubmissionContext = React.createContext({
+  submission: null,
+  onSubmissionObtained: () => {},
+  onDestroySession: () => {},
+});
+
+const SubmissionProvider = ({
```
robinmolen, line 425: I think it would be nicer to move this to a separate file.
Maybe a `src/context/SubmissionContext`, which exports the `SubmissionProvider` and the `useSubmissionContext`

sergei-maertens: I disagree because I'd prefer keeping the code together where it's used instead of smearing out stuff that is used together across different files. When all this is done, the `Form` component should be a small component with only minimal callbacks defined at this layer - the rest should be pushed down to individual routes.

robinmolen (approved):
Looks great to me! Love the big cleanup 👌 👌

PR #774, opened 2025-01-15
React router upgrade part 2

sergei-maertens:
Partly closes open-formulieren/open-forms#4929

* Refactored the `FormStep` component so that it can be extracted into the statically defined routes
* Had to refactor some supporting components around it to achieve that :-)

### DIFF
--- a/src/components/Form.jsx
+++ b/src/components/Form.jsx
@@ -29,7 +29,6 @@ import {
   STEP_LABELS,
   SUBMISSION_ALLOWED,
 } from 'components/constants';
-import {findNextApplicableStep} from 'components/utils';
 import {flagActiveSubmission, flagNoActiveSubmission} from 'data/submissions';
 import useAutomaticRedirect from 'hooks/useAutomaticRedirect';
 import useFormContext from 'hooks/useFormContext';
@@ -178,16 +177,6 @@ const Form = () => {
     });
     flagActiveSubmission();
     setSubmissionId(submission.id);
-  };
-
-  const onStepSubmitted = async formStep => {
-    const currentStepIndex = form.steps.indexOf(formStep);
-
-    const nextStepIndex = findNextApplicableStep(currentStepIndex, state.submission);
-    const nextStep = form.steps[nextStepIndex]; // will be undefined if it's the last step
-
-    const nextUrl = nextStep ? `/stap/${nextStep.slug}` : '/overzicht';
-    navigate(nextUrl);
   };
 
   const onSubmitForm = processingStatusUrl => {
@@ -380,16 +369,7 @@ const Form = () => {
         element={
           <ErrorBoundary useCard>
             <SessionTrackerModal expiryDate={expiryDate}>
-              <RequireSubmission
-                form={form}
-                submission={state.submission}
-                onLogicChecked={submission =>
-                  dispatch({type: 'SUBMISSION_LOADED', payload: submission})
-                }
-                onStepSubmitted={onStepSubmitted}
-                component={FormStep}
-                onDestroySession={onDestroySession}
-              />
+              <RequireSubmission submission={state.submission} component={FormStep} />
             </SessionTrackerModal>
           </ErrorBoundary>
         }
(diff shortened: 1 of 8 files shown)


src/components/FormStep/data.js
sergei-maertens, line 1: This was just extracted from index.jsx and moved into its own module to de-clutter, nothing changed.

src/components/FormStep/index.jsx
```diff
@@ -509,7 +410,7 @@ const FormStep = ({form, submission, onLogicChecked, onStepSubmitted, onDestroyS
       previouslyCheckedData.current = cloneDeep(data);
 
       // report back to parent component
-      onLogicChecked(submission, step);
```
sergei-maertens, line 512: the second argument was not expected/required so it's dropped now.

src/components/FormStep/index.jsx
```diff
@@ -602,9 +503,15 @@ const FormStep = ({form, submission, onLogicChecked, onStepSubmitted, onDestroyS
...
-    onStepSubmitted(formStep);
+    const {submission: updatedSubmission} = await doLogicCheck(submissionStep.url, data);
+    onSubmissionObtained(updatedSubmission); // report back to parent component
+
+    // navigate to the next page (either the next step or the overview)
+    const currentStepIndex = form.steps.indexOf(formStep);
+    const nextStepIndex = findNextApplicableStep(currentStepIndex, submission);
+    const nextStep = form.steps[nextStepIndex]; // will be undefined if it's the last step
+    const nextUrl = nextStep ? `/stap/${nextStep.slug}` : '/overzicht';
+    navigate(nextUrl);
```
sergei-maertens, line 514: localized/moved the callback from `Form` component into `Formstep` component.

src/components/RequireSubmission.jsx
```diff
@@ -1,44 +1,68 @@
...
 import {IsFormDesigner} from 'headers';
+import useFormContext from 'hooks/useFormContext';
 
 /**
  * Higher order component to enforce there is an active submission in the state.
  *
  * If there is no submission, the user is forcibly redirected to the start of the form.
+ *
+ * Provide either the component or children prop to render the actual content. The
+ * `component` prop is deprecated in favour of specifying explicit elements.
```
robinmolen, line 16: Nice, i completely agree. Not a fan of the magic component/props.

Im see that this is used in a couple of places. Are you gonna change those in future PR's? Otherwise, maybe nice to change them now (so that we don't have to support deprecated stuff)

src/components/Sessions/SessionTrackerModal.jsx
```diff
@@ -9,8 +9,10 @@ import {OFButton} from 'components/Button';
 import ErrorMessage from 'components/Errors/ErrorMessage';
 import {Toolbar, ToolbarList} from 'components/Toolbar';
 import Modal from 'components/modals/Modal';
+import useSessionTimeout from 'hooks/useSessionTimeout';
 
 const WARN_SESSION_TIMEOUT_FACTOR = 0.9; // once 90% of the session expiry time has passed, show a warning
+const TEN_HOURS_MS = 1000 * 3600 * 10; // ten hours in miliseconds
```
robinmolen, line 15: Funny that you changed this one around (from `10 * 3600 * 1000`), i find that notation a bit easier to read

src/components/Form.jsx
robinmolen, line 1: This bug is currently still in the code, right? (Although, moved to a timeout of 10 hours)

Can't we disable the session timeout completely for these pages? Or am i missing something?

src/components/RequireSubmission.jsx
sergei-maertens: Indeed I will change them in future PRs! By the end of this entire ticket, all of them should use the children approach :) This is a necessary intermediate state to get to the end!

src/components/Sessions/SessionTrackerModal.jsx
sergei-maertens: ah, I think we had both flavours! I'll change it to yours, I have no opinions about it :)

src/components/Form.jsx
sergei-maertens: which one do you mean?

robinmolen: I think this was some confusion on my side.. And also some miss-communication in my comment.

I was thinking about the timeout issue on the confirmation and payment pages. About users who completes a form and postpone the last action (downloading the submission pdf or finalizing the payment) to a later time.
But this would mean users keep the form in a half-finished state for more that 10 hours.. Which is a bit weird/extreme..

This comment can be ignored, seeing how this (probably) isn't something that will ever happen

PR #775, opened 2025-01-16
[OF#5006] Update addressNL component to manually fill in city and streetname

vaszig:
Closes open-formulieren/open-forms#5006 partly

Until now the `addressNL` component would retrieve the city and street name based on the postcode and housenumber. These fields were always read-only but now we want to be able to fill in city and street name manually if something goes wrong with the API call.


src/formio/components/AddressNL.jsx
```diff
@@ -368,6 +382,7 @@ const AddressNLForm = ({initialValues, required, deriveAddress, layout, setFormi
 const FormikAddress = ({required, setFormioValues, deriveAddress, layout}) => {
   const {values, isValid, setFieldValue} = useFormikContext();
   const {baseUrl} = useContext(ConfigContext);
+  const [isAddressDisabled, setAddressDisabled] = useState(true);
```
robinmolen, line 385: I think that the name `isAddressDisabled` could cause some confusion. Maybe something like this describes the situation more clearly:

```suggestion
const [isAddressAutoFilled, setAddressAutoFilled] = useState(true);
```

src/formio/components/AddressNL.stories.js
```diff
@@ -246,6 +246,115 @@ export const WithDeriveCityStreetNameWithData = {
   },
 };
 
+export const WithDeriveCityStreetNameWithNoDataNotRequired = {
```
robinmolen, line 249: There already exists a `WithDeriveCityStreetNameNoData` test, which covers the same functionality.

Lets remove `WithDeriveCityStreetNameWithNoDataNotRequired`, and add the `not disabled` assertions to `WithDeriveCityStreetNameNoData`

src/formio/components/AddressNL.jsx
vaszig: Yes, you are right!It's much better.

src/formio/components/AddressNL.jsx
```diff
@@ -430,21 +451,23 @@ const FormikAddress = ({required, setFormioValues, deriveAddress, layout}) => {
...
               />
             }
-            disabled
+            disabled={isAddressAutoFilled}
+            isRequired={required}
           />
           <TextField
             name="city"
             label={
               <FormattedMessage
```
robinmolen, line 464: Maybe nice to use the `FIELD_LABELS.city` here, instead of a separate label (similar to the `PostCodeField` and `HouseNumberField`)

src/formio/components/AddressNL.jsx
```diff
@@ -430,21 +451,23 @@ const FormikAddress = ({required, setFormioValues, deriveAddress, layout}) => {
             name="streetName"
             label={
               <FormattedMessage
-                description="Label for addressNL streetName read only result"
+                description="Label for addressNL streetName result or entered value"
```
robinmolen, line 454: Same as the city input, maybe nice to use `FIELD_LABELS.streetName`

src/formio/components/AddressNL.jsx
vaszig: Maybe I should add all of them to the `FIELD_LABELS` then to be consistent (houseletter and housenumberaddition as well)?

robinmolen: Ah yeah, sounds good!

I just noticed a comment from Laurens on another AddressNL ticket https://github.com/open-formulieren/open-forms/issues/4699#issuecomment-2454224841. If you could change the dutch translation of `city` to `Plaats`, that would be amazing

src/formio/components/AddressNL.stories.js
```diff
@@ -84,6 +85,28 @@ export const ClientSideValidation = {
...
+    await sleep(500);
+    const canvas = within(canvasElement);
+
+    (await canvas.findByLabelText('Huisnummer')).focus();
+    await userEvent.tab();
+
+    expect(await canvas.findByText('Postcode is verplicht.')).toBeVisible();
+    expect(await canvas.findByText('Huisnummer is verplicht.')).toBeVisible();
+    expect(await canvas.findByText('required')).toBeVisible();
+  },
```
robinmolen, line 107: Can we make this a part of the `ClientSideValidation` story?

robinmolen (changes requested):
Some things can/should be simplified/cleaned-up.

I'm wondering if the extra logic in `checkComponentValidity` should be there. Maybe you can discuss this tomorrow with Sergei, or we can take a look next week

vaszig:
> Some things can/should be simplified/cleaned-up.
> 
> I'm wondering if the extra logic in `checkComponentValidity` should be there. Maybe you can discuss this tomorrow with Sergei, or we can take a look next week

Yes, I am not exactly sure about this either, but I chose to do it like this after some investigation in the code.
We already do it in several places like in the `FileField` component:

 https://github.com/open-formulieren/open-forms-sdk/blob/main/src/formio/components/FileField.js#L305.

src/formio/components/AddressNL.jsx
```diff
@@ -61,6 +61,50 @@ export default class AddressNL extends Field {
     if (this.component.validate.plugins && this.component.validate.plugins.length) {
       updatedOptions.async = true;
     }
+
+    if (!dirty) {
+      return super.checkComponentValidity(data, dirty, row, updatedOptions);
+    }
+
+    const {postcode, houseNumber, city, streetName} = this.dataValue;
+    if (this.component?.validate?.required) {
```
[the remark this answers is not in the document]
sergei-maertens: it should definitely be possible to re-use the zod schema by calling `const schema = addressNLSchema(...)`. We also have access to the `intl` object via `this.options.intl` (which is set [here](https://github.com/open-formulieren/open-forms-sdk/blob/49bd7cb09c3f6527b934b4d313dfb0aea3676766/src/components/FormStep/index.jsx#L793)). Calling [`safeParse`](https://zod.dev/?id=safeparse) on it will tell you whether it's valid or not.

W/r to displaying the errors in the right place, it would be great if we can again use the schema - Formik has an option `validateOnMount` and perhaps more interesting, a [`validateForm`](https://formik.org/docs/api/formik#validateform-values-any--promiseformikerrorsvalues) function you can call imperatively.

So it would make sense that if you detect at the Formio level that there are errors, to then call `validateForm` *somehow* to ensure that validation errors get displayed.

src/formio/components/AddressNL.stories.js
sergei-maertens: I prefer short (interaction) tests - you might also want to consider writing unit tests for this rather than doing it in stories (in `AddressNL.spec.jsx`)

src/formio/components/AddressNL.jsx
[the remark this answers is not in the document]
vaszig: This link does not work for me..

src/components/FormStep/FormStep.stories.jsx
```diff
@@ -422,3 +422,36 @@ export const SummaryProgressNotVisible = {
...
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+
+    const postcodeInput = await canvas.findByLabelText('Postcode');
+    const submitButton = await canvas.findByRole('button', {name: 'Next'});
+
+    await userEvent.type(postcodeInput, '1017 CJ');
+
+    // wait for the check logic api call
+    await sleep(1800);
```
vaszig, line 454: If I understood it correctly, the button is disabled for all this time, so I have to wait for it and then click on it.

PR #776, opened 2025-01-16
React router upgrade part 3

sergei-maertens:
Partly closes open-formulieren/open-forms#4929

This moves the remainder of the components to the statically defined routes.


src/components/PostCompletionViews/StatusUrlPoller.jsx
```diff
@@ -71,10 +72,9 @@ const StatusUrlPoller = ({statusUrl, onFailure, onConfirmed, children}) => {
 
   // FIXME: https://github.com/open-formulieren/open-forms/issues/3255
   // errors (bad gateway 502, for example) appear to result in infinite loading
-  // spinners
-  if (error) {
-    console.error(error);
-  }
+  // spinners. Throwing during rendering will at least make it bubble up to the nearest
+  // error boundary.
+  if (error) throw error;
```
robinmolen, line 79: This throw is not needed, right? Because the error already gets thrown by line 50

src/components/PostCompletionViews/StatusUrlPoller.jsx
```diff
@@ -71,10 +74,9 @@ const StatusUrlPoller = ({statusUrl, onFailure, onConfirmed, children}) => {
...
 
   const {
     result,
```
robinmolen, line 82: The `if (result === RESULT_FAILED)` on lines R91-R93 can now be removed, right?

Seeing how `onFailureNavigateTo` is required, and unhandled `result === RESULT_FAILED` should no-longer be possible

At the very least, the error message should probably be changed to `Failure should have been handled with the onFailureNavigateTo prop.`

src/components/PostCompletionViews/StatusUrlPoller.jsx
robinmolen, line 1: The `if (result === RESULT_FAILED)` on lines R91-R93 can now be removed, right?

Seeing how `onFailureNavigateTo` is required, and unhandled `result === RESULT_FAILED` should no-longer be possible

At the very least, the error message should be changed to `Failure should have been handled with the onFailureNavigateTo prop.`

src/components/PostCompletionViews/StatusUrlPoller.jsx
sergei-maertens: indeed - something didn't go right here in the refactor, good catch!

src/components/PostCompletionViews/StatusUrlPoller.jsx
sergei-maertens: nice catch - since the prop is required and the failure handling is localized, this is indeed obsolete now

src/components/Summary/SubmissionSummary.jsx
```diff
@@ -89,19 +78,7 @@ const SubmissionSummary = ({onConfirm}) => {
...
-  useTitle(pageTitle, form.name);
-
-  const getErrors = () => {
-    let errors = [];
-    const processingError = location.state?.errorMessage;
-    if (processingError) errors.push(processingError);
-    if (state.error) errors.push(state.error);
-    return errors;
-  };
+  const errorMessages = [location.state?.errorMessage, submitError].filter(Boolean);
```
robinmolen, line 89: Ah nice, i didn't know you could do `.filter(Boolean)` 👌

src/components/Form.spec.jsx
```diff
@@ -193,6 +193,48 @@ test('Submitting the form with failing background processing', async () => {
...
+  // confirm the submission and complete it
+  vi.useFakeTimers();
+  await user.click(screen.getByRole('button', {name: 'Confirm'}));
+  expect(await screen.findByRole('heading', {name: 'Processing...'})).toBeVisible();
+  const loader = await screen.findByRole('status');
+  vi.runOnlyPendingTimers();
+  vi.useRealTimers();
+  await waitForElementToBeRemoved(loader);
+
+  // due to the error we get redirected back to the summary page.
```
robinmolen, line 235: I think this comment is incorrect?

src/api-mocks/submissions.js
```diff
@@ -167,6 +167,6 @@ export const mockSubmissionProcessingStatusErrorGet = http.get(
     })
 );
 
-export const mockSubmissionPaymentStartGet = http.post(`${BASE_URL}payment/:uuid/demo/start`, () =>
-  HttpResponse.json({data: {method: 'get', action: 'https://example.com'}})
-);
+export const mockSubmissionPaymentStartPost = (
+  data = {type: 'get', url: 'https://example.com', data: {}}
+) => http.post(`${BASE_URL}payment/:uuid/demo/start`, () => HttpResponse.json(data));
```
robinmolen, line 172: Is there a reason to change `method` to `type`? If i understand it correctly, `type` is used for something else https://developer.mozilla.org/en-US/docs/Web/API/Response/type, https://github.com/mswjs/msw/blob/e3234fdce5a36f41c77719492e2eca0f77d7aaa6/src/core/HttpResponse.ts#L9.

If this doesn't clash, its fine. But this could result into different behavior in the tests vs real `http.post` actions.

`data` should probably also be changed to `body`

src/components/AbortButton/AbortButton.jsx
```diff
@@ -1,15 +1,14 @@
...
-import {AnalyticsToolsConfigContext} from 'Context';
 import {OFButton} from 'components/Button';
+import {useAnalyticsToolsConfig} from 'components/analytics/AnalyticsToolConfigProvider';
 import {buildGovMetricUrl} from 'components/analytics/utils';
 import useFormContext from 'hooks/useFormContext';
 
 const AbortButton = ({isAuthenticated, onDestroySession}) => {
   const intl = useIntl();
-  const analyticsToolsConfig = useContext(AnalyticsToolsConfigContext);
+  const analyticsToolsConfig = useAnalyticsToolsConfig;
```
robinmolen, line 11: Shouldn't this be:
```suggestion
  const analyticsToolsConfig = useAnalyticsToolsConfig();
```

src/api-mocks/submissions.js
sergei-maertens: the mock was completely wrong with what our API endpoint actually returns, I corrected this. This has nothing to do with web standards but everything with the API endpoint being mocked.

the `data` argument is the response data returned by our API, which we mock.

src/components/analytics/AnalyticsToolConfigProvider.jsx
```diff
@@ -0,0 +1,40 @@
...
+
+AnalyticsToolsConfigContext.displayName = 'AnalyticsToolsConfigContext';
+
+const AnalyticsToolsConfigProvider = ({children}) => {
+  const {locale} = useIntl();
+  const {baseUrl} = useContext(ConfigContext);
+
+  const {value} = useAsync(async () => {
+    return await get(`${baseUrl}analytics/analytics-tools-config-info`);
+  }, [locale]);
```
robinmolen, line 25: Should the dependency also not include the `baseUrl`?

src/components/FormProgressIndicator.jsx
```diff
@@ -0,0 +1,121 @@
...
+};
+
+/**
+ * Determine the 'step' title to render for the accessible mobile menu label.
+ * @param  {IntlShape} intl  The `useIntl` return value.
+ * @param  {String} pathname The pathname ('url') of the current location.
+ * @param  {Object} form     The Open Forms form instance being rendered.
+ * @return {String}          The (formatted) string for the step title/name.
+ */
+const getMobileStepTitle = (intl, pathname, form) => {
```
robinmolen, line 41: could we do this with a switch?

codecov.yml
robinmolen, line 1: Should we also ignore `"src/**/*.d.ts"` and `"src/**/*.stories.{js,jsx}"`?

vite.config.mts
```diff
@@ -145,6 +145,7 @@ export default defineConfig(({mode}) => ({
         'src/**/*.d.ts',
         'src/**/*.stories.{js,jsx,ts,tsx}',
         'src/api-mocks/*',
+        'src/**/mocks.{js,jsx}',
```
robinmolen, line 148: Should this exclude list also contain `"src/**/**.spec.js"`?

codecov.yml
sergei-maertens: I'm not entirely sure what's going on here - I had expected the vite.config.mts excludes to result in the api mocks being ignored, but these are visible in https://app.codecov.io/github/open-formulieren/open-forms-sdk/tree/main/src, however type definitions and stories are not reported here at all.

My suspicion is that some files get picked by by storybook tests and/or vitest and others not and it may also depend on the bundle entrypoint. Storybook test coverage is now handled by jest internally, while for our unittests we use vitest and it's all confusing. So I decided to only add the patterns for files that are visible on https://app.codecov.io/github/open-formulieren/open-forms-sdk/tree/main/src that shouldn't be tracked so that no false impression is created that they're excluded because of the codecov config (while it appears to be some other mechanism that I don't fully understand).

vite.config.mts
sergei-maertens: that's part of the default exclude already, it seems :-) .test.* and .spec.* are not visible on codecov

src/components/FormProgressIndicator.jsx
sergei-maertens: that was my initial goal too, but it can't really be done since the match returns a match record with particular structure (or is null), but it also needs to take the parent paths into account. We have an open TODO to use relative paths in these checks instead of absolute, which we haven't been able to deal with because the behaviour of react-router was weird with the splat routes.

On top of that, the `/stap/:step` pattern needs to be compiled by react-router and the path matching code is surprisingly complicated and does not translate well/easily to a switch, unfortunately.

src/components/analytics/AnalyticsToolConfigProvider.jsx
sergei-maertens: yes indeed - I'm surprised the linter didn't pick up on this :thinking:

src/components/FormProgressIndicator.jsx
robinmolen: Mhh okay, thats a bummer..

robinmolen (approved):
Looks great! Let's merge!

PR #777, opened 2025-01-20
Build and distribute ESM bundle (next to UMD/ESM package)

### LINKED ISSUE #76: Re-enable dynamic loading of i18n modules & put our bundle on a diet
The webpack config causes these to be loaded from a relative path to the page where the SDK is embedded, rather than relative to the `sdk.js` module.

A quick fix right now is to always bundle these files instead of loading them dynamically, but this will make the bundle fatter as we add more translations.

- [x] Create ESM bundle build
- [x] Dynamically load message catalogs (#781)
- [ ] Dynamically import leaflet/map features (should trim the main bundle with ~500kB)
- [ ] Dynamically import appointments code/routes.

sergei-maertens:
Partly closes #76

Vite/Rollup doesn't support chunking when bundling with UMD (which results in `window.OpenForms` being available), but when generating an ESM bundle this does work.

This adds a build/compile step which produces an experimental (un-optimized) ESM bundle to the build artifact so that our backend can start opting in to `<script type="module">...</script>` to initialize the SDK instead of using the UMD bundle. If that works as expected, I want to manage the chunks to avoid loading unused stuff (appointments, map...) unless they are actually used.

### DIFF
--- a/vite.config.mts
+++ b/vite.config.mts
@@ -16,7 +16,7 @@ import {packageRegexes} from './build/utils.mjs';
 declare global {
   namespace NodeJS {
     interface ProcessEnv {
-      BUILD_TARGET: 'umd' | 'esm';
+      BUILD_TARGET: 'umd' | 'esm' | 'esm-bundle';
     }
   }
 }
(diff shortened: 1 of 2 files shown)


vite.config.mts
```diff
@@ -100,9 +129,20 @@ export default defineConfig(({mode}) => ({
         "this\['Interpreter'\]": "window['Interpreter']",
       },
     }),
+    // plugin to ignore (S)CSS when bundling to ESM bundle target, since these have
+    // already been produced by the UMD build.
+    {
```
robinmolen, line 134: The UMD build is deprecated, right? So should the (S)CSS really be ignored when bundling?

Or at least, should a `@TODO` be added to make sure that when UMD build is removed, this plugin is also removed?

sergei-maertens: It's deprecated but I don't expect it to be removed any time soon - in particular people on old mobile devices may have browsers that don't support ESM and they use the UMD build. However, it might be good to let the ESM bundle generate the CSS and the UMD bundle can piggyback on that instead of the other way around. I'll amend the comment!

PR #778, opened 2025-01-21
[OF#5003] Update styling in addressNL component

vaszig:
Closes open-formulieren/open-forms#5003


vaszig:
I couldn't find a more elegant way to remove the asterisk from the component's label. If you think something else should be done please let me know.

robinmolen (approved):
Yeah looks good!

The temporary solution for the asterisk looks indeed fine for now 👍

PR #781, opened 2025-01-22
Lazy load translations

### LINKED ISSUE #76: Re-enable dynamic loading of i18n modules & put our bundle on a diet

sergei-maertens:
Partly closes #76

Verified by creating a static build and using it in our backend, switching between languages works as expected :tada:

### DIFF
--- a/src/i18n.jsx
+++ b/src/i18n.jsx
@@ -12,9 +12,6 @@ import {logError} from 'components/Errors/ErrorBoundary';
 import ErrorMessage from 'components/Errors/ErrorMessage';
 import Loader from 'components/Loader';
 
-import messagesEN from './i18n/compiled/en.json';
-import messagesNL from './i18n/compiled/nl.json';
-
 // ensure flatpickr locales are included in bundle
 flatpickr.l10ns.nl = Dutch;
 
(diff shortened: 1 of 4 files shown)


src/i18n.jsx
```diff
@@ -24,22 +21,32 @@ const setLanguage = langCode => {
...
+      break;
+    }
+    case 'en':
+    // in case (accidentally) a locale is set that we don't ship translations for yet,
+    // fall back to English.
+    // eslint-disable-next-line no-fallthrough
+    default: {
+      localeToLoad = 'en';
+    }
   }
```
sergei-maertens, line 38: this may look a bit over-engineered, but our backend *does* support things like `nl_NL` and `nl_BE` locales and doing things this way makes it future proof for languages we may add in the future that have regional variants.

robinmolen (approved):
Nice, looks good to me!

I was slightly concerned that `await import('./i18n/compiled/${localeToLoad}.json')` would load multiple times, but it doesn't. It seems to load only once, even after switching from NL to EN and back to NL

sergei-maertens:
> Nice, looks good to me!
> 
> I was slightly concerned that `await import('./i18n/compiled/${localeToLoad}.json')` in the `i18n.jsx` would load multiple times, but it doesn't. It seems to load only once, even after switching from NL to EN and back to NL

Indeed - browsers/bundlers are smart enough to cache that, luckily!

PR #784, opened 2025-01-24
Optimize bundle chunks

sergei-maertens:
WIP


src/sdk.jsx
```diff
@@ -212,4 +188,4 @@ class OpenForm {
 
 export default OpenForm;
 export {ANALYTICS_PROVIDERS} from 'hooks/usePageViews';
-export {VERSION, OpenForm, Formio, Templates, OFLibrary, OpenFormsModule};
```
sergei-maertens, line 215: These have never been documented as public API :)

robinmolen (changes requested):

src/sdk.jsx
sergei-maertens: They were added with the idea of "could be useful if people want to make further customizations", but to the best of our knowledge nobody every did this (and it was deliberately undocumented). It was definitely with the intent to expose it in the ESM npm package of the SDK, but that has been experimental/broken since it's initial inception (even now with Vite there are still things broken w/r to localization/missing assets).

src/routes/app.jsx
```diff
@@ -45,6 +35,11 @@ const routes = [
           </ErrorBoundary>
         ),
       },
+      // appointments splat
+      {
+        path: '*',
+        children: appointmentRoutes,
```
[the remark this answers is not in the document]
sergei-maertens: I did it deliberately to keep all the lazy imports to the `appoinments/index.js` in a single place. If something changes there, then you don't have to worry about accidentally forgetting another place to update the imports. The bundle optimizations are nice, but it's quite a fragile setup (I hope to be able to address this with an ESLint rule some day)

src/sdk.jsx
robinmolen: Check, okay yeah if there isn't active (known) usage, then i guess we can safely remove them for the export. 👍

robinmolen (approved):

PR #786, opened 2025-01-27
Don't render AddressNL component when it hidden

robinmolen:
Partly closes: open-formulieren/open-forms#4699

The AddressNL component should only be rendered when it's not hidden. Otherwise, the attach function will call `createRoot` with an argument that isn't a DOM element. (Which causes an error being thrown in the front-end..)

So the creating of the root and, therefor, the rendering of the React component cannot be done. If the reactRoot doesn't exist, then it also doesn't (and cannot) be unmounted. So this is also something that shouldn't happen when the component is hidden.

This works fine when the component is hidden/shown using form logic, as the `hidden` property will be updated.

### DIFF
--- a/src/formio/components/AddressNL.jsx
+++ b/src/formio/components/AddressNL.jsx
@@ -100,14 +100,16 @@ export default class AddressNL extends Field {
       addressNLContainer: 'single',
     });
     return super.attach(element).then(() => {
-      this.reactRoot = createRoot(this.refs.addressNLContainer);
-      this.renderReact();
+      if (!this.component?.hidden) {
+        this.reactRoot = createRoot(this.refs.addressNLContainer);
+        this.renderReact();
+      }
     });
   }
 
   destroy() {
     const container = this.refs.addressNLContainer;
-    if (container) this.reactRoot.unmount();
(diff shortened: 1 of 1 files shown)


vaszig (commented):
- I a bit confused..which part of the ticket is this solving?

- I see here we have a css rule for hidden components but I guess this is not useful in this situation. And where do we handle the "hidden" components and prevent them from being rendered?

https://github.com/open-formulieren/open-forms-sdk/blob/main/src/scss/components/_formio-component.scss#L39

https://github.com/open-formulieren/open-forms-sdk/blob/main/src/formio/components/Component.js

vaszig:
> * I a bit confused..which part of the ticket is this solving?
> 
>     * I see here we have a css rule for hidden components but I guess this is not useful in this situation. And where do we handle the "hidden" components and prevent them from being rendered?
> 
> 
> https://github.com/open-formulieren/open-forms-sdk/blob/main/src/scss/components/_formio-component.scss#L39
> 
> https://github.com/open-formulieren/open-forms-sdk/blob/main/src/formio/components/Component.js

Discussed these remarks/questions in person with Robin

PR #791, opened 2025-01-30
Missing map shapes in summary step

robinmolen:
Part of open-formulieren/open-forms#5038

Because the geoJsonGeometry is available before the featureGroupRef is set, can the useEffect hook not draw the map shapes.

This fix makes sure that featureGroupRef is set, so that map shapes can be drawn

### DIFF
--- a/src/components/Map/LeafletMap.jsx
+++ b/src/components/Map/LeafletMap.jsx
@@ -19,6 +19,7 @@ import {
   leafletGestureHandlingText,
   searchControlMessages,
 } from './translations';
+import {GeoJsonGeometry} from './types';
 
 // Run some Leaflet-specific patches...
 initialize();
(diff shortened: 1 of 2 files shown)


src/components/Map/LeafletMap.jsx
```diff
@@ -92,17 +92,27 @@ const LeaftletMap = ({
...
+      if (!geoJsonGeometry) {
+        return;
+      }
+      // Add the current `geoJsonGeometry` as shape
+      const layer = Leaflet.GeoJSON.geometryToLayer(geoJsonGeometry);
+      featureGroupRef.current.addLayer(layer);
+    },
+    // `featureGroupRef.current` is needed as dependency to make sure that
+    // the featureGroupRef can be used.
+    [geoJsonGeometry, featureGroupRef.current]
```
sergei-maertens, line 114: Can you check these linter warnings?

robinmolen: I've changed the setup a bit to 'solve' this issue.

The problem was that the featureGroupRef was set after this useEffect was be called. Resulting in `featureGroupRef.current` not being defined/targetable. By adding the `featureGroupRef.currect` to the useEffect dependencies, featureGroupRef would trigger the useEffect once when it was defined.
But React doesn't like `ref.currect` as dependency, resulting in warnings/errors.

So i've changed the setup by creating a React component that handles the featureGroup updating. With this setup `featureGroupRef.currect` isn't needed as a dependency, because now `featureGroupRef.current` is set before we use it. This also separates the featureGroup updating functionality from the other map functionality.

PR #792, opened 2025-02-04
Display backend validation errors when submitting the form

sergei-maertens:
Closes open-formulieren/open-forms#4510 - requires the backend PR to be merged as that changes the API endpoint used here.

I've opted to collect all the errors at the top and group them by step, rather than trying to weave this in the summary table. User research from Utrecht showed that this was the best way to show validation errors.

Ideally, there would be some aria-describedby options, but that requires a lot more work to link everything together.


vaszig (approved):
Nothing to add/mention!

PR #793, opened 2025-02-06
Remove login buttons when the submission limit is reached

viktorvanwijk:
Closes open-formulieren/open-forms#5046

### DIFF
--- a/src/components/FormStart/index.jsx
+++ b/src/components/FormStart/index.jsx
@@ -56,6 +56,9 @@ const FormStart = () => {
 
   const hasActiveSubmission = !!submission;
   const isAuthenticated = hasActiveSubmission && submission.isAuthenticated;
+  // If there is no active submission, only show the login options if the submission limit has not
+  // been reached.
+  const showLoginOptions = !hasActiveSubmission && !form.submissionLimitReached;
 
   const doStart = useStartSubmission();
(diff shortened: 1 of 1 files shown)


viktorvanwijk:
Not exactly sure how the coverage has decreased from this change. Any ideas?

vaszig (approved):
I don't know either what's wrong with the coverage.

robinmolen (changes requested):
Could you place this logic outside the conditional rendering? Chaining if statements is something we should try to avoid. I think this also makes it a bit more clear why things are happening

Maybe something like:
```
const showStartButton = !hasActiveSubmission && !form.submissionLimitReached;

...

       {hasActiveSubmission && (
          <ExistingSubmissionOptions
            form={form}
            onDestroySession={onDestroySession}
            isAuthenticated={isAuthenticated}
          />
        )}

        {showStartButton && (
          <LoginOptions
            // if cosign allows links in emails, we don't need to display the cosign
            // login options, so strip them out
            form={form.cosignHasLinkInEmail ? {...form, cosignLoginOptions: []} : form}
            onFormStart={onFormStart}
            extraNextParams={extraNextParams}
          />
        )}
```

robinmolen (approved):

sergei-maertens (approved):

PR #795, opened 2025-02-17
[OF#5086] Fix showing soft requirement warning in hidden components

vaszig:
Fixes open-formulieren/open-forms#5086

### DIFF
--- a/src/formio/components/SoftRequiredErrors.js
+++ b/src/formio/components/SoftRequiredErrors.js
@@ -30,10 +30,12 @@ class SoftRequiredErrors extends FormioContentField {
     });
 
     const missingFieldLabels = [];
-    // check which components have an empty value
+    // check which components are visible and have an empty value
     for (const component of softRequiredComponents) {
       const isEmpty = component.isEmpty();
-      if (isEmpty) missingFieldLabels.push(component.label);
+      const isComponentVisible = component.visible;
+
+      if (isEmpty && isComponentVisible) missingFieldLabels.push(component.label);
     }
 
     if (!missingFieldLabels.length) return '';
--- a/src/formio/components/SoftRequiredErrors.stories.js
+++ b/src/formio/components/SoftRequiredErrors.stories.js
@@ -109,3 +109,192 @@ export const FillField = {
     });
   },
 };
+
+export const WithFieldSetAndNestedComponent = {
+  name: 'With fieldset and nested component',
+  args: {
+    components: [
+      {
+        type: 'fieldset',
+        key: 'fieldset',
+        label: "Auto's",
+        groupLabel: 'Auto',
+        components: [
+          {
+            type: 'file',
+            key: 'file',
+            label: 'Soft required upload',
+            openForms: {softRequired: true},
+          },
+        ],
+      },
+
+      {
+        type: 'softRequiredErrors',
+        html: `
+        <p>Not all required fields are filled out. That can get expensive!</p>
+
+        {{ missingFields }}
+
+        <p>Are you sure you want to continue?</p>
+          `,
+      },
+    ],
+  },
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+
+    // needed for formio
+    await sleep(100);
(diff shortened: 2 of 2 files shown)


src/formio/components/EditGrid.stories.jsx
```diff
@@ -357,3 +357,101 @@ export const WithLayoutComponents = {
...
+    const addButton = await canvas.findByRole('button', {name: 'Add Another'});
+    await userEvent.click(addButton);
+
+    const saveButton = await canvas.findByRole('button', {name: 'Save'});
+    await userEvent.click(saveButton);
+
+    await canvas.findByText('Not all required fields are filled out. That can get expensive!');
+    const list = await canvas.findByRole('list', {name: 'Empty fields'});
+    const listItem = within(list).getByRole('listitem');
+    expect(listItem.textContent).toEqual('Upload');
```
viktorvanwijk, line 410: Because there is a check on the label, would it be good to include this (custom) label in the component properties, instead of relying on the default value?

src/formio/components/EditGrid.stories.jsx
```diff
@@ -357,3 +357,101 @@ export const WithLayoutComponents = {
...
+  },
+};
+
+export const WithSoftRequiredComponentHiddenEditGrid = {
+  name: 'With soft required component hiedden editgrid',
```
viktorvanwijk, line 415: Typo in 'hiedden' :)

src/formio/components/SoftRequiredErrors.js
```diff
@@ -33,7 +33,11 @@ class SoftRequiredErrors extends FormioContentField {
     // check which components have an empty value
     for (const component of softRequiredComponents) {
       const isEmpty = component.isEmpty();
-      if (isEmpty) missingFieldLabels.push(component.label);
+      const isParentHidden = !component.parent?.visible;
+      const isComponentHidden = !component.visible;
+
+      if (isEmpty && !isComponentHidden && !isParentHidden)
+        missingFieldLabels.push(component.label);
```
viktorvanwijk, line 40: Perhaps change the name to `isParentVisible` and `isComponentVisible`, and get rid of the double negation?

viktorvanwijk (changes requested):
Few minor remarks/questions :)

viktorvanwijk (approved):

src/formio/components/EditGrid.stories.jsx
sergei-maertens, line 1: I think it makes more sense to define these stories/tests in https://github.com/open-formulieren/open-forms-sdk/blob/main/src/formio/components/SoftRequiredErrors.stories.js as it has to do with the soft required component more than the edit grid, no? Same for the fieldset stories

src/formio/components/SoftRequiredErrors.js
```diff
@@ -33,7 +33,11 @@ class SoftRequiredErrors extends FormioContentField {
     // check which components have an empty value
     for (const component of softRequiredComponents) {
       const isEmpty = component.isEmpty();
-      if (isEmpty) missingFieldLabels.push(component.label);
+      const isParentVisible = component.parent?.visible;
```
sergei-maertens, line 36: what it the grand parent is hidden?

vaszig: Yes, you are right I didn't think about it.

vaszig: After some research I saw that even if we have nested components this works. If, as you mentioned, the grand parent is hidden, all the children are hidden as well. I understood that formio (the UI) does not render the children of a hidden component, even if the json configuration has `hidden=false`.

sergei-maertens: That's correct - I just didn't know if `component.visible` took into account is parents or not :)

sergei-maertens (approved):

sergei-maertens:
Backported to 3.0.x in 29ccec81c4b7c440f28e61075ededb91843039ac

PR #804, opened 2025-03-06
Convert some low-hanging fruit to TypeScript

### LINKED ISSUE #445: Adopt typescript
Check if we can start gradual typescript adoption, normally CRA should support it out of the box?

Steps:

- [x] Use @open-formulieren/formio-renderer where possible to reduce the amount of code that needs converting
- [x] Add typescript dependency and config file(s)
- [x] Upgrade dependencies
- [x] Convert low-hanging fruit
- [x] Convert entrypoint and publish type declaration files

sergei-maertens:
Partly closes #445


src/api.ts
```diff
@@ -12,21 +12,29 @@ import {
...
 import {setLanguage} from './i18n';
 
-const fetchDefaults = {
-  credentials: 'include', // required for Firefox 60, which is used in werkplekken
+interface ApiCallOptions extends Omit<RequestInit, 'headers'> {
+  headers?: Record<string, string>;
+}
+
+const fetchDefaults: ApiCallOptions = {
+  credentials: 'include',
```
robinmolen, line 20: Maybe keep the `// required for Firefox 60, which is used in werkplekken` comment? So that we don't accidentally remove/change it

src/api.ts
robinmolen, line 1: Do we want to include return types? In these cases it will be a lot of `void` and `Promise<Record<string, any> | null>` i think, but still maybe nice to add it?


(If we want to be fancy, we could even add generic types to the `get`, `post`, `put`, to ensure/hint that type X is returned)

src/components/AppDebug.tsx
```diff
@@ -1,20 +1,25 @@
...
+import {getVersion} from '@/utils';
 
-const DebugInfo = ({label, value, children}) => (
+export interface DebugInfoProps {
+  label: string;
+  value?: string;
+  children?: React.ReactNode;
+}
+
+const DebugInfo: React.FC<DebugInfoProps> = ({label, value, children}) => (
```
robinmolen, line 13: I'll create a ticket for this, but i think it would be best to remove `value` and make `children` required (seeing how it's only used in the `AppDebug`, and `value` and `children` are treated the same way (and `React.ReactNode` includes `string`, so we even don't have to change the typing))

robinmolen (changes requested):
Some comments are questions if we want to be strict/specific with the typing. Just take a look and let me know you're thoughts

src/utils.ts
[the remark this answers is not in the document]
sergei-maertens: Yes, I'm definitely in favour of explicit return types :-)

src/components/Price.tsx
```diff
@@ -1,17 +1,20 @@
...
+const Price: React.FC<PriceProps> = ({price = '0'}) => {
   return (
     <div className={getBEMClassName('price')}>
       <div className={getBEMClassName('price__label')}>
         <FormattedMessage description="Label for the total price to pay" defaultMessage="Total" />:
       </div>
       <div className={getBEMClassName('price__amount')}>
         <FormattedNumber
-          value={price}
+          value={typeof price === 'number' ? price : parseFloat(price)}
```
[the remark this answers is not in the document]
sergei-maertens: The default value of `'0'` for price ensures that it's never `undefined` by the time it gets passed to `FormattedNumber`. Removing the default correctly gives a type error, so working as intended I'd say.

src/api.ts
sergei-maertens: Firefox 60 was released in May 2018, which is almost seven years ago. This is code that was copied between like 6-7 different projects now with the left over comment - initially we didn't specify it but had to do so because old browser versions were used within the municipalities.

The default value is `same-origin`, which was sufficient in those older projects as there was no CORS involved (this is different for the SDK), but most notably, this default only came into effect in Firefox 61 (see [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials)), hence we needed to explicitly state it.

With the default being `same-origin` which is not sufficient for the SDK (as it does CORS) and requires it to be set to `include`, the original comment is actually not relevant at all anymore.

src/api.ts
sergei-maertens: yeah I didn't want to go all out yet with generics, but ultimately that's what will be needed for the API calls. I can set that up already if you don't think it will clutter too much :)

I will add the other return types!

src/components/AppDebug.tsx
sergei-maertens: The thought also occurred to me and I wasn't sure whether the refactor immediately or leave that for another day. If you don't mind, I can refactor it in this PR.

src/components/Price.tsx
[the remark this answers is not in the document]
robinmolen: Ahhh, i'm blind... I was looking at the typescript definition, where `price` is optional. I didn't see the default value...

Okay, yeah all good 😄

src/api.ts
robinmolen: Yeah i think we can add these generics quite safely, just so this this file is "done". If it makes no difference for you, then lets just add them 🙂 (for me it's also okay to do this later. Just making sure we don't "forget" this 🙂)

src/components/AppDebug.tsx
robinmolen: Yeah, lets do it!

src/api.ts
robinmolen: Ah check, yeah then the comment doesn't really add anything 👍  (it might even under-sell the function/use)

src/components/AppDebug.tsx
sergei-maertens: done!

src/components/AppDebug.tsx
```diff
@@ -6,14 +6,13 @@ import {getVersion} from '@/utils';
 
 export interface DebugInfoProps {
   label: string;
-  value?: string;
   children?: React.ReactNode;
```
robinmolen, line 9: Sorry, a bit of nitpicking...

Should children be required? The three instances that we have all return a value, and i don't think this would be very useful without some sort of value

Or do we maybe add a default value? Like how `src/components/AppDebug.jsx:25` defaults to `'-'`

robinmolen (approved):
I've added one final comment, but it's not very important. It can be discussed/addressed or you can merge it, both are fine 😄

PR #805, opened 2025-03-11
:label: [#445] Add TS type definitions for Form/Submission

sergei-maertens:
This adds some of the type definitions for Form and Submission resources that we interact with from the v2 API endpoint(s), set up to strongly type the createSubmission function.


src/data/forms.ts
robinmolen, line 1: Could we use something like https://www.npmjs.com/package/openapi-typescript for the API spec types? Otherwise i think updating the types will be a pain (and error-prone)..

src/data/submissions.ts
```diff
@@ -0,0 +1,93 @@
...
+  anonymous = false
+): Promise<Submission> => {
+  const createData: SubmissionCreateBody = {
+    form: form.url,
+    formUrl,
+    anonymous,
+    initialDataReference: initialDataReference ? initialDataReference : undefined,
+  };
+  const submissionResponse = await post<Submission>(`${baseUrl}submissions`, createData, signal);
+  return submissionResponse.data!;
```
robinmolen, line 74:
```suggestion
  return submissionResponse.data;
```

src/data/submissions.ts
```diff
@@ -0,0 +1,93 @@
...
+  formUrl: string,
+  signal: AbortSignal,
+  initialDataReference: string,
+  anonymous = false
+): Promise<Submission> => {
+  const createData: SubmissionCreateBody = {
+    form: form.url,
+    formUrl,
+    anonymous,
+    initialDataReference: initialDataReference ? initialDataReference : undefined,
```
robinmolen, line 71: Is it important to differentiate between empty string and `undefined`? Otherwise we could just do:
```suggestion
    initialDataReference,
```

robinmolen: Because, when fetch the submission, it is returned with a empty string (if undefined) right? (Looking at the `test_retrieve_submission_nothing_submitted` backend test)

robinmolen (changes requested):

src/data/forms.ts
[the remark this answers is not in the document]
sergei-maertens: `.d.ts` files are generated by TSC, `.ts` is for source code and even if it only contains types, it's just source code for the compiler. https://www.reddit.com/r/typescript/comments/jxadan/should_i_use_dts_for_my_types/

sergei-maertens: I'm leaving out the bits that are not relevant for the SDK and are only used in the backend, otherwise it's confusing what the SDK needs to support and what is irrelevant. This is to declutter. So, assuming that openapi-typescript or similar supports this, it will still be maintenance heavy because you need to keep the allowlist in sync.

Honestly, if the API spec changes and you use something, it will break at runtime and should be caught by tests/storybook. If the API is expanded and you try to use it, the typescript compiler will complain about the type not being defined (correctly) and then you expand the type definitions.

I don't think there will much overhead from updating the types tbh.

src/data/submissions.ts
sergei-maertens: If the HTTP response is 204, then it can return `null` and the type checker correctly warns about the potential return type being `null`. However, we know that it's a create operation that will result in either an error or HTTP 201 and thus we can tell the compiler that it will not be `null`, which the exclamation mark is for.

src/data/submissions.ts
sergei-maertens: I'm not sure and I opted for the defensive approach - the current version in `main` does not send the parameter at all to the backend if no reference is provided, and that's replicated with setting it to `undefined`. I don't know what will happen if we send an empty string.

src/data/forms.ts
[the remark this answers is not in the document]
robinmolen: Yeah I saw similar messages. With previous projects i typically had `.ts` for component specific types, and `.d.ts` for global stuff like objects that are used in more places.

But just `.ts` is fine for me 👍

robinmolen: Okay, I prefer knowing (and having types for) entire objects. Just to have a clear overview of what i can expect. But I get wanting to keep it contained to just SDK specific stuff...

I've worked with one of such packages (not 100% sure if it was openapi-typescript, but it sounds very familiar), but there we could specify which object schemas to use when generating the types (which should help filtering to the information that is needed/interesting for the SDK).

And yeah, we will still have to update the SDK when the API spec changes, but this would cut back on the amount of double manual work..

But we can continue like this for now

src/data/submissions.ts
robinmolen: Ahh okay, i didn't recognize the notation, seemed like a typo 😄.

Okay, cool 👍

src/data/submissions.ts
robinmolen: ah okay, yeah being save/explicit can't hurt. Okay, lets keep it like this

PR #806, opened 2025-03-12
Convert Card component to Typescript

sergei-maertens:
... and cleaned up its usage along the way.

The CardTitle component is used in either one of two ways:

* implicitly, through the Card where it's basically _the_ title and should be displayed as h1 element
* explicitly, in which case it was always being rendered as an h2 element

So, the default heading level is now 2 for the low-level component, while using it implicitly through Card ensures that it's level 1, matching the current behaviour but cleaning up the code quite a bit. This cleanup removes the need for polymorphic component props to support multiple HTML elements - our codebase does not use anythign other than heading elements at the moment (and those landmarks are important for a11y).

Next, the Card component itself. The modifiers list is gone in favour of explicit (boolean) props that set/determine the modifiers under the hood, allowing proper combination and usage tracking while being able to document their behaviour individually. Usages have been updated accordingly, but let's wait for Chromatic to confirm nothing broke.

The caption did not appear to be used anywhere as Card prop, so that's been removed.


src/components/ProgressIndicator/index.jsx
```diff
@@ -23,11 +23,6 @@ const ProgressIndicator = ({
   const [resizeCounter, setResizeCounter] = useState(0);
   const buttonRef = useRef(null);
 
-  const modifiers = [];
-  if (expanded) {
-    modifiers.push('expanded');
-  }
```
sergei-maertens, line 29: Couldn't find any CSS anymore that still did something with it, I reckon we missed this when we rewrote the progress indicator.

robinmolen (changes requested):

robinmolen (approved):

PR #807, opened 2025-03-18
:art: Use button-group-css isolated package

sergei-maertens:
The @utrecht/button-group-react package pulls in the button-group-css dependency which keeps implementation + styles in sync. So, we can use the new CSS package in favour of the (outdated and no longer published) @utrecht/components package.


vaszig (approved):
No changes in the UI tests so good to go I guess!

PR #809, opened 2025-03-21
[open-formulieren/open-forms#5155] Add initial data reference to `OpenForm`

viktorvanwijk:
Partly closes open-formulieren/open-forms#5155

Previously, the initial data reference would be lost when switching language during filling out a form. This is because, as of [open-formulieren/open-forms#4600], we reload the entire page on a language switch to update all links.

Now, the initial data reference is passed to the `onLanguageChange` callback, so it can be added again.


src/sdk.jsx
```diff
@@ -128,6 +130,11 @@ class OpenForm {
     ).href;
   }
 
+  getInitialDataReference() {
```
sergei-maertens, line 133:
```suggestion
  extractInitialDataReference() {
```
if I see a method `foo.getInThereLewis()`, I expect to get the thing when I call, not a side-effect that sets it on the instance.

So I'd go for `extract` or `store` or something like that

viktorvanwijk: This method was written for Lewis in a Ferrari; you expect to get something but actually end up with nothing

sergei-maertens (approved):

sergei-maertens:
Backport only to 3.0.x (994a0a13e7fbb6454a94d4e28b3f32b390ec455e) - this feature didn't exist before then.

PR #811, opened 2025-03-24
Add typescript files to package.json scripts

robinmolen:
The typescript files (`*.ts` and `*.tsx`) were missing from the `makemessages-nl`, `makemessages-en`, resulting in translations being removed while the corresponding `formatMessage(...)` functions remained.

To be save i've also added the typescript file extensions to the `format` and `checkformat` scripts.


sergei-maertens (approved):
Oof, painful. How bad is the damage?

robinmolen:
> Oof, painful. How bad is the damage?

Oh not too bad. For the release prep I was running the makemessages and was just making sure all deleted translations should indeed be deleted. Spoiler alert: they weren't 😄

PR #813, opened 2025-03-25
More TS conversion

sergei-maertens:


src/components/Errors/FormMaximumSubmissionsError.tsx
```diff
@@ -0,0 +1,14 @@
+import {FormattedMessage} from 'react-intl';
+
+import ErrorMessage from '@/components/Errors/ErrorMessage';
+
+const FormMaximumSubmissionsError = () => (
```
viktorvanwijk, line 5: Do we want an explicit `React.FC` type here, or is it ok like this if there are no props?

src/components/Errors/types.ts
```diff
@@ -0,0 +1,4 @@
+import {APIError} from '@/errors';
+
+// you can pretty much throw anything in JS
```
viktorvanwijk, line 3: Interesting :thinking:

src/hooks/useScrollIntoView.ts
```diff
@@ -1,7 +1,9 @@
 import {useEffect, useRef} from 'react';
 
-const useScrollIntoView = (options = {behavior: 'smooth'}) => {
-  const ref = useRef();
+const useScrollIntoView = <T extends HTMLElement = HTMLElement>(
+  options: ScrollIntoViewOptions = {behavior: 'smooth'}
+) => {
```
viktorvanwijk, line 5: Do we want implicit or explicit return types?

src/hooks/useScrollIntoView.ts
```diff
@@ -1,7 +1,9 @@
 import {useEffect, useRef} from 'react';
 
-const useScrollIntoView = (options = {behavior: 'smooth'}) => {
-  const ref = useRef();
+const useScrollIntoView = <T extends HTMLElement = HTMLElement>(
```
viktorvanwijk, line 3: What is the purpose of adding `= HTMLElement`?

src/components/Errors/ErrorBoundary.stories.tsx
```diff
@@ -0,0 +1,159 @@
+import {Meta, StoryObj} from '@storybook/react';
+import {MemoryRouter} from 'react-router';
+
+import {PermissionDenied, ServiceUnavailable, UnprocessableEntity} from '@/errors';
+
+import ErrorBoundary from './ErrorBoundary';
+
+// in JS, you actually *can* throw anything
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const Throw: React.FC<{error: any}> = ({error}) => {
```
viktorvanwijk, line 10: I see you've defined an `AnyError` type. Can we use that here as well?

viktorvanwijk (changes requested):
A few minor things, and some questions for my understanding :). I guess a general question is if we want implicit or explicit return types?

sergei-maertens:
> A few minor things, and some questions for my understanding :). I guess a general question is if we want implicit or explicit return types?

Earlier with Robin we settled on explicit return types :)

src/components/Errors/types.ts
sergei-maertens: ThisIsFine

src/components/Errors/ErrorBoundary.stories.tsx
sergei-maertens: done!

PR #815, opened 2025-03-27
Ensure anonymous form start button is not displayed for cosign start

sergei-maertens:
Closes open-formulieren/open-forms#5195

Cosign always requires users to authenticate and whether the form can be started anonymously or not has no influence on the cosign start options.


vaszig (approved):

sergei-maertens:
Backported to 3.0.x in a1aeb98a9a66b0ef8b955b9e032c874cc01b4325

PR #818, opened 2025-04-08
Dropping support for browser translations

robinmolen:
Closes: open-formulieren/open-forms#5242

Browser translations, in participle the translation functions of Chrome, change the DOM tree when translating texts. This causes React to lose track of certain DOM nodes, which leads to the SDK throwing errors and crashing.

This solution ensures that no SDK content can be translated. With this we lose some accessibility, but this seems to be the only reliable solution that would "solve" the issue for any and all browsers and translation plugins.

### DIFF
--- a/src/sdk.jsx
+++ b/src/sdk.jsx
@@ -138,6 +138,11 @@ class OpenForm {
   async init() {
     ReactModal.setAppElement(this.targetNode);
 
+    // Fixing an issue where browser (in particular Chrome) translations change the DOM
+    // tree, causing React to lose track of DOM nodes and crashing the SDK.
+    // See https://github.com/open-formulieren/open-forms/issues/5242
+    this.targetNode.setAttribute('translate', 'no');
+
     this.url = `${this.baseUrl}forms/${this.formId}`;
     this.targetNode.textContent = `Loading form...`;
     this.baseTitle = document.title;


robinmolen:
I don't think we can cover this with tests, as we (i'm pretty sure) cannot trigger a language switch with JS. Perhaps we can do this via vitest `user.click` actions, but this would then require multiple testing approaches for the different browsers we use.

I can take a deeper look if that is preferred.

src/sdk.jsx
```diff
@@ -138,6 +138,11 @@ class OpenForm {
   async init() {
     ReactModal.setAppElement(this.targetNode);
 
+    // Fixing an issue where browser (in participle Chrome) translations change the DOM
```
viktorvanwijk, line 141: Do you mean 'in particular'?

viktorvanwijk (approved):

sergei-maertens:
> I don't think we can cover this with tests, as we (i'm pretty sure) cannot trigger a language switch with JS. Perhaps we can do this via vitest `user.click` actions, but this would then require multiple testing approaches for the different browsers we use.
> 
> I can take a deeper look if that is preferred.

agree that we can't test this, this is entirely browser-addon behaviour that we cannot pull in scope for our project.

src/sdk.jsx
robinmolen: Ah haha, yeah... I'll update it 😅

PR #821, opened 2025-04-09
Rework map delete button functionality

robinmolen:
Partly closes open-formulieren/open-forms#5191

The delete button will no-longer show the delete-menu, instead it will just immediately remove the shape from the map. With that, I've also removed the translations for this delete-menu, as it no-longer will be shown.

### Extra context/thought process
The comment on open-formulieren/open-forms#5191 suggests relabeling the "save" (confirm delete action) button. While working on that, i took a look if could do anything more to make the deletion more expected/understandable. Unfortunately we cannot control what the sub-menu items do, or which sub-menu items are shown, so that idea quickly died...

When reading the open-formulieren/open-forms#5191 comments again, I thought of another approach: what if the delete button just deletes the drawn shape? This solution seemed the simplest, for us as developers but also for end-users, so thats what this PR contains.

### DIFF
--- a/src/components/Map/LeafletControl.tsx
+++ b/src/components/Map/LeafletControl.tsx
@@ -1,0 +1,99 @@
+import clsx from 'clsx';
+import L from 'leaflet';
+import {createRef, useEffect, useState} from 'react';
+import {useMap} from 'react-leaflet';
+
+interface ControlProps {
+  position: L.ControlPosition;
+  children: React.ReactNode;
+}
+
+const POSITION_CLASSES = {
+  bottomleft: 'leaflet-bottom leaflet-left',
+  bottomright: 'leaflet-bottom leaflet-right',
+  topleft: 'leaflet-top leaflet-left',
+  topright: 'leaflet-top leaflet-right',
+};
+
+const LeafletControl: React.FC<ControlProps> = ({position, children}) => {
+  const [portalRoot, setPortalRoot] = useState<HTMLElement>(document.createElement('div'));
+  const controlContainerRef = createRef<HTMLDivElement>();
+  const map = useMap();
+
+  /**
+   * Whenever the control container ref is created, ensure the click/scroll propagation
+   * is removed. This way click/scroll events do not bubble down to the map.
+   */
+  useEffect(() => {
+    if (controlContainerRef.current !== null) {
+      L.DomEvent.disableClickPropagation(controlContainerRef.current);
+      L.DomEvent.disableScrollPropagation(controlContainerRef.current);
+    }
+  }, [controlContainerRef]);
+
+  /**
+   * Mount the component to the position container of the map component.
+   */
+  useEffect(() => {
+    const mapContainer = map.getContainer();
+    const targetDiv = mapContainer.getElementsByClassName(POSITION_CLASSES[position]);
+    setPortalRoot(targetDiv[0] as HTMLElement);
+  }, [map, position]);
+
+  /**
+   * Whenever the portal root is complete, append the control container to the portal
+   * root.
+   */
+  useEffect(() => {
+    if (portalRoot && controlContainerRef.current) {
+      portalRoot.append(controlContainerRef.current);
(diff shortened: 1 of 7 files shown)


sergei-maertens (commented):
Yeah that's a much better approach!

To prevent oopsies, it's probably wise to gate it behind a `window.confirm` to ask the user if they really want to remove their hard artistic labour :grimacing: I can already see myself accidentally tapping that button on a touch screen.

robinmolen:
> Yeah that's a much better approach!
> 
> To prevent oopsies, it's probably wise to gate it behind a `window.confirm` to ask the user if they really want to remove their hard artistic labour 😬 I can already see myself accidentally tapping that button on a touch screen.

Ah yeah, for mobile that will be a definite issue... Will do 👍

robinmolen:
Currently this gives issues with React-leaflet-draw. Due to a bug the event handlers aren't unmounted correctly, resulting in the callback being executed multiple times https://github.com/alex3165/react-leaflet-draw/issues/192.

I've opened a PR to React-leaflet-draw which should fix the issue. While waiting for the project to be fixed, I'll do a temporary solution with patch-package

robinmolen:
Update: using the React-leaflet-draw `EditControl` delete handling didn't nicely fit together with our needs. When a user would decline the new confirmation message, the delete toolbar (sub-menu below the delete button) could not automatically be closed. Due to the setup of Leaflet, you cannot interact with the controllers via the map object.

So after some trial and error I decided to remove the React-leaflet-draw `EditControl` delete handling and add a custom Leaflet control. (I think that the setup of this control is also very interesting for the new "use current location" feature https://github.com/open-formulieren/open-forms-sdk/pull/819)

src/components/Map/LeafletControl.tsx
sergei-maertens, line 1: I was looking at this code and wondering if this was taken from react-leaflet as it seems fairly generic, but couldn't find it as is. However, I did look again at https://react-leaflet.js.org/docs/v4/core-architecture/ and I think there's quite some utilities in there (see https://react-leaflet.js.org/docs/v4/core-api/#high-level-component-factories) that might help here instead of having the implement the whole component yourself?

robinmolen: I've opened a new branch to handle this functionality (this branch is wildly outdated, and a fresh started seemed like the right approach). I've tried a couple of approaches, and landed on a rather elegant solution.

First of, I've tried the `createControlComponent`, which seems the right tool for the job. But had some difficulty making the UI consistent (on initial render the delete button would always appear before the "draw shape X" buttons, and after state changes the button would appear below the "draw shape X" buttons).

I've also tried mounting the custom control directly onto the map:
```
import {useMap} from 'react-leaflet';

const LeafletMapDeleteControl: React.FC<LeafletMapDeleteControlProps> = () => {
  const map = useMap();
  ...
  map.addControl(deleteControl);
};
```
Which solved the inconsistent UI issue, but had issues with properly reflecting the state of "has the map any shaped drawn on it?".

Then I tried creating a new control via Leaflet.js, but quickly abandoned that after a couple big limitations in the leaflet types definitions.

Finally, after some hacking, i found that we can simply overload the Leaflet-Draw controls. The whole edit controls widget (the "draw shape X" and delete button) is added by Leaflet-Draw, which extensions the functionalities of `Leaflet` (the basic leaflet.js). Leaflet-Draw adds a `EditToolbar` class with all the possible edit controls (adding shapes, editing shapes and removing shapes). By overloading/extending the `Leaflet.EditToolbar.Delete` we can change its behaviour.

sergei-maertens: sounds promising!

robinmolen:
Closing this PR in favor of https://github.com/open-formulieren/open-forms-sdk/pull/872.

[PR 872](https://github.com/open-formulieren/open-forms-sdk/pull/872) provides a far simpeler solution, while keeping the UI intact and predictable.

PR #822, opened 2025-04-09
Fit map to geometry bounds

robinmolen:
Partly closes open-formulieren/open-forms#5191

This ensures that the map shows the entire geometry shape in the summary view. Additionally, this re-fits the map when the user draws a shape.


sergei-maertens (commented):
Looking at the patch and Viktor's remark, I think we can look at the geometry type and only call the `fitBounds` if it's not a `Point` geometry. Then we keep the old behaviour for markers, while properly displaying the preview for the more complex geometries.

robinmolen:
Ah yeah, good point! I've updated the logic to only perform the `map.fitBounds(...)` for polygon and linestring features 👍

PR #828, opened 2025-05-08
Some more components conversion from JS -> TS

sergei-maertens:
Part of #445


src/components/auth/constants.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * DigiD, eHerkenning and eIDAS all use these error codes.
+ */
+export type AuthErrorCode = 'login-cancelled' | 'error';
+
+export type MessageParamName = '_digid-message' | '_eherkenning-message' | '_eidas-message';
+
+// Uses a list of tuples instead of a mapping for type safety in the consuming code,
+// as otherwise the union information is lost through Object.keys|entries
```
robinmolen, line 9: I think tuples are fine, but I don't understand the comment. Objects.entries() will keep the union of the information, turning the mapping item into a [`key`, `value`] tuple. So effectively this will result in the same stuff

src/components/auth/constants.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * DigiD, eHerkenning and eIDAS all use these error codes.
+ */
+export type AuthErrorCode = 'login-cancelled' | 'error';
+
+export type MessageParamName = '_digid-message' | '_eherkenning-message' | '_eidas-message';
```
robinmolen, line 6: Can we turn/add enums for these types? So we can use enum values for the switch cases, instead of relying on "keeping the string the same" (less error prone). Also it makes it a LOT easier to see/find any cases where these values are used (making it future save)

If we are doing this, could you replace the hard-coded strings with these enums? (For example, src/components/auth/AuthenticationError.stories.ts should then be updated)

src/components/auth/AuthenticationError.stories.ts
robinmolen, line 1: Lets change to `.tsx` :)

robinmolen (changes requested):
I have some comments. Its a mix of file extension consistency, some small simplifications and type safety :)

src/components/AbortButton/AbortButton.stories.ts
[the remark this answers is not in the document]
sergei-maertens: `.ts` is preferred for stories, we only to `.tsx` if we're rendering JSX. For the majority of the stories, no JSX is required, usually only when a custom `render` function or decorators are used.

src/components/auth/AuthenticationError.stories.ts
sergei-maertens: see reasoning above - the majority of the stories are `.ts` already

src/components/auth/AuthenticationError.tsx
```diff
@@ -0,0 +1,90 @@
...
+ * return them.
+ *
+ * As soon as one of the candidates is found, it is returned, as they are mutually
+ * exclusive.
+ */
+const useDetectAuthErrorMessage = (): [MessageParamName, AuthErrorCode] | undefined => {
+  const [params] = useSearchParams();
+  for (const [key] of MAPPING_PARAMS_SERVICE) {
+    const errorCode = params.get(key);
+    if (errorCode) return [key, errorCode as AuthErrorCode];
```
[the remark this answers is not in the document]
sergei-maertens: I've really tried but TS loses the context of the union of literals and makes the types a generic `string` key. The tuple keeps it.

Additionally, making it an object implies that you can have multiple key-value pairs which actually is not the case. There's always ever one error key and code pair, so a tuple is the semantically correct type.

src/components/auth/constants.ts
sergei-maertens: I've always read/heard that you should avoid Typescript enums. I don't really see a benefit here either, this is a set of param names that should match the backend and it's easy to CTRL + F them plus we have type safety through the union.

There is also a possibility that in the future the set of available auth error codes will depend on the message param name, now it's coincidence that these combinations "work out", so then the type would have to move away from an enum anyway.

There's also no precedent in our codebase anywhere for enums, so I'd rather stick to the simplicity of a union here, less concepts/data types to keep in mind.

src/components/auth/constants.ts
sergei-maertens: See https://github.com/open-formulieren/open-forms-sdk/pull/828/files#r2084386797

It didn't, I've tried with all kinds of variations of `Object.entries` and `Objects.keys/values` + even explicit `for` loops, every time the typescript inference lost the union nature - it doesn't see `keyof typeof ServiceMapping` but instead infers the common type, which is `string`. See https://stackoverflow.com/questions/55012174/why-doesnt-object-keys-return-a-keyof-type-in-typescript

Feel free to take a stab at it, but I know that `Object.FOO` is quite notorious in getting exactly what you want.

src/components/auth/constants.ts
robinmolen: I was unaware of the static type checking (fricking nice btw!), so this comment is indeed invalid

src/components/auth/constants.ts
robinmolen: Okay check. Then lets keep it as is

src/components/auth/AuthenticationError.tsx
```diff
@@ -0,0 +1,90 @@
...
+          defaultMessage: 'Je hebt het inloggen met {service} geannuleerd.',
+        },
+        {service: service}
+      );
+    }
+  }
+
+  // otherwise, look at the service itself to return the appropriate message
+  switch (parameter) {
+    case '_digid-message': {
```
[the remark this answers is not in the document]
robinmolen: Check. Yeah my fear was regarding magic strings, but i didn't know static type checking fixed that.

Yeah for exhaustiveness checks i cannot think of a nice/clean solution (which, enums would indeed not solve either..).

Okay, lets keep this as is

robinmolen (approved):
Looks good!

PR #829, opened 2025-05-20
Fix unexpected initial data reference behaviour

viktorvanwijk:
Closes open-formulieren/open-forms#5266

**Changes:**
* Initial data reference is now used together with the form ID as a storage key
* Switched from local storage to session storage to keep submissions contained within a single tab

### DIFF
--- a/.storybook/decorators.tsx
+++ b/.storybook/decorators.tsx
@@ -5,25 +5,13 @@ import {Document} from '@utrecht/component-library-react';
  * Storybook does not have a before/after cleanup cycle, and localStorage would in
  * these situations break story/test isolation.
  *
+ * Also, the submission ID is persisted in the sessionStorage once a submission is started, and we
+ * need to clear it for stories.
+ *
  * This decorator is applied to every story to reset the storage state.
  */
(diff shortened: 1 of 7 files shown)


viktorvanwijk:
There are still two problems with this:
1. Using hash-based routing, the `Form` does not re-render when adding a `initial_data_reference` query param, e.g. from `...#/test-form` to `...#/test-form?initial_data_reference=foo`. However, it does re-render when removing it.
2. When there is no submission in the local/session storage, it will try to fetch one from the `submission_uuid` query param. If there is an initial data reference present for this submission, the `setSubmissionId` and `removeSubmissionId` in `useRecycleSubmission` will be incorrect, because they are fetched from the storage (without the initial data reference in the storage key) before the submission is loaded.

I'm a bit blocked on this @sergei-maertens. Tried solving the first one by updating the state (this should trigger a component re-render if I understand correctly) or reloading the page completely (this had some side-effects of multiple unwanted reloads and couldn't figure out why). And for the second one I'm not entirely sure how to rewrite everything to fix it.

viktorvanwijk:
#831 was implemented  to fix the related issue

PR #831, opened 2025-05-26
Don't offer to re-use submission for (new) initial data reference

sergei-maertens:
Fixes open-formulieren/open-forms#5266

The presence of an initial data reference parameter implies it's a new reference, which would invalidate any existing submissions stored in the local storage. So, the easiest (but not most elegant) patch for the observed problem is to ignore any existing submission ID from the local storage (or session storage, if you will) when this query param is present.

This works because the parameter is only used to bootstrap the process and kept around long enough to be able to send it with the submission create call. After that, it's no longer passed around in the URL, and navigating back to the form start step while on step 2 or further will present the 'continue or reset' screen as usual because the parameter is missing.

When the parameter is present, it's because of the user clicking a link or manually pasting the link back into the browser, at which point we can reasonably assume they're trying to start a new form submission.

This patch does not interfere with hitting F5 anywhere in the form, even refreshing on the start page again has the expected behaviour, since a submission still needs to be created then.

### DIFF
--- a/src/hooks/useRecycleSubmission.js
+++ b/src/hooks/useRecycleSubmission.js
@@ -5,10 +5,18 @@ import {useAsync, useLocalStorage} from 'react-use';
 import {ConfigContext} from 'Context';
 import {apiCall} from 'api';
 
+import useInitialDataReference from './useInitialDataReference';
+
 const useRecycleSubmission = (form, currentSubmission, onSubmissionLoaded, onError = () => {}) => {
   const location = useLocation();
   const config = useContext(ConfigContext);
   const [params] = useSearchParams();
+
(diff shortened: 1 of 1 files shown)


sergei-maertens:
> Yeah, this works as well.
> 
> The issue with hashed-based routing still exists though. If you are on the start page of a form and add an initial data reference to the url, the component doesn't re-render and still shows the 'continue submssion' buttons. Is this something we should look into further?

Let's not worry about that since it's not a realistic/valid use case for someone to manually mess with URLs like that. As long as it works in the described situation from the ticket, I'm good with this.

> Also, do we still want to switch to session storage?

Let's do this yeah, but in a separate PR then!

PR #843, opened 2025-07-04
[OF#5268] Fix duplicated data in the Partners' summary page

vaszig:
Closes open-formulieren/open-forms#5268 partly

### DIFF
--- a/src/components/FormStepSummary/ComponentValueDisplay.jsx
+++ b/src/components/FormStepSummary/ComponentValueDisplay.jsx
@@ -1,4 +1,3 @@
-import {DataList, DataListItem, DataListKey, DataListValue} from '@utrecht/component-library-react';
 import PropTypes from 'prop-types';
 import React, {Suspense} from 'react';
 import {FormattedDate, FormattedMessage, FormattedNumber, FormattedTime, useIntl} from 'react-intl';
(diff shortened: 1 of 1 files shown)


src/components/FormStepSummary/ComponentValueDisplay.jsx
```diff
@@ -174,44 +171,9 @@ const NumberDisplay = ({component, value}) => {
...
-
-          {/* Divider between partners */}
-          {index < value.length - 1 && <hr className="utrecht-hr" />}
-        </div>
-      ))}
-    </>
-  );
+const PartnersDisplay = () => {
+  // the partners data is handled in the backend, so we do not want to show it twice
+  return;
```
viktorvanwijk, line 176: returning `<EmptyDisplay />` is not necessary?

viktorvanwijk (approved):

src/components/FormStepSummary/ComponentValueDisplay.jsx
vaszig: Yes, I had it in this way at the beginning but I show that an extra gap was added and I didn't find it useful tbh. If you think it's needed I can change it of course.

viktorvanwijk: It does keep it consistent with the rest, but if it looks weird, leave it out I would say

PR #844, opened 2025-07-09
Convert more components to TS

sergei-maertens:
Part of #445


src/components/LoginOptions/index.tsx
```diff
@@ -1,36 +1,50 @@
-import PropTypes from 'prop-types';
 import React from 'react';
 import {FormattedMessage} from 'react-intl';
 import {useSearchParams} from 'react-router';
 
 import {Literal} from 'components/Literal';
 import {getCosignLoginUrl, getLoginUrl} from 'components/utils';
```
viktorvanwijk, line 6: Also with '@' or not?

viktorvanwijk (approved):

PR #847, opened 2025-07-18
[#825] Add Children custom component

### LINKED ISSUE #825: Implement custom component type "children"
Following the changes to the backend and the builder, ensure the new component type can be used in the SDK.

We take into account that multiple children can be registered in the BRP, but allow adding the details of only one.

**Acceptance criteria**

- [x] When children details have been retrieved, they are displayed and cannot be modified. They are displayed in a table.
- [x] If no children details were retrieved, the user can add child details.
- [x] Both cases result in the same component variable being submitted.
- [x] If enabled, a subset of values can be selected (checkbox per item).
- [x] There is backend validation that prevents tampering.

**Tasks**

- [x] Add a custom formio component with a custom template
- [x] Ensure that the read-only fields are displayed with the fixed layout: BSN, first name(s), date of birth
- [x] Ensure that labels are translatable
- [x] Check if the NL DS utrecht community components have tables we can use
- [x] If the `value.length === 0`, render a button to add children
- [x] Upon button click, a modal opens with fixed fields to enter the details
    - The modal is pure React. The button click is formio and activates the modal. See the email verification feature for how to do this. These are the same fields as for the partners - even though not all of them are displayed in the table.
    - On submit of the modal, run client-side validation and if okay, update the formio value/state.
    - You can add multiple children
- [x] Ensure the component is documented in Storybook
- [ ] NTH: implement it in the new formio-renderer too so we have less catch-up to do. Check what's easiest to coordinate this.

vaszig:
Closes #825 partly

### Requires

- [x] open-formulieren/formio-builder#230
- [x] open-formulieren/types#71
- [x] open-formulieren/design-tokens#58

### DIFF
--- a/src/formio/components/Children/Children.stories.js
+++ b/src/formio/components/Children/Children.stories.js
@@ -224,9 +224,14 @@ export const DuplicateBsns = {
     const user = userEvent.setup();
 
     await step('Edit the child', async () => {
-      const editIcons = await canvasElement.querySelectorAll('i.fa-pen');
-      await userEvent.click(editIcons[1]);
+      const table = await canvas.findByRole('table');
+      const rows = await within(table).findAllByRole('row');
+      expect(rows).toHaveLength(3);
 
+      const editIcons = await canvasElement.querySelectorAll('i.fa-pen');
+      expect(editIcons).toHaveLength(2);
+
+      await userEvent.click(editIcons[1]);
       const modal = await screen.findByRole('dialog');
       const modalWithin = within(modal);
 
--- a/src/formio/components/Children/ChildrenForm.tsx
+++ b/src/formio/components/Children/ChildrenForm.tsx
@@ -10,7 +10,7 @@ import {
 } from '@utrecht/component-library-react';
 import '@utrecht/table-css';
 import {UUID} from 'crypto';
-import {FormattedMessage} from 'react-intl';
+import {FormattedMessage, useIntl} from 'react-intl';
 
 import FAIcon from 'components/FAIcon';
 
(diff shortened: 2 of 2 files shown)


robinmolen (changes requested):
A few improvements to the code and the test structure 🙂

src/formio/components/Children/Children.jsx
```diff
@@ -0,0 +1,238 @@
...
+  handleChildSave(newChild) {
+    const childExists = newChild.__addedManually === true;
+    const oldChild = this.state.childBeingEdited;
+
+    let updated;
+    if (childExists) {
+      // Update the existing child
+      updated = this.dataValue.map(child =>
+        child.bsn === oldChild.bsn ? {...child, ...newChild} : child
+      );
```
vaszig, line 154: @robinmolen apart from your comments I changed this here too because each time the user would modify the BSN a new child was added (which didn't make sense to me, if I want to modify the specific child I want to do exactly this).

src/formio/components/Children/types.ts
```diff
@@ -0,0 +1,10 @@
+import type {ChildDetails} from '@open-formulieren/types';
+
+export interface ChildExtendedDetails extends ChildDetails {
+  // distinguish a child added manually by the user (when no data was retreived from the
+  // relevant service) - this is used only for the SDK to serve extra functionality for
+  // the children
+  __addedManually?: boolean;
+  // distinguish a child selected by the user (checkbox)
+  selected?: boolean;
```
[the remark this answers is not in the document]
robinmolen: Nice!
I think "__addedManually" could have been left as `__addedManually?: true`. This is because, either the child was added manually (in that case, `__addedManually` will be `true`), or the child was added through some service (in which case, `__addedManually` would be `undefined`)

I'll do a second review and check your code to see if my understand (the situation I described above) is correct :)

robinmolen (changes requested):
I found a bug :) Besides that, it looks great!

If we/you can patch that bug, I think we can merge it

src/formio/components/Children/Children.jsx
vaszig: Thanks for the proper review/testing!Yes, my first thought was to disable the BSN field in the edit mode but couldn't do that (requires changes in the form renderer I guess). I think I have found somthing though

robinmolen (changes requested):
I have 1 small comment 🫣

robinmolen (approved):

PR #849, opened 2025-07-23
Convert i18n code to Typescript

### LINKED ISSUE #445: Adopt typescript

sergei-maertens:
Partly closes #445

Working my way up to convert sdk.jsx - now the i18n modules have been converted and cleaned up a bit.


src/sdk.tsx
```diff
@@ -25,15 +28,51 @@ import('./formio-init');
...
+  // resolved form definition from backend
+  protected formObject: Form | null;
+
+  protected useHashRouting: boolean;
+  protected routerBasePath: string;
+  protected browserBasePath: string;
+  protected clientBaseUrl: string;
+  protected baseDocumentTitle: string;
+
+  private root: Root | null;
```
sergei-maertens, line 66: I opted to make most of the API protected/private/internal here because it *is* our public API that we expose to embedders, but that requires some error suppression in `sdk.spec.ts` :/

viktorvanwijk (approved):

PR #851, opened 2025-07-25
Fix exports for downstream consumers

### LINKED ISSUE #850: OpenFormsModule isn't exported anymore from version 3.1.0 and higher
At the Municipality of The Hague, we include all components from Open Forms in our own Storybook with the The Hague theme (https://nl-design-system.github.io/denhaag/?path=/story/open-forms-bsn--default). However, after upgrading from version 3.0.0 to version 3.1.0, we’re seeing the message "Unknown component: <name>" for several components. I did some debugging, and it seems to be caused by the OpenFormsModule not being exported anymore, which means the custom Open Forms components are no longer being registered in FormIO.

Image

sergei-maertens:
Closes #850

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,8 @@
   "exports": {
     ".": "./dist/open-forms-sdk.js",
     "./styles.css": "./dist/open-forms-sdk.css",
+    "./formio-module": "./dist/esm/formio/module.js",
+    "./formio-templates": "./dist/esm/formio/templates/library.js",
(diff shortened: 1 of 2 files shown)


sergei-maertens:
Backports:

* 3.2.x: 3cedf3db093fb947fc519b7102073c61be140943
* 3.1.x: 099b53fda06278b75fcd2a29576290ffaadd1fd7

PR #853, opened 2025-08-01
Fix the behaviour of `selected` property in the Children component

vaszig:


src/formio/components/Children/AddChildModal.stories.ts
```diff
@@ -3,7 +3,7 @@ import {fn} from '@storybook/test';
 
 import {ConfigDecorator} from 'story-utils/decorators';
 
-import AddChildModal from './AddChildModal';
+import {AddChildModal} from './AddChildModal';
```
sergei-maertens, line 6: why the import change?

sergei-maertens: yeah, in general using the default export makes it a bit more clear what the public API is

PR #862, opened 2025-08-29
Some more JSX -> TS conversion

sergei-maertens:
Part of #445


src/scss/components/_login-button-logo.scss
```diff
@@ -33,7 +30,7 @@
   }
 
   &__image {
-    max-height: $grid-row-height;
-    max-width: $grid-row-height;
+    max-height: var(--of-login-button-logo-image-max-max-height, 46px);
```
vaszig, line 33: Where is this coming from?

vaszig (commented):
Is there a reason that `components/Button/OFButton.stories.jsx` is still jsx?

src/scss/components/_login-button-logo.scss
sergei-maertens: sass-variables are static and set at build-time, which is not very flexible. So I stumbled on it and decided to change it into a CSS variable which can be overridden in custom themes!

PR #864, opened 2025-09-03
Implement WMS tile layers in map component

robinmolen:
Partly closes open-formulieren/open-forms#5575

Implementing WMS tile layers, allowing for more data and visuals to be presented to end-users. The WMS layers can be toggled on and off, using the newly added `layers` menu. By default, all WMS layers are shown.

~This PR relies on https://github.com/open-formulieren/types/pull/74. Its `open-forms/types` dependency should be updated when https://github.com/open-formulieren/types/pull/74 has been merged~
The types have been updated. So this PR can safely be merged

### DIFF
--- a/src/components/Map/LeafletMapSearchControl.tsx
+++ b/src/components/Map/LeafletMapSearchControl.tsx
@@ -6,6 +6,7 @@ import {useMap} from 'react-leaflet';
 
 import {ConfigContext} from 'Context';
 
+import './LeafletMapSearchControl.scss';
 import OpenFormsProvider from './provider';
 import {searchControlMessages} from './translations';
 
--- a/src/components/Map/LeafletMapLayersControl.scss
+++ b/src/components/Map/LeafletMapLayersControl.scss
@@ -1,4 +1,10 @@
-.leaflet-control.leaflet-control-layers {
-  // Preventing layers menu from expanding beyond the zoom-buttons on mobile devices.
-  max-inline-size: calc(100vw - 110px);
+.leaflet-control-layers {
+  // Making sure that the expanded map layers menu don't go over the other map controls.
+  &-expanded {
+    max-inline-size: calc(100% - 50px);
+  }
+
+  &-overlays {
+    word-break: break-word;
+  }
 }
--- a/src/components/Map/LeafletMapSearchControl.scss
+++ b/src/components/Map/LeafletMapSearchControl.scss
@@ -1,0 +1,6 @@
+@media (max-width: 400px) {
+  // Making sure that the search input doesn't go over the other map controls.
+  .leaflet-control-geosearch > form > input {
+    min-width: 180px;
+  }
+}


src/components/Map/LeafletMapSearchControl.tsx
```diff
@@ -0,0 +1,109 @@
...
+    updateMap,
+    notFoundMessage,
+  } = options;
+
+  const buttonLabel = intl.formatMessage(searchControlMessages.buttonLabel);
+
+  useEffect(() => {
+    const provider = new OpenFormsProvider(baseUrl);
+    // Leaflet-geosearch isn't very typescript friendly...
+    const searchControl = new (GeoSearchControl as unknown as GeoSearchControlConstructor)({
```
viktorvanwijk, line 66: Seems cursed hehe

src/components/Map/translations.ts
```diff
@@ -155,6 +155,7 @@ const applyLeafletTranslations = intl => {
       },
     },
     buttons: {
+      ...Leaflet.drawLocal.edit.toolbar.buttons,
```
viktorvanwijk, line 158: Why is this necessary?

src/components/Map/translations.ts
```diff
@@ -184,6 +186,7 @@ const applyLeafletTranslations = intl => {
     },
   };
   Leaflet.drawLocal.draw.handlers.polyline = {
+    ...Leaflet.drawLocal.draw.handlers.polyline,
```
viktorvanwijk, line 189: Markers and polygons do not need this addition?

src/components/Map/Map.stories.jsx
```diff
@@ -145,3 +145,39 @@ export const MapWithInteractions = {
...
+      type: 'Point',
+      coordinates: [5.284580856043387, 52.120930596779296],
+    },
+    wmsTileLayers: [
+      {
+        uuid: '931f18f0-cedc-453b-a2d5-a2c1ff9df523',
+        url: 'https://service.pdok.nl/lv/bag/wms/v2_0',
+        label: 'BAG Pand and Verblijfsobject layer',
+        layers: ['pand', 'verblijfsobject'],
+      },
```
viktorvanwijk, line 162: Maybe worth adding another layer to see how the interaction works?

viktorvanwijk (commented):
Few questions :)

A minor thing I noticed when interacting in storybook is that the max zoom level seems to be ignored when you select a layer:
image
image
When clicking on '+', the map turns gray

src/components/Map/translations.ts
robinmolen: No, for Markers and Polygons we supply all the possible translations. For polyline we don't supply a translation for one property, so by destructing the polyline props, we make sure all properties are provided (so we don't miss any translations).

I'll make the applying of the default props more explicit.

src/components/Map/translations.ts
robinmolen: To add the default translations 🙂

robinmolen:
I've fixed the "selecting a layer removed the max zoom level limit". I'm not sure why this happend, as we don't do anything with this... But explicitly defining the min and max levels solves it 👍

src/components/Map/LeafletMap.tsx
```diff
@@ -1,30 +1,38 @@
+import {MapComponentSchema} from '@open-formulieren/types';
```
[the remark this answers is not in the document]
robinmolen: Haha, nice. I've updated them 👍

src/components/Map/types.ts
```diff
@@ -0,0 +1,22 @@
+import type {MapComponentSchema} from '@open-formulieren/types';
+
+export type Coordinates = [number, number];
+
+type PointGeometry = {
+  type: 'Point';
+  coordinates: Coordinates;
+};
```
sergei-maertens, line 8:
```suggestion
interface PointGeometry {
  type: 'Point';
  coordinates: Coordinates;
};
```
or is there a particular reason why you use a `type` instead of `interface`?

src/components/Map/provider.tsx
```diff
@@ -0,0 +1,45 @@
+import {get} from 'api';
+
+type AddressSearchResult = {
```
sergei-maertens, line 3: this can also be an `interface`

src/components/Map/provider.tsx
```diff
@@ -0,0 +1,45 @@
...
+    this.endpoint = `${baseUrl}geo/address-search`;
+  }
+
+  parse(results: AddressSearchResult[]) {
+    return results.map(location => ({
+      label: location.label,
+      x: location.latLng.lng,
+      y: location.latLng.lat,
+    }));
+  }
```
sergei-maertens, line 30: can you add an explicit return type?

src/components/Map/provider.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  async search({query}: {query: string}) {
```
sergei-maertens, line 32: explicit return type would be nice

src/components/Map/NearestAddress.tsx
```diff
@@ -9,11 +8,16 @@ import Loader from 'components/Loader';
 import {getBEMClassName} from 'utils';
 
 import './MapAddress.scss';
+import {Coordinates} from './types';
```
sergei-maertens, line 11:
```suggestion
import type {Coordinates} from './types';
```

src/components/Map/mocks.tsx
sergei-maertens, line 1: can be `mocks.ts` instead of `mocks.tsx`, no?

src/components/Map/provider.tsx
sergei-maertens, line 1: Can be `.ts` instead of `.tsx`

src/components/Map/provider.spec.tsx
sergei-maertens, line 1: Can be `.ts` instead of `.tsx`

src/components/Map/LeafletMapLayersControl.tsx
```diff
@@ -0,0 +1,49 @@
...
+  Record<Overlay['type'], React.ElementType<TileLayerControlProps>>
+> = {
+  wms: WMSTileLayerControl,
+};
+
+interface LayersControlProps {
+  overlays: MapComponentSchema['overlays'];
+}
+
+const LayersControl: React.FC<LayersControlProps> = ({overlays}) => {
```
sergei-maertens, line 34: see the remark in Chromatic about the mobile viewport where this overlaps the zoom buttons - perhaps we need some CSS to limit the max width

sergei-maertens (commented):
Nice to get this in TS too!

src/components/Map/LeafletMapLayersControl.tsx
robinmolen: Mhhh, good that you mention mobile. It gives me a reminder that that other ticket might also needs some special mobile treatment.

I'll take a look at the Chromatic 👍

src/components/Map/provider.tsx
robinmolen: Yep, added it

src/components/Map/NearestAddress.tsx
robinmolen: Damn, good eye 😉

src/components/Map/LeafletMapLayersControl.tsx
robinmolen: I've added some `max-inline-size` styling that prevents the layers menu from becoming too big. Tested on screen-size 320px

PR #865, opened 2025-09-03
Upgrade to formio-renderer 0.8.x

sergei-maertens:
Some initial pre-work for the bigger new-renderer integration (opt-in experimental stuff), part of open-formulieren/open-forms#5133

This upgrades to the latest renderer package version and replaces the code that was ported.


src/components/forms/SelectField/AsyncSelectField.tsx
```diff
@@ -0,0 +1,24 @@
+// TODO: fixup the import once the export in the package is fixed
```
sergei-maertens, line 1: self-reminder

src/components/forms/SelectField/SelectField.stories.jsx
```diff
@@ -63,47 +64,20 @@ export const Static = {
...
-};
-
-const delay = async delay => await new Promise(resolve => setTimeout(resolve, delay));
-
 export const Async = {
   render: function Render({getOptionsDelay, dynamicOptions, ...args}) {
     const serializedOptions = JSON.stringify(dynamicOptions);
     const getOptions = useCallback(async () => {
-      await delay(getOptionsDelay);
+      await sleep(getOptionsDelay);
```
vaszig, line 71: Is sleep used for more general purposes and delay for user events like typing?This is what I understood.

vaszig (commented):
So, what you aim to do with this is upgrade the version of the renderer, which version contains the new `NumberField` and the `SelectField` (and of course we make use of them instead of the old ones), right?

vaszig (approved):

PR #866, opened 2025-09-10
Immediate drawing on map, when only 1 interaction is available

robinmolen:
Closes https://github.com/open-formulieren/open-forms/issues/5378

When only one interaction is active on a map component, you no-longer have to "pick" which shape you want to draw. Instead, you can click on the map to immediately start drawing a new shape. You can click the right mouse button, to cancel the drawing.

To reduce the visual noise (and remove "useless" elements), the "draw" buttons are only shown if multiple interactions are active.


robinmolen:
For users om mobile devices this might not be desired. Will have to check out if normal navigation is still possible, or if directly "drawing" shapes prevents normal interaction

PR #867, opened 2025-09-18
[open-formulieren/open-forms#5619] 3.3.0 alpha.2

SonnyBA:
Adds changes for the `3.3.0-alpha.2` release. Partially closes open-formulieren/open-forms#5619.

### DIFF
--- a/src/i18n/compiled/nl.json
+++ b/src/i18n/compiled/nl.json
@@ -96,7 +96,7 @@
   "/jQkSu": [
     {
       "type": 0,
-      "value": "Add child"
+      "value": "Kind toevoegen"
     }
   ],
   "/oxZic": [
@@ -334,7 +334,7 @@
   "6EoG3C": [
     {
       "type": 0,
-      "value": "Save"
+      "value": "Opslaan"
     }
   ],
   "6GNK65": [
@@ -522,7 +522,7 @@
   "8MVAfQ": [
     {
       "type": 0,
-      "value": "Edit child"
+      "value": "Kind bewerken"
     }
   ],
   "8MZ9TV": [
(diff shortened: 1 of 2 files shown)


CHANGELOG.rst
```diff
@@ -2,6 +2,36 @@
...
+
+Third preview release of the upcoming 3.3.0 version.
+
+.. note:: This release adds the ability to opt-in to the new experimental form renderer.
+   The feature can be enabled on a per-form basis through the admin interface.
+
+New features
+------------
+
+* [open-formulieren/open-forms#5575] Implemented WMS layers in the map component.
```
vaszig, line 16: Should it be like this or just `[#5575]`? I don't remember the correct one but I see all the previous ones have the short one. Of course it's not clickable but just curious.

vaszig (approved):
Let's also update the `.bumpversion.toml` with the new version.

CHANGELOG.rst
vaszig: The correct way to refer to the backend issue (because it is an issue created there) is indeed this one but I mentioned this because I saw in the code that we do not do it this way.

sergei-maertens: indeed - the people reading the changelogs are not expected to understand this format so we just mention the issue numbers. Possibly we have a future improvement here to properly link to SDK/backend issues, but for now this is good enough.

src/i18n/compiled/en.json
sergei-maertens, line 1: a lot of translations seem to be removed here - did you make sure that your dependencies were up to date locally before running the translation scripts?

sergei-maertens (changes requested):
Please also check the Chromatic comments - it's okay to create follow up issues for the final release and not block this alpha release for them.

vaszig:
FYI, open-formulieren/open-forms#5630 was created for the broken css in the `children` component.

sergei-maertens:
I accepted the changes in Chromatic because we have issues to track the CSS and translation work :)

PR #872, opened 2025-09-22
Feature/5191 custom delete button

robinmolen:
Partly closes https://github.com/open-formulieren/open-forms/issues/5191

Re-using the Leaflet-Draw delete control to provide "single" click deletion.

The default Leaflet-Draw delete button opens a delete menu and requires multiple user input steps to remove a drawn shape from the map component. To improve the useability, the delete menu has been removed, allowing "single" click shape removal.

By clicking the delete button, a confirmation message will be shown to make sure that user data isn't accidentally removed. After confirming the removal, the drawn shape on the map is automatically removed.

## Technical note:
This is done by extending the Leaflet-Draw `Leaflet.EditToolbar.Delete` `enable` functionality. This changes the behaviour of "onclick" for the delete button.

React leaflet provides another solution for this, by defining and connection a custom controller https://react-leaflet.js.org/docs/v4/core-api/#createcontrolcomponent. This however has the drawback of unpredictable UI rendering, as the placement in the DOM is done by React leaflet, and the placement/order in the `.tsx` React component doesn't have any impact.

The other approach, which we use for most of our other custom controls, is to connect the controller directly via the `useMap` hook. This, for some reason, had some difficulty with correctly providing information about the current drawn shapes. In addition, the UI would be changed in a manner that would separate the "draw shape X" buttons and the delete button. By extending the Leaflet-Draw `Leaflet.EditToolbar.Delete` we keep the UI intact.

See https://github.com/open-formulieren/open-forms-sdk/pull/821/files#r2368521776 for further technical comments.

### DIFF
--- a/src/components/Map/LeafletMap.tsx
+++ b/src/components/Map/LeafletMap.tsx
@@ -21,6 +21,7 @@ import LayersControl from './LeafletMapLayersControl';
 import SearchControl, {type GeoSearchShowLocationEvent} from './LeafletMapSearchControl';
 import NearestAddress from './NearestAddress';
 import {DEFAULT_INTERACTIONS, DEFAULT_LAT_LNG, DEFAULT_ZOOM} from './constants';
+import {overloadLeafletDeleteControl} from './deleteControl';
 import {CRS_RD, TILE_LAYER_RD, initialize} from './init';
 import {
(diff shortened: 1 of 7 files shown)


sergei-maertens (approved):
A work of art!

PR #878, opened 2025-09-26
More typescript

sergei-maertens:
Part of #445


src/components/Sessions/SessionTrackerModal.tsx
```diff
@@ -1,32 +1,32 @@
...
 
-RelativeTimeToExpiry.propTypes = {
-  numSeconds: PropTypes.number.isRequired,
-};
-
-const useTriggerWarning = ms => {
-  let timeout;
+type UseTriggerWarning = [boolean, () => void];
 
+const useTriggerWarning = (ms: number | null): UseTriggerWarning => {
```
robinmolen, line 28: `ms` is always a number, right? I only see it being used in the `SessionTrackerModal`, and i believe there should always be a number value as input

PR #879, opened 2025-09-26
:recycle: Use formio-renderer date field

sergei-maertens:
* Removed dead Date picker related code (everything was ported to formio-renderer)
* Use formio-renderer Date field component and widget options
* Updated tests to account for async l10n loading
* Fixed loading appointment times when date input has invalid ISO-8601 date


robinmolen (approved):
LGTM!

PR #884, opened 2025-09-30
Convert appointments to typescript

sergei-maertens:
Part of #445

A big one - this is quite a complex setup and I don't want to refactor the implementation too much tbh!

### DIFF
--- a/src/components/appointments/CreateAppointment/CreateAppointmentState.tsx
+++ b/src/components/appointments/CreateAppointment/CreateAppointmentState.tsx
@@ -36,7 +36,6 @@ const extractStepErrors = (
 
   switch (currentStep) {
     case 'producten': {
-      console.log('initialErrors', initialErrors);
       const stepInitialTouched: StepTouched<'producten'> = {};
       const stepInitialErrors: StepErrors<'producten'> = {};
       for (const key of ERROR_KEYS_BY_STEP.producten) {
@@ -45,7 +44,6 @@ const extractStepErrors = (
         stepInitialErrors[key] = errors;
         stepInitialTouched[key] = initialTouched[key];
       }
-      console.log('stepInitialTouched', stepInitialTouched);
       return {stepInitialTouched, stepInitialErrors};
     }
     case 'kalender': {
@@ -81,11 +79,15 @@ const extractStepErrors = (
   }
 };
 
+type PartialAppointmentDataByStep = Partial<{
+  [K in keyof AppointmentDataByStep]: Partial<AppointmentDataByStep[K]>;
+}>;
+
 interface BuildContextValueOpts {
   submission: Submission | null;
   currentStep: AppoinmentStep | '';
-  appointmentData: Partial<AppointmentDataByStep>;
-  setAppointmentData?: (values: Partial<AppointmentDataByStep>) => void;
+  appointmentData: PartialAppointmentDataByStep;
+  setAppointmentData?: (values: PartialAppointmentDataByStep) => void;
   appointmentErrors?: {
     initialTouched: FormikTouched<AppointmentData>;
     initialErrors: FormikErrors<AppointmentData>;
@@ -106,14 +108,10 @@ export const buildContextValue = ({
   setAppointmentErrors = () => {},
   resetSession = () => {},
 }: BuildContextValueOpts): CreateAppointmentContextType => {
-  const submittedSteps = (Object.keys(appointmentData) as AppoinmentStep[]).filter(
-    subObject => Object.keys(subObject).length
-  );
   const mergedAppointmentData: Partial<AppointmentData> = (
     Object.keys(appointmentData) as AppoinmentStep[]
   ).reduce((accumulator, key) => {
     const stepData = appointmentData[key];
-    if (!stepData) return accumulator;
     return {...accumulator, ...stepData};
   }, {} satisfies Partial<AppointmentData>);
 
@@ -127,7 +125,7 @@ export const buildContextValue = ({
     submission,
     appointmentData: mergedAppointmentData,
     stepData: currentStep === '' ? {} : (appointmentData[currentStep] ?? {}),
-    submittedSteps,
+    submittedSteps: Object.keys(appointmentData) as AppoinmentStep[],
     submitStep: values => setAppointmentData({...appointmentData, [currentStep]: values}),
     setErrors: setAppointmentErrors,
     stepErrors: {initialTouched: stepInitialTouched, initialErrors: stepInitialErrors},
@@ -157,7 +155,7 @@ export const CreateAppointmentState: React.FC<CreateAppointmentStateProps> = ({
   resetSession,
   children,
 }) => {
-  const [appointmentData, setAppointmentData] = useSessionStorage<Partial<AppointmentDataByStep>>(
+  const [appointmentData, setAppointmentData] = useSessionStorage<PartialAppointmentDataByStep>(
     SESSION_STORAGE_KEY,
     {}
   );
--- a/src/components/appointments/types.ts
+++ b/src/components/appointments/types.ts
@@ -3,8 +3,6 @@ import type {FormikErrors, FormikTouched} from 'formik';
 
 import type {AppointmentProduct} from '@/data/appointments';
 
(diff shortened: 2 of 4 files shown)


sergei-maertens:
@robinmolen *when* this is ready to merge, please don't yet do so since I want to squash-and-merge this PR due to the large number of commit messages that pollute the history otherwise.

src/components/appointments/types.ts
```diff
@@ -0,0 +1,39 @@
+import type {AppointmentProduct} from '@/data/appointments';
+
+export type AppoinmentStep = 'producten' | 'kalender' | 'contactgegevens';
```
robinmolen, line 6: Would it not be better to use:
```suggestion
export type AppoinmentStep = keyof AppointmentDataByStep;
```
This way, the `AppoinmentStep` and `AppointmentDataByStep` will always be in sync.

src/components/appointments/types.ts
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Information gathered per step to create an appointment.
+ *
+ * The keys are the slugs/identifiers of the individual steps.
+ */
+export interface AppointmentDataByStep {
+  producten: {
+    products: AppointmentProduct[];
+  };
+  kalender: {
```
robinmolen, line 17: This might go a bit far, but wouldn't it be nice to define the AppointmentSteps ('product', 'kalender', 'contactgegevens') as const/enum, and use their values in the `AppointmentDataByStep` type definition?

Just to make it super obvious that these values should not be changed. We could then also re-use them, instead of using/relaying on *"magic"* strings.

Something like:
```suggestion
enum AppointmentSteps {
  PRODUCTEN = 'producten',
  KALENDER = 'kalender',
  CONTACTGEGEVENS = 'contactgegevens',
}

/**
 * Information gathered per step to create an appointment.
 *
 * The keys are the slugs/identifiers of the individual steps.
 */
export interface AppointmentDataByStep {
  [AppointmentSteps.PRODUCTEN]: {
    products: AppointmentProduct[];
  };
  [AppointmentSteps.KALENDER]: {
  ...
```

src/components/appointments/CreateAppointment/CreateAppointment.stories.tsx
```diff
@@ -19,6 +22,24 @@ import {
...
 import CreateAppointment from './';
 
+const Wrapper: React.FC<{form: Form}> = ({form}) => {
+  const router = createMemoryRouter(routes, {
+    initialEntries: ['/afspraak-maken/'],
+    future: FUTURE_FLAGS,
+  });
+  return (
+    <FormContext.Provider value={form}>
+      <RouterProvider router={router} />
```
robinmolen, line 32: Is `PROVIDER_FUTURE_FLAGS` no-longer needed?

src/components/appointments/CreateAppointment/CreateAppointment.spec.tsx
```diff
@@ -51,11 +47,12 @@ const renderApp = (initialRoute = '/') => {
         basePath: '',
         baseTitle: '',
         requiredFieldsWithAsterisk: true,
+        debug: false,
       }}
     >
       <IntlProvider locale="en" messages={messagesEN}>
         <FormContext.Provider value={form}>
-          <RouterProvider router={router} future={PROVIDER_FUTURE_FLAGS} />
+          <RouterProvider router={router} />
```
robinmolen, line 56: Double checking that `PROVIDER_FUTURE_FLAGS` is indeed no-longer needed

src/components/appointments/CreateAppointment/Summary.spec.tsx
```diff
@@ -100,7 +105,7 @@ const renderSummary = errorHandler => {
   });
   realRender(
     <FormContext.Provider value={form}>
-      <RouterProvider router={router} future={PROVIDER_FUTURE_FLAGS} />
+      <RouterProvider router={router} />
```
robinmolen, line 109: Another double check for `PROVIDER_FUTURE_FLAGS`

robinmolen (changes requested):
Reviewed until commit 9a3b83f, will check the rest tomorrow.

Added some questions and suggestions.

Also saw some removal of `PROVIDER_FUTURE_FLAGS`, of which I wasn't sure if it was correct. If those changes were intentional, you can just resolve those comments

src/components/appointments/CreateAppointment/CreateAppointmentState.tsx
```diff
@@ -0,0 +1,195 @@
...
+  currentStep,
+  appointmentData,
+  setAppointmentData = () => {},
+  appointmentErrors = {initialTouched: {}, initialErrors: {}},
+  setAppointmentErrors = () => {},
+  resetSession = () => {},
+}: BuildContextValueOpts): CreateAppointmentContextType => {
+  const submittedSteps = (Object.keys(appointmentData) as AppoinmentStep[]).filter(
+    subObject => Object.keys(subObject).length
+  );
```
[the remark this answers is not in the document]
sergei-maertens: It determines based on the appointment data which steps have been processed client-side.

It considers a step as submitted if it's not an empty-ish object. The empty object `{}` is truthy in JS, so your suggestion is functionally different:

image

`subObj` holds the formik values for each step and will therefore always be (at least) an object/key-value pairs, but there are no guarantees from the session storage that the object will not be empty. It's a bit of a trade-off here between runtime behaviour and TS definitions because we do want to model the final appointment data strictly for the submission and that is enforced with validation (frontend and backend).

Maybe the appointment data state can be a mapped type of nested partials :thinking:

src/components/appointments/CreateAppointment/CreateAppointmentState.tsx
```diff
@@ -0,0 +1,195 @@
...
+  const mergedAppointmentData: AppointmentData = (
+    Object.keys(appointmentData) as AppoinmentStep[]
+  ).reduce((accumulator, key) => {
+    const stepData = appointmentData[key];
+    if (!stepData) return accumulator;
+    return {...accumulator, ...stepData};
```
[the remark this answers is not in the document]
sergei-maertens: yes, `stepData` can be `undefined`. I'm surprised the type checker doesn't complain about that.

src/components/appointments/CreateAppointment/Summary.spec.tsx
sergei-maertens: See previous remark

src/components/appointments/types.ts
robinmolen: Yeah your setup defines `AppointmentDataByStep` clearly, and with `AppoinmentStep` you can target the individual steps clearly. But this all relays on manually making sure the strings are still aligned.

I agree that:
```
export interface AppointmentDataByStep {
  producten: {
    products: AppointmentProduct[];
  };
}
```
is far easier to read, but it also feels kinda loose/"fragile". Definitely as this is suppose to connect to Formio keys, and any typo could cause issues.

robinmolen (approved):

src/components/appointments/CreateAppointment/CreateAppointmentState.tsx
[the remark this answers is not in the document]
robinmolen: Maybe you meant this:
```
const submittedSteps = Object.entries(appointmentData)
    .filter(([, subObject]) => Object.keys(subObject).length)
    .map(([stepKey]) => stepKey) as AppoinmentStep[];
```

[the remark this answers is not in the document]
sergei-maertens: wow this has been wrong since forever and just happens to do the right thing by coincidence :) 

checking the key length is indeed sufficient in this case, I will update!

src/components/appointments/types.ts
sergei-maertens: > Definitely as this is suppose to connect to Formio keys

that's not true though, the keys are fixed and well defined, only the `contactgegevens` step has a nested object with Formio keys inside and that's typed as `JSONObject`

The majority of the appointment flow does *not* use Formio.

PR #890, opened 2025-10-02
 add 'back to top' link to the form navigation

annashamray:
fixes: https://github.com/open-formulieren/open-forms/issues/5546 (back to top part)
depends on https://github.com/open-formulieren/design-tokens/pull/61

* [x] Create "Back to top" link in the form navigation
* [x] use "back to top" text parameter from sdk in the label
* [x] navigate to some anchor
* [x] add design token with icon variable

### DIFF
--- a/src/components/FormNavigation/FormNavigation.tsx
+++ b/src/components/FormNavigation/FormNavigation.tsx
@@ -1,11 +1,8 @@
 import {ButtonGroup} from '@utrecht/button-group-react';
 import {Icon, LinkButton} from '@utrecht/component-library-react';
-import {useContext} from 'react';
-
-import {ConfigContext} from 'Context';
 
 import AbortButton from '@/components/AbortButton';
-import Anchor from '@/components/Anchor';
+import BacktotopLink from '@/components/BacktotopLink';
 import FAIcon from '@/components/FAIcon';
 import {Literal} from '@/components/Literal';
 import PreviousLink from '@/components/PreviousLink';
@@ -29,8 +26,6 @@ const FormNavigation: React.FC<FormNavigationProps> = ({
   previousPage,
   onDestroySession,
 }) => {
-  const {backToTopText, backToTopRef} = useContext(ConfigContext);
-
   return (
     <ButtonGroup className="openforms-form-navigation" direction="column">
       {submitButton}
@@ -56,14 +51,7 @@ const FormNavigation: React.FC<FormNavigationProps> = ({
         <AbortButton isAuthenticated={isAuthenticated} onDestroySession={onDestroySession} />
       )}
 
-      {!!backToTopText && backToTopRef && (
-        <Anchor href={`#${backToTopRef}`} className="openforms-backtotop-link">
-          <Icon>
-            <FAIcon icon="" className="fa-fw" />
-          </Icon>
-          {backToTopText}
-        </Anchor>
-      )}
+      <BacktotopLink />
     </ButtonGroup>
   );
 };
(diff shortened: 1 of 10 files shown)


src/data/forms.ts
```diff
@@ -69,6 +69,7 @@ export interface Form {
     beginText: ButtonText;
     changeText: ButtonText;
     confirmText: ButtonText;
+    backToTopText: ButtonText;
```
sergei-maertens, line 72: we didn't intend to make this configurable in the backend, but rather pass it in the template down to the SDK constructor, see: 

https://github.com/open-formulieren/open-forms/blob/49fdc74dbcf14601f6f2158cef67d79e2d18acc6/src/openforms/forms/templates/forms/sdk_snippet.html#L4

and https://github.com/open-formulieren/open-forms/blob/49fdc74dbcf14601f6f2158cef67d79e2d18acc6/src/openforms/static/sdk-wrapper.mjs#L25

src/components/FormNavigation/FormNavigation.tsx
```diff
@@ -47,6 +48,13 @@ const FormNavigation: React.FC<FormNavigationProps> = ({
     {!hideAbortButton && (
       <AbortButton isAuthenticated={isAuthenticated} onDestroySession={onDestroySession} />
     )}
+
+    <Link to="#main-content" className="openforms-backtotop-link">
```
sergei-maertens, line 52: the `to` value should be passed in as an option, because the SDK itself does not render this anywhere, it's managed by the backend: https://github.com/open-formulieren/open-forms/blob/49fdc74dbcf14601f6f2158cef67d79e2d18acc6/src/openforms/templates/master.html#L34

This will be broken for customers that embed the SDK in their own CMS (see https://open-forms.readthedocs.io/en/stable/developers/embedding.html for docs on embedding)

src/data/forms.ts
annashamray: I've change it to a data set parameter in the template

src/components/FormNavigation/FormNavigation.tsx
annashamray: I've added `backToTopRef` to store id of the target node

src/components/FormNavigation/FormNavigation.tsx
```diff
@@ -24,29 +27,49 @@ const FormNavigation: React.FC<FormNavigationProps> = ({
...
+
+      {!hideAbortButton && (
+        <AbortButton isAuthenticated={isAuthenticated} onDestroySession={onDestroySession} />
+      )}
+
+      {!!backToTopText && backToTopNode && (
+        <LinkButton
+          name="backToTop"
+          className="openforms-form-navigation__backtotop-button"
+          onClick={() => backToTopNode.scrollIntoView({behavior: 'smooth'})}
```
sergei-maertens, line 63: is this because otherwise with hash-based routing things would break? Or is there another reason this can't be a plain `<a href="#ref">...</a>` construct with native browser link-click behaviour?

src/components/FormNavigation/FormNavigation.tsx
```diff
@@ -24,29 +27,49 @@ const FormNavigation: React.FC<FormNavigationProps> = ({
...
-    )}
-
-    {!hideAbortButton && (
-      <AbortButton isAuthenticated={isAuthenticated} onDestroySession={onDestroySession} />
-    )}
-  </ButtonGroup>
-);
+}) => {
+  const {backToTopText, backToTopRef} = useContext(ConfigContext);
+  const backToTopNode = backToTopRef ? document.getElementById(backToTopRef) : null;
```
sergei-maertens, line 32: doing DOM queries during React component rendering doesn't fee like something that should be allowed. Either way it's quite wasteful if you could (if you absolute need the DOM Node reference instead of just the ID value) instead resolve the element in `sdk.tsx` and put that in the configuration context instead.

PR #894, opened 2025-10-21
[OF#5694] Update texts in appointments for location and time step

vaszig:
Closes open-formulieren/open-forms#5694 partly

### DIFF
--- a/src/components/appointments/fields/LocationSelect.stories.ts
+++ b/src/components/appointments/fields/LocationSelect.stories.ts
@@ -38,7 +38,7 @@ export const MultipleCandidates: Story = {
     await userEvent.keyboard('[ArrowDown]');
     expect(await canvas.findByRole('option', {name: 'Open Gem'})).toBeVisible();
     expect(
-      await canvas.findByRole('option', {name: 'Bahamas (Nassau, Winsome Dr, 1014 EG)'})
+      await canvas.findByRole('option', {name: 'Bahamas (Winsome Dr, 1014 EG, Nassau)'})
     ).toBeVisible();
   },
 };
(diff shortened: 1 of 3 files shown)


src/data/appointments.ts
```diff
@@ -20,6 +20,8 @@ export interface Product {
 export interface Location {
   identifier: string;
   name: string;
+  city?: string;
+  address?: string;
```
viktorvanwijk: These will not be undefined, right, with the changes from open-formulieren/open-forms#5715?

src/components/appointments/fields/LocationSelect.stories.ts
```diff
@@ -37,7 +37,7 @@ export const MultipleCandidates: Story = {
     dropdown.focus();
     await userEvent.keyboard('[ArrowDown]');
     expect(await canvas.findByRole('option', {name: 'Open Gem'})).toBeVisible();
-    expect(await canvas.findByRole('option', {name: 'Bahamas'})).toBeVisible();
+    expect(await canvas.findByRole('option', {name: 'Bahamas (Nassau, Winsome Dr)'})).toBeVisible();
```
viktorvanwijk: Maybe good to also add a test that shows the behaviour when (some of) the parts are an empty string?

src/api-mocks/appointments.ts
```diff
@@ -71,11 +71,17 @@ interface LocationConfig {
 const LOCATIONS: LocationConfig[] = [
   {
     products: ['166a5c79', 'e8e045ab'],
-    location: {identifier: '1396f17c', name: 'Open Gem'},
+    location: {identifier: '1396f17c', name: 'Open Gem', city: '', address: '', postalcode: ''},
```
vaszig: We have 2 locations with empty strings as well, so these are shown in the stories. If you think we do need separate stories let me know.

src/components/appointments/fields/LocationSelect.tsx
```diff
@@ -38,10 +38,22 @@ const LocationSelect: React.FC<LocationSelectProps> = ({products}) => {
   const intl = useIntl();
   const {baseUrl} = useContext(ConfigContext);
   const productIds = products.map(prod => prod.productId).sort(); // sort to get a stable identity
+
+  const getAddressDetails = (location: Location): string => {
+    const {name, city, address, postalcode} = location;
+
+    const details = [city, address, postalcode].filter(Boolean).join(', ');
```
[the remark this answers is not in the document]
vaszig: Sure, I went with their request in the screenshot but indeed the city should be at the end. This is the way we treat it in AddressNL as well.

PR #900, opened 2025-11-04
[OF#5548] Update add-button to secondary action

vaszig:
Close open-formulieren/open-forms#5548 partly

This PR handles only the following in the SDK. Some other changes are needed in the new formio renderer (for example in the edit grid component, which is used in the appointments). Discussed the email verification and concluded that the button will stay as is because the secondary would make it hard to find in some themes.

- EditGrid
- Partners
- Children
- Formio's multiple True

### DIFF
--- a/src/formio/components/Children/ChildrenForm.tsx
+++ b/src/formio/components/Children/ChildrenForm.tsx
@@ -52,7 +52,7 @@ export const ChildrenComponent: React.FC<ChildrenComponentProps> = ({
       toggleChildSelection={toggleChildSelection}
     />
     {enableCreation && (
-      <OFButton onClick={onAddChild} variant="primary" appearance="primary-action-button">
+      <OFButton onClick={onAddChild} variant="secondary" appearance="primary-action-button">
         <FormattedMessage
           description="Add child: add child button text"
           defaultMessage="Add Child"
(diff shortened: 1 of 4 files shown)


src/components/appointments/CreateAppointment/CreateAppointment.stories.tsx
```diff
@@ -181,7 +181,11 @@ export const HappyFlow: Story = {
       expect(await canvas.findByText('Jouw product(en)')).toBeVisible();
       expect(await canvas.findByText('Paspoort aanvraag: 1x')).toBeVisible();
       expect(await canvas.findByText('Jouw afspraak')).toBeVisible();
-      expect(await canvas.findByText(`Open Gem, Amsterdam, ${tomorrow}, 07:00`)).toBeVisible();
+
+      // the variable `tomorrow` has leading zeros which is not the case in the shown result
```
sergei-maertens: can't we update the format then to not have these leading zeros?

sergei-maertens: I still want the locale switcher (en/nl) to work properly in SB, but it looks like `date-fns` might be doing something different than `react-intl` here :thinking: 

Maybe we can use the imperative API from react-intl instead of date-fns here?

vaszig: Unfortunately this (the above suggestion) doesn't work either..

sergei-maertens (approved):

PR #902, opened 2025-11-05
:green_heart: Fix CI

viktorvanwijk:
Using the 'P' format string will add leading zeroes to the day and month, which `useIntl().formatDate` from 
react-intl does not do, so we just hardcode the format for Dutch locale here.

This worked previously because it was merged on the 30th of October, which does not contain leading
zeroes :upside_down_face:

### DIFF
--- a/src/components/appointments/CreateAppointment/CreateAppointment.stories.tsx
+++ b/src/components/appointments/CreateAppointment/CreateAppointment.stories.tsx
@@ -169,9 +169,7 @@ export const HappyFlow: Story = {
     });
 
     await step('Fill out the contact details', async () => {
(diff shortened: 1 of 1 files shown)


sergei-maertens (approved):
@vaszig Viktor just informed me of this CI failure and fix.

I originally didn't want to format the date with a fixed format, because the data will be formatted differently if using the `en` locale and that would cause local test failures.

However, Viktor points out that the tests now fail anyway because they're hardcoded on Dutch translations anyway, so we don't even get to the date format failure. Given that my reason to block this fixed format is moot because of this, we're applying this patch and you can rebase your PR :)

PR #903, opened 2025-11-05
Wire up upload/destroy parameters for file uploads/component

sergei-maertens:
Partly closes open-formulieren/formio-renderer#99

This adds the backend implementation for the file upload functionality, actually interfacing with our API.

We need some lower-level handling of the `window.fetch` calls here to properly provide the existing mechanisms for validation errors. Once the legacy Formio.js is removed, we can use plain JSON for validation errors and use the standard `apiCall` machinery again.

TODO:

- [x] Tests
- [x] Release and upgrade to new renderer version


src/data/file-uploads.ts
```diff
@@ -0,0 +1,120 @@
+import {apiCall, destroy} from '@/api';
+import type {ApiCallOptions} from '@/api';
+
+import type {Submission} from './submissions';
+
+/**
+ * @see `#/components/schemas/TemporaryFileUpload` in the API spec.
+ */
+export interface TemporaryFileUploadBody {
```
sergei-maertens: remove - this isn't used anywhere because we do low-level `FormData` stuff

src/api.ts
```diff
@@ -202,6 +206,7 @@ const _unsafe = async <T = unknown, U = unknown>(
     method,
     headers: {
       'Content-Type': 'application/json',
+      // TODO: is this necessary? See apiCall and addHeaders
```
viktorvanwijk: Is this a TODO for now or for later?

src/data/file-uploads.ts
```diff
@@ -0,0 +1,106 @@
...
+    switch (response.status) {
+      case 400: {
+        errorMessage = await response.text();
+        break;
+      }
+      default: {
+        const contentType = response.headers.get('content-type');
+        if (contentType && contentType.indexOf('application/json') !== -1) {
+          const responseData = await response.json();
+          errorMessage = responseData.detail;
```
viktorvanwijk: This is the custom error handling required because of Formio.js?

src/api.ts
sergei-maertens: TODO for later, didn't want to fall into the distraction trap/rabbit hole now :see_no_evil:

sergei-maertens:
The lazy loading of the `File` component with this setup works as expected, we have a nice separate chunk:

image

it contains the formio-renderer code + react-dropzone dependency, shaving some 64kB of the main bundle.

PR #904, opened 2025-11-06
Pin date in appointments contact details stories

viktorvanwijk:
This step now includes a summary of the appointment details, which changed every day. To avoid regressions in the chromatic snapshots, set it to the day it was first introduced.


sergei-maertens (approved):
Was worried about the date endpoint mocks causing breakage here, but those mocks aren't used in these stories it seems.

PR #905, opened 2025-11-07
Release 3.4.0 alpha.1

sergei-maertens:
Updated changelog, bumped version number.

Most of the work is of course done in the formio-renderer.


src/i18n/messages/nl.json
```diff
@@ -274,6 +274,7 @@
   "BBWisr": {
     "defaultMessage": "{name}: {amount}x",
     "description": "Product summary on appointments contact details step",
+    "isTranslated": true,
```
vaszig: I guess this is needed because they are the same, correct?

PR #906, opened 2025-11-13
Wire up the email verification API interaction

sergei-maertens:
Closes open-formulieren/formio-renderer#181


sergei-maertens:
@SonnyBA feel free to create a backlog issue - the coverage for frontend code in all our repo's (renderer, builder, SDK) is a bit annoying. I'm not sure if tuning it is the right approach, or looking into a different instrumenter as I don't fully understand yet why it's always off by small bits. We need some time to properly look into it, for now the reviewer can make a judgement call if the coverage is sufficient or not :)

PR #909, opened 2025-11-25
Wire up `fetchDigitalAddresses` API and portalUrl

robinmolen:
Partly closes open-formulieren/formio-renderer#178

Added component parameters for customerProfile component. The `fetchDigitalAddresses` allows fetching of the digital addresses of the authenticated user. The `portalUrl` is used to direct users to the online customer interactions portal, were they can update their user information.

### DIFF
--- a/src/components/FormStep/hooks.ts
+++ b/src/components/FormStep/hooks.ts
@@ -15,6 +15,7 @@ import {get} from '@/api';
 import {getCosignStatus as getCosignStatus_} from '@/components/CoSign';
 import {getLoginUrl} from '@/components/LoginOptions/utils';
 import {assertSubmission, useSubmissionContext} from '@/components/SubmissionProvider';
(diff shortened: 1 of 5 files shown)


src/components/FormStep/hooks.ts
```diff
@@ -191,6 +191,14 @@ export const useFormioFormConfigurationParameters = (): Pick<
     [form, submissionId]
   );
 
+  const fetchDigitalAddresses = useCallback(
+    async (componentKey: string) => {
+      const endpoint = `${baseUrl}prefill/plugins/communication-preferences`;
+      return (await get(endpoint, {submissionId, componentKey}))!;
```
sergei-maertens: please implement this like the rest of the api interaction in the `src/data` folder, so that we can add types at that layer and have strongly typed code down the line.

PR #102, opened 2021-11-05
[#8] Style multiple-value input

SilviaAmAm:
Fixes #8 
Screenshot from 2021-11-05 16-42-44

PR #910, opened 2025-11-25
[#102] use LeafletMap from formio-renderer

SonnyBA:


src/data/geo.tsx
```diff
@@ -0,0 +1,115 @@
...
+    house_number: houseNumber,
+  };
+  try {
+    const result = await get<AutoCompleteResult>(`${baseUrl}geo/address-autocomplete`, params);
+    return result!;
+  } catch (error) {
+    logError(error);
+    return null;
+  }
+};
```
SonnyBA: Note that this was left untouched but is marked as addition because of the file rename,

Edit: Note to self,  the rename can be undone as the file does not contain any jsx syntax anymore.

src/data/geo.tsx
```diff
@@ -0,0 +1,115 @@
+import {get} from '@/api';
+import {logError} from '@/components/Errors';
+import { NearestLookupBody } from '@open-formulieren/formio-renderer/registry/map/types.js';
+import { SearchResult  as LeafletSearchResult, ParseArgument } from 'leaflet-geosearch/src/providers/provider.js';
+import AbstractProvider, { ProviderOptions } from 'leaflet-geosearch/src/providers/provider.js';
```
sergei-maertens: prettier won't like this formatting, see the CI build

sergei-maertens (changes requested):
See also the Chromatic denial w/r to the missing interactions.

sergei-maertens (approved):

PR #962, opened 2026-04-03
[OF#6144] Load the correct variant according to the form type

### LINKED ISSUE #6144: Load the form/SDK according to the form type
Now that 3 types of form exist (regular, appointment and single page), determine the type and render the appropriate variant/UI.

vaszig:
Closes open-formulieren/open-forms#6144 partly

### DIFF
--- a/src/components/FormLandingPage.tsx
+++ b/src/components/FormLandingPage.tsx
@@ -8,7 +8,7 @@ const FormLandingPage: React.FC = () => {
   const {preserveQueryParams} = useQueryParams();
 
   let startPageUrl: string;
-  if (type === 'single_page') {
+  if (type === 'single_step') {
     startPageUrl = 'sp';
   } else {
     startPageUrl = introductionPageContent ? 'introductie' : 'startpagina';
--- a/src/components/FormLandingPage.tsx
+++ b/src/components/FormLandingPage.tsx
@@ -4,9 +4,15 @@ import useFormContext from '@/hooks/useFormContext';
 import useQueryParams from '@/hooks/useQueryParams';
 
 const FormLandingPage: React.FC = () => {
-  const {introductionPageContent = ''} = useFormContext();
+  const {introductionPageContent = '', type} = useFormContext();
   const {preserveQueryParams} = useQueryParams();
-  const startPageUrl = introductionPageContent ? 'introductie' : 'startpagina';
+
+  let startPageUrl: string;
+  if (type === 'single_page') {
+    startPageUrl = 'sp';
+  } else {
+    startPageUrl = introductionPageContent ? 'introductie' : 'startpagina';
+  }
 
   const targetUrl = preserveQueryParams(startPageUrl);
   return <Navigate replace to={targetUrl} />;
--- a/src/routes/single-page-form.tsx
+++ b/src/routes/single-page-form.tsx
@@ -1,0 +1,28 @@
+import type {RouteObject} from 'react-router';
+
+import ErrorBoundary from '@/components/Errors/ErrorBoundary';
+import {ConfirmationView} from '@/components/PostCompletionViews';
+import {SessionTrackerModal} from '@/components/Sessions';
+
+const singlePageRoutes: RouteObject[] = [
+  {
+    index: true,
+    element: (
+      <ErrorBoundary useCard>
+        <SessionTrackerModal>
+          <></>
+        </SessionTrackerModal>
+      </ErrorBoundary>
+    ),
+  },
+  {
+    path: 'bevestiging',
+    element: (
+      <ErrorBoundary useCard>
+        <ConfirmationView onFailureNavigateTo="/:step" />
+      </ErrorBoundary>
+    ),
+  },
+];
+
+export default singlePageRoutes;
--- a/src/Context.ts
+++ b/src/Context.ts
@@ -20,6 +20,7 @@ const FormContext = React.createContext<Form>({
     previousText: {resolved: 'Previous'},
   },
   slug: '',
+  type: 'regular',
   url: '',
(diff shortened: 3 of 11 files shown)


src/routes/single-page-form.tsx
```diff
@@ -0,0 +1,38 @@
...
+  {
+    path: 'introductie',
+    element: <IntroductionPage />,
+  },
+  {
+    path: ':step',
+    element: (
+      <ErrorBoundary useCard>
+        <SessionTrackerModal>
+          <></>
```
vaszig, line 13: This one is on purpose because the component (SingleFormStepNewRenderer) is implemented in the next PR.

src/components/IntroductionPage/index.tsx
```diff
@@ -10,10 +10,15 @@ import Link from '@/components/Link';
...
-  const {name, introductionPageContent = ''} = useContext(FormContext);
+  const {name, introductionPageContent = '', type, steps} = useContext(FormContext);
   const {preserveQueryParams} = useQueryParams();
 
-  const startPageUrl = preserveQueryParams('startpagina');
+  const isSinglePageForm = type === 'single_page';
+
+  const startPageUrl = !isSinglePageForm
+    ? preserveQueryParams('startpagina')
+    : `../${steps[0].slug}`; // single page forms have only one form step
```
viktorvanwijk: For a single-page form, there is no `/stap/` before the step slug in the URL?

EDIT: I see that it will always be `/sp/`, right?

src/components/FormLandingPage.tsx
```diff
@@ -4,9 +4,18 @@ import useFormContext from '@/hooks/useFormContext';
...
+
+  const isSinglePageForm = type === 'single_page';
+
+  const startPageUrl = !isSinglePageForm
+    ? introductionPageContent
+      ? 'introductie'
+      : 'startpagina'
+    : introductionPageContent
+      ? 'sp/introductie'
+      : `sp/${steps[0].slug}`; // single page forms have only one form step
```
viktorvanwijk: IMHO, this is easier to read :)
```suggestion
  let startPageUrl: string;
  if (type === 'single_page') {
    startPageUrl = introductionPageContent ? 'sp/introductie' : `sp/${steps[0].slug}`;
  } else {
    startPageUrl = introductionPageContent ? 'introductie' : 'startpagina';
  }
```

src/components/IntroductionPage/index.tsx
vaszig: Yes, this was a first thought. Having something static for a single page form and then the step's slug.

src/components/FormLandingPage.tsx
vaszig: I agree!I had it like this at the beginning and I got warnings from the linter, that's why I changed it.
Maybe because I didn't cast the variable. I will give it one more try.

https://github.com/open-formulieren/open-forms-sdk/compare/c1cb3c383cfa9291ef9ee2b4bb0863aa163c14ea..c249c72aa54cdddd84d805183d49760cbfd58fdd

src/components/IntroductionPage/index.tsx
sergei-maertens: I didn't mention it explicitly, but I wouldn't support the introduction page flow for the single-page/single-step form either. We already skip the start page with introduction text.

I did describe the proposed URL routes in the frontend with just having two routes essentially:

* `/` -> the step itself
* `/confirmation` (or equivalent, would have to check) for the confirmation page

So, given that the use case here is embedded forms and pretty URLs etc. aren't relevant either, I'd even omit the `slug` from the URL, unless that makes things more complicated

src/components/FormLandingPage.tsx
```diff
@@ -4,9 +4,16 @@ import useFormContext from '@/hooks/useFormContext';
...
   const {preserveQueryParams} = useQueryParams();
-  const startPageUrl = introductionPageContent ? 'introductie' : 'startpagina';
+
+  let startPageUrl: string;
+  if (type === 'single_page') {
+    // single page forms have only one form step
+    startPageUrl = introductionPageContent ? 'sp/introductie' : `sp/${steps[0].slug}`;
+  } else {
+    startPageUrl = introductionPageContent ? 'introductie' : 'startpagina';
+  }
```
sergei-maertens, line 15: see earlier remark about the introduction page :-)

src/routes/single-page-form.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+const singlePageRoutes: RouteObject[] = [
+  {
+    index: true,
+    element: <FormLandingPage />,
+  },
+  {
+    path: 'introductie',
+    element: <IntroductionPage />,
+  },
```
sergei-maertens: see earlier remark - these routes can probably be simplified a lot

sergei-maertens (changes requested):

vaszig:
About the `/confirmation` I saw that everywhere we use the dutch word, should we stick to the english here?Because in the epic itself this is also mentioned as `/bevestiging`

sergei-maertens:
> About the `/confirmation` I saw that everywhere we use the dutch word, should we stick to the english here?Because in the epic itself this is also mentioned as `/bevestiging`

Dutch please!

I haven't found it worthwhile yet to try to translate URLs :P

vaszig:
One more change because we changed the name of `single_page` to `single_step` in the backend.

sergei-maertens (approved):
Approving it, but please don't merge it until the backend PR for the API endoint is merged otherwise end to end tests will break in the backend ;)

sergei-maertens:
Backend PR is merged, let's go :)

PR #963, opened 2026-04-03
Remove formiojs / legacy renderer

sergei-maertens:
Part of open-formulieren/open-forms#6164

It's happening GIF


vaszig (approved):
I don't see something weird and since everything is green I will hit the button.

PR #965, opened 2026-04-09
Ensure list comparisons in json logic match backend behavior

### LINKED ISSUE #6166: List comparison in JSON logic causes form crash (frontend logic evaluation)
Discovered while testing with a large ALO form (1009 on the test env). Error in sentry: 467632.

This form has the following logic trigger which causes the crash: `{"==": [{"var": "partnercol"}, []]}`, where 'partnercol' is a user-defined variable with prefilled partners.

There are (at least) two things going wrong here:
1. When the trigger is partially evaluated for execution in the frontend, the variable expression is substituted with the prefilled data from the user-defined variable. Resulting in: 
   ```javascript
   {
      "==": [
       [
          {
                "affixes": "den",
                "bsn": "999994542",
                "dateOfBirth": "1968-03-18",
                "dateOfBirthPrecision": "date",
                "firstNames": "Gerrit",
                "initials": "G.",
                "lastName": "Braber",
            }
       ],
       []
     ]
   }
   ```
   This expression will be passed to `jsonLogic`, because it is fully resolved. However, this causes a crash as it tries to parse the object as a JSON logic expression: `ValueError: Unrecognized operation bsn`.
2. Solving the first problem does not solve the issue completely, because this expression cannot be evaluated properly in the frontend either. On the [sandbox](https://json-logic.github.io/json-logic-engine/docs/logic#equals) `{"==": [[], []]}` evaluates to `false`, which suggests a simple identity check, and not a deep equality check. (Note that in the backend, this expression evaluates to `True` instead.) Adding a unit test shows different behavior, though:
   ```javascript
   const expression = {'==': [[], []]};
   const data = {};
   const engine = new LogicEngine();
   const result = engine.run(expression, data); // Throws a "NaN"
   ```
   Furthermore, it should also be possible to evaluate the prefilled logic expressions in the frontend as well.

viktorvanwijk:
Partly closes open-formulieren/open-forms#6166

### DIFF
--- a/src/logic/json-logic/extensions/comparators.ts
+++ b/src/logic/json-logic/extensions/comparators.ts
@@ -1,5 +1,8 @@
 import {defaultMethods} from 'json-logic-engine';
+import type {LogicEngine} from 'json-logic-engine';
+import {isEqual} from 'lodash';
 
+import {INVALID_ARGUMENTS} from './constants';
 import type {JsonLogicEngineMethod} from './types';
 
 /**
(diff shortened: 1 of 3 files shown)


src/logic/json-logic/extensions/comparators.ts
```diff
@@ -1,4 +1,6 @@
 import {defaultMethods} from 'json-logic-engine';
+import {type LogicEngine} from 'json-logic-engine';
```
sergei-maertens:
```suggestion
import type {LogicEngine} from 'json-logic-engine';
```
unfortunately the linter can't be configured to report his 😬

src/logic/json-logic/extensions/comparators.ts
```diff
@@ -39,18 +41,59 @@ export const customLessThanEquals: JsonLogicEngineMethod = comparatorOverrideFac
...
+
+  return engine.run(logic, data, {above});
+}
+
 export const customEquals: JsonLogicEngineMethod = comparatorOverrideFactory(
-  defaultMethods['=='].method
+  (args, context, above, engine): boolean => {
+    // While the original implementation in json-logic-engine allows for more than two arguments,
+    // the backend does not, so throw whenever it is not two.
+    if (!Array.isArray(args) || args.length !== 2) throw 'Invalid arguments';
```
sergei-maertens: You can use `INVALID_ARGUMENTS` from the `./constants` module

sergei-maertens (changes requested):
Some cosmetic changes

Shouldn't this PR also update the org-github submodule for the updated shared tests?

src/logic/json-logic/extensions/comparators.ts
viktorvanwijk: Damn, and I even created a separate import for the type :sweat_smile:

sergei-maertens (approved):

PR #967, opened 2026-04-10
[OF#6175] Implement new form step for a single page form

### LINKED ISSUE #6175: SDK: Implement the form steps for a single page form
The new form type added (single page form) requires a new flow for the form steps due to not creating a submission at the beginning and having exactly one form step.

This should have one form step and the confirmation page.

vaszig:
Closes open-formulieren/open-forms#6175

### DIFF
--- a/src/api-mocks/forms.ts
+++ b/src/api-mocks/forms.ts
@@ -79,7 +79,7 @@ export const SINGLE_STEP_FORM_DEFAULTS = {
   name: 'Mock single step form',
   slug: 'mock-single-step',
   type: 'single_step',
-  url: `${BASE_URL}forms/mock-single-step`,
+  url: `${BASE_URL}forms/05c6e422-7f39-4366-ae7a-1da642e1f5ff`,
   loginRequired: false,
   loginOptions: [],
   showProgressIndicator: false,
@@ -103,7 +103,7 @@ export const SINGLE_STEP_FORM_DEFAULTS = {
   },
   steps: [
     {
-      uuid: '8ae216fc-53d9-4e22-b803-54773b5f9b0c',
+      uuid: '3ad1734b-d095-4da0-8b7e-9f6900ffda17',
       slug: 'step-1',
       formDefinition: 'Step 1',
       index: 0,
@@ -112,7 +112,7 @@ export const SINGLE_STEP_FORM_DEFAULTS = {
         saveText: {resolved: 'Save', value: ''},
         nextText: {resolved: 'Next', value: ''},
       },
-      url: `${BASE_URL}forms/05c6e422-7f39-4366-ae7a-1da642e1f5ff/steps/8ae216fc-53d9-4e22-b803-54773b5f9b0c`,
+      url: `${BASE_URL}forms/05c6e422-7f39-4366-ae7a-1da642e1f5ff/steps/3ad1734b-d095-4da0-8b7e-9f6900ffda17`,
     },
   ],
   introductionPageContent: '',
@@ -136,11 +136,9 @@ const FORM_STEP_DETAILS_DEFAULT = {
   configuration: DEFAULT_FORMIO_CONFIGURATION,
   formDefinition: `${BASE_URL}/form-definitions/65e2a836-cd31-45b0-adfd-5fc68afe2038`,
   name: 'Single step',
-  internalName: '',
-  url: `${BASE_URL}/forms/05c6e422-7f39-4366-ae7a-1da642e1f5ff/steps/8ae216fc-53d9-4e22-b803-54773b5f9b0c`,
+  url: `${BASE_URL}/forms/05c6e422-7f39-4366-ae7a-1da642e1f5ff/steps/3ad1734b-d095-4da0-8b7e-9f6900ffda17`,
   isApplicable: true,
   loginRequired: false,
-  isReusable: false,
   literals: {
     previousText: {
       resolved: 'Previous page',
@@ -198,10 +196,7 @@ export const mockFormGet = (formDetail = buildForm(), once = false) =>
     {once: once}
   );
 
-export const mockFormStepGet = (
-  formStepDetails: FormStep = FORM_STEP_DETAILS_DEFAULT,
-  status: 200 | 201 = 200
-) =>
+export const mockFormStepGet = (formStepDetails: FormStep = FORM_STEP_DETAILS_DEFAULT) =>
   http.get(`${BASE_URL}forms/:uuid/steps/:stepUuid`, () => {
-    return HttpResponse.json(formStepDetails, {status: status});
+    return HttpResponse.json(formStepDetails, {status: 200});
   });
--- a/src/api-mocks/forms.ts
+++ b/src/api-mocks/forms.ts
@@ -2,9 +2,10 @@ import {produce} from 'immer';
 import {HttpResponse, http} from 'msw';
 
 import {PRIVACY_POLICY_ACCEPTED} from '@/components/SummaryConfirmation/mocks';
-import type {Form} from '@/data/forms';
+import type {Form, FormStep} from '@/data/forms';
 
 import {BASE_URL, getDefaultFactory} from './base';
+import {DEFAULT_FORMIO_CONFIGURATION} from './submissions';
 
 export const FORM_DEFAULTS = {
   uuid: 'e450890a-4166-410e-8d64-0a54ad30ba01',
@@ -73,6 +74,101 @@ export const FORM_DEFAULTS = {
   communicationPreferencesPortalUrl: '',
 } satisfies Form;
 
+export const SINGLE_STEP_FORM_DEFAULTS = {
+  uuid: '05c6e422-7f39-4366-ae7a-1da642e1f5ff',
+  name: 'Mock single step form',
+  slug: 'mock-single-step',
+  type: 'single_step',
+  url: `${BASE_URL}forms/mock-single-step`,
+  loginRequired: false,
+  loginOptions: [],
+  showProgressIndicator: false,
+  showSummaryProgress: false,
+  maintenanceMode: false,
+  active: true,
+  submissionAllowed: 'yes',
+  submissionLimitReached: false,
+  suspensionAllowed: false,
+  sendConfirmationEmail: true,
+  displayMainWebsiteLink: true,
+  submissionStatementsConfiguration: [PRIVACY_POLICY_ACCEPTED],
+  appointmentOptions: {
+    supportsMultipleProducts: null,
+  },
+  literals: {
+    beginText: {resolved: 'Begin', value: ''},
+    changeText: {resolved: 'Change', value: ''},
+    confirmText: {resolved: 'Confirm', value: ''},
+    previousText: {resolved: 'Previous', value: ''},
+  },
+  steps: [
+    {
+      uuid: '8ae216fc-53d9-4e22-b803-54773b5f9b0c',
+      slug: 'step-1',
+      formDefinition: 'Step 1',
+      index: 0,
+      literals: {
+        previousText: {resolved: 'Previous', value: ''},
+        saveText: {resolved: 'Save', value: ''},
+        nextText: {resolved: 'Next', value: ''},
+      },
+      url: `${BASE_URL}forms/05c6e422-7f39-4366-ae7a-1da642e1f5ff/steps/8ae216fc-53d9-4e22-b803-54773b5f9b0c`,
+    },
+  ],
+  introductionPageContent: '',
+  explanationTemplate: '',
+  requiredFieldsWithAsterisk: true,
+  resumeLinkLifetime: 7,
+  autoLoginAuthenticationBackend: '',
+  translationEnabled: false,
+  hideNonApplicableSteps: false,
+  cosignLoginOptions: [],
+  cosignHasLinkInEmail: false,
+  paymentRequired: false,
+  submissionReportDownloadLinkTitle: '',
+  communicationPreferencesPortalUrl: '',
+} satisfies Form;
+
+const FORM_STEP_DETAILS_DEFAULT = {
+  uuid: '3ad1734b-d095-4da0-8b7e-9f6900ffda17',
+  index: 0,
+  slug: 'single-step',
+  configuration: DEFAULT_FORMIO_CONFIGURATION,
+  formDefinition: `${BASE_URL}/form-definitions/65e2a836-cd31-45b0-adfd-5fc68afe2038`,
+  name: 'Single step',
+  internalName: '',
+  url: `${BASE_URL}/forms/05c6e422-7f39-4366-ae7a-1da642e1f5ff/steps/8ae216fc-53d9-4e22-b803-54773b5f9b0c`,
+  isApplicable: true,
+  loginRequired: false,
+  isReusable: false,
+  literals: {
+    previousText: {
+      resolved: 'Previous page',
+      value: '',
+    },
+    saveText: {
+      resolved: 'Save current information',
+      value: '',
+    },
+    nextText: {
+      resolved: 'Next',
+      value: '',
+    },
+  },
+  translations: {
+    nl: {
+      previousText: '',
+      saveText: '',
+      nextText: '',
+    },
+    en: {
+      previousText: '',
+      saveText: '',
+      nextText: '',
+    },
+  },
+};
+
 /**
  * Return a form object as if it would be returned from the form detail API endpoint.
  * @param  overrides Key-value mapping with overrides from the defaults. These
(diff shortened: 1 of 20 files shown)


src/api-mocks/forms.ts
```diff
@@ -101,3 +197,11 @@ export const mockFormGet = (formDetail = buildForm(), once = false) =>
     },
     {once: once}
   );
+
+export const mockFormStepGet = (
+  formStepDetails: FormStep = FORM_STEP_DETAILS_DEFAULT,
+  status: 200 | 201 = 200
```
sergei-maertens: 201 status for a GET call can not happen. I'm not sure why this is a parameter, since for anything else than an HTTP 200, the shape of the response data is going to be different than the `FormStep` type anyway.

src/api-mocks/submissions.ts
```diff
@@ -203,10 +243,10 @@ export const mockSubmissionSummaryGet = () =>
     );
   });
 
-export const mockSubmissionCompletePost = () =>
+export const mockSubmissionCompletePost = (submissionDetails = SUBMISSION_DETAILS) =>
```
sergei-maertens:
```suggestion
export const mockSubmissionCompletePost = (submissionId: number = SUBMISSION_DETAILS.id) =>
```
makes more sense

src/components/FormStart/tests.spec.tsx
sergei-maertens, line 1: Nice catch

src/components/FormStep/Progress.tsx
```diff
@@ -14,8 +14,14 @@ export interface ProgressProps {
...
   if (!form.showSummaryProgress) return null;
-  const applicableSteps = submission.steps.filter(step => step.isApplicable);
-  const currentSubmissionStepIndex = applicableSteps.indexOf(currentStep);
+
+  // single page forms do not use submission (we create the submission when the form is
+  // submitted) and have only one step
+  const applicableSteps = submission
+    ? submission.steps.filter(step => step.isApplicable)
+    : [currentStep];
+  const currentSubmissionStepIndex = submission ? applicableSteps.indexOf(currentStep) : 0;
```
sergei-maertens: This is always going to be step one-of-one, right? Can't we exit early?

```tsx
if (submission === null) return null;
```

effectively disabling this for single-step forms.

src/data/submissions.ts
sergei-maertens, line 1: These are just formatting changes, right?

src/data/forms.ts
```diff
@@ -37,6 +39,31 @@ export interface MinimalFormStep {
...
+  loginRequired: boolean;
+  isReusable: boolean;
+  literals: {
+    previousText: ButtonText;
+    saveText: ButtonText;
+    nextText: ButtonText;
+  };
+  // TODO
+  // fix type
+  translations: object;
```
sergei-maertens: if we don't use this, you can omit it :-) afaik the backend takes care of returning the active translations in the main properties already.

src/data/forms.ts
```diff
@@ -37,6 +39,31 @@ export interface MinimalFormStep {
...
+ * @see `#/components/schemas/MinimalFormStep` in the API spec.
+ */
+export interface FormStep {
+  uuid: string;
+  index: number;
+  slug?: string | null;
+  configuration: FormioConfiguration;
+  formDefinition: string;
+  name: string;
+  internalName?: string;
```
sergei-maertens: this should not be exposed to the public frontend. If it is, then that's a bug in the backend.


```suggestion
```

src/data/forms.ts
```diff
@@ -37,6 +39,31 @@ export interface MinimalFormStep {
...
+  url: string;
+  isApplicable: boolean;
```
sergei-maertens:
```suggestion
  isApplicable?: boolean;
```

src/data/forms.ts
```diff
@@ -37,6 +39,31 @@ export interface MinimalFormStep {
...
+  loginRequired: boolean;
+  isReusable: boolean;
```
sergei-maertens:
```suggestion
```
should also be exposed only to staff users

src/data/forms.ts
sergei-maertens: Checking the API spec, this should only be exposed to staff users even, so you can remove it and fix the backend serializer.

src/data/forms.ts
```diff
@@ -37,6 +39,31 @@ export interface MinimalFormStep {
...
+  literals: {
```
sergei-maertens:
```suggestion
  literals?: {
```
can be absent according to the API spec

src/components/FormStep/SingleFormStepNewRenderer.tsx
sergei-maertens, line 1: I'd like to see some unit tests (not storybook) that prove you can submit a single step form from start to end

src/components/FormStep/SingleFormStepNewRenderer.tsx
```diff
@@ -0,0 +1,271 @@
+import {FormioForm, LoadingIndicator} from '@open-formulieren/formio-renderer';
+import type {FormStateRef} from '@open-formulieren/formio-renderer/components/FormioForm.js';
+import type {JSONObject} from '@open-formulieren/formio-renderer/types.js';
```
sergei-maertens: I think we can import that from `@open-formulieren/types`, can you check which we use most commonly?

src/components/FormStep/SingleFormStepNewRenderer.tsx
```diff
@@ -0,0 +1,271 @@
...
+  type Submission,
+  type SubmissionCompleteBody,
+  completeSubmission,
+  createSubmission,
+} from '@/data/submissions';
+import {ValidationError} from '@/errors';
+import useFormContext from '@/hooks/useFormContext';
+
+import StatementCheckbox from '../StatementCheckboxes/StatementCheckbox';
+import ValidationErrors from '../Summary/ValidationErrors';
```
sergei-maertens: can you make those `@/...` imports?

src/components/FormStep/SingleFormStepNewRenderer.tsx
```diff
@@ -0,0 +1,271 @@
...
+  const formStep = form.steps[0];
+
+  // retrieve the form step in order to obtain the components
+  const {
+    loading: formStepLoading,
+    value: formStepData,
+    error,
+  } = useAsync(() => {
+    return get<FormStep>(formStep.url);
+  }, [formStep.uuid]);
```
sergei-maertens:
```suggestion
  // single page forms have exactly one step
  const formStepUrl = form.steps[0].url;

  // retrieve the form step in order to obtain the components
  const {
    loading: formStepLoading,
    value: formStep,
    error,
  } = useAsync(async () => await get<FormStep>(formStepUrl), [formStepUrl]);
```
I think this would be equivalent and uses less intermediate variables while being a tad more readable

In particular, I don't like the non-static reference in the useAsync dependency array, and that you put `uuid` in there while `url` is used for the lookup. It effectively works and means the same thing, but we can make this a bit more robust.

src/components/FormStep/SingleFormStepNewRenderer.tsx
```diff
@@ -0,0 +1,271 @@
...
+
+  const isLoading = formStepLoading || navigationState !== 'idle';
+  const components = formStepData?.configuration.components ?? [];
+
+  const formRef = useRef<FormStateRef>(null);
+  // keep track of the current values in a mutable ref to avoid re-renders when values
+  // change, which take time and can produce 'lag'. We typically only need to read the
+  // values in callbacks at this component level (in the tree), while the `FormioForm`
+  // component itself is responsible for managing the state.
+  const valuesRef = useRef<JSONObject | null>(null);
```
sergei-maertens: Having these hooks *after* the error handling (`throw error`) is unsafe in React, because different renders can result in different amount of hooks being called, and that's a big no-no in React.

I'd move this refs up as high as possible.

src/components/FormStep/SingleFormStepNewRenderer.tsx
```diff
@@ -0,0 +1,271 @@
...
+      if (e instanceof ValidationError) {
+        const {initialErrors} = e.asFormikProps();
+        setSubmitErrors(initialErrors);
+      } else {
+        setSubmitErrors(e.message as string);
+      }
+      return;
+    }
+    return statusUrl;
+  };
```
sergei-maertens: I'd put this all in a single `onSubmit` callback if the individual parts aren't re-used (and it doesn't seem that way). This way, we're creating less function references in each renderer and declutter the code.

Some additional remarks:

1. We can hardcode the submission creation to be anonymous, the options argument is not necessary. Login is not supported in these flows. You can do a sanity check  like `if (form.loginOptions.length !== 0) throw new Error('Login on single step forms is not supported.');`
2. `onSubmitSaveStepData` doesn't need the check for empty submission if I read your types correctly - it will never be falsy.
3. I notice that the step data can raise validation errors, while the submission create will likely have succeeded. There will need to be some code that checks the submission from the state and only creates it in the `onSubmit` handler if the submission from the state is `null`/empty, otherwise we'll keep on creating new submissions for each submit attempt when they're handling validation errors from the server.
4. `onSubmitCompleteSubmission` and the navigate to overview page can also be inlined in the single `onSubmit` handler

src/components/FormStep/SingleFormStepNewRenderer.tsx
```diff
@@ -0,0 +1,271 @@
...
+        {form.submissionLimitReached && <FormMaximumSubmissionsError />}
+        {form.explanationTemplate && (
+          <Body
+            component="div"
+            modifiers={['wysiwyg']}
+            dangerouslySetInnerHTML={{__html: form.explanationTemplate}}
+          />
+        )}
+
+        <PreviousLink to={'/sp'} position="start" />
```
sergei-maertens: does this make sense? it navigates back to the current page, so I think this should be left out?

src/components/FormStep/SingleFormStepNewRenderer.tsx
```diff
@@ -0,0 +1,271 @@
...
+        <Progress form={form} submission={null} currentStep={formStep} />
```
sergei-maertens: I think this component also doesn't make much sense for single step forms

src/components/FormStep/SingleFormStepNewRenderer.tsx
```diff
@@ -0,0 +1,271 @@
...
+        <CardTitle title={'jd'} headingType="subtitle" padded />
```
sergei-maertens: what's this `title` prop value?

src/components/FormStep/SingleFormStepNewRenderer.tsx
```diff
@@ -0,0 +1,271 @@
...
+            </div>
+
+            <FormStepNavigation
+              submissionAllowed="yes"
+              isLastStep
+              stepSubmissionAllowed
+              hideAbortButton
+              // for now the logic call is not made, see if there is a case where this
+              // would be needed
+              isCheckingLogic={false}
```
sergei-maertens, line 248: server side logic calls can't be done at all because you need a submission in the session for that

I think this requires an assertion or update in the types that forces `requireBackendLogicEvaluation` to be `false` (new logic evaluation). So if we do support it in the future, it must be limited to rules that can be evaluated client-side.

src/scss/components/_single-step.scss
sergei-maertens, line 1: Instead of adding a new component and hardcoded CSS, I think think you can re-use the class name `openforms-form-field-container` from the renderer instead, which uses flexbox layout and a row-gap that can be configured with design tokens :)

sergei-maertens (changes requested):

src/api-mocks/forms.ts
vaszig: Copy paste error..fixed that.

src/data/submissions.ts
vaszig: I am using it in the `SingleFormStepNewRenderer` https://github.com/open-formulieren/open-forms-sdk/pull/967/changes/BASE..bd9bfebef2c3d58b16969ff9cd1994cff6f7e6cd#diff-de2b720e309bf4d09885fbcb259bbc967e5e2f43dd566481bc202d6efb7fd5dfR50

src/data/forms.ts
vaszig: It is present in the public fields, I guess I have to remove it?
https://github.com/open-formulieren/open-forms/blob/main/src/openforms/forms/api/serializers/form_step.py#L122

src/data/forms.ts
vaszig: The same here:
https://github.com/open-formulieren/open-forms/blob/main/src/openforms/forms/api/serializers/form_step.py#L126

src/components/FormStep/SingleFormStepNewRenderer.tsx
vaszig: We mostly use the types repo for this import so I changed that in three more places to use only the types.

src/components/FormStep/SingleFormStepNewRenderer.tsx
vaszig: Which types do you mean exactly?Just to be sure..

src/scss/components/_single-step.scss
vaszig: I tried that but with no effect because I saw that the css class used is `.openforms-statement-checkboxes` so I would rather add a modifier there especially for the single step forms?
https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/StatementCheckboxes/StatementCheckbox.scss#L11

src/data/forms.ts
sergei-maertens: Yes :)

src/components/FormStep/SingleFormStepNewRenderer.tsx
sergei-maertens: requiresBackendLogic type could be false instead of boolean, but I'm not sure how special the single step form types are compared to normal

src/scss/components/_single-step.scss
sergei-maertens: You can put a wrapper div with the classname around the elements to properly space them. Semantically this is the correct approach.

I want to avoid hardcoding layout spacing but also avoid introducing even more design tokens if there are existing mechanisms that already can provide the functionality

src/components/FormStep/tests.spec.tsx
```diff
@@ -65,37 +54,52 @@ const Wrap: React.FC<WrapperProps> = ({
...
-test('Start single step form', async () => {
-  http.all('*', ({request}) => {
-    console.log('REQUEST:', request.method, request.url);
-    return HttpResponse.json({});
-  });
-  let startSingleStepFormRequest: Request | undefined;
+test('Single step form happy flow', async () => {
+  mswWorker.use(
+    mockFormGet(FORM_DATA),
+    mockAnalyticsToolConfigGet(),
```
sergei-maertens, line 70: This one shouldn't be necessary if the `AnalyticsToolsConfigProvider` is not rendered, right?

src/components/FormStep/tests.spec.tsx
```diff
@@ -65,37 +54,52 @@ const Wrap: React.FC<WrapperProps> = ({
...
+
+  const nextButton = screen.getByRole('button', {name: 'Next'});
+
+  await expect.element(nextButton).toBeEnabled();
+  await nextButton.click();
+  await expect.element(screen.getByText(/Confirmation/)).toBeVisible();
+});
+
+test('Single step form with missing statement checkbox', async () => {
+  mswWorker.use(mockFormGet(FORM_DATA), mockAnalyticsToolConfigGet(), mockFormStepGet());
```
sergei-maertens, line 94: here too I'd expect the analytics mock not to be necessary

src/components/StatementCheckboxes/StatementCheckbox.scss
```diff
@@ -12,4 +12,8 @@
   display: flex;
   flex-direction: column;
   gap: var(--of-privacy-checkbox-spacing, 8px);
+
+  &--single-step-form {
+    margin-block-start: var(--of-form-field-container-gap, 24px);
+  }
```
sergei-maertens, line 19: Can you use the bem helpers?

```scss

...
.openforms-statement-checkboxes {
  @include bem.modifier('single-step-form') { ... }
}
```

so that the strict selectors are produced?

Your code now emits `.openforms-statement-checkboxes--single-step-form`, with the helper, it becomes `.openforms-statement-checkboxes.openforms-statement-checkboxes--single-step-form` which is more correct :)

sergei-maertens (changes requested):

src/components/FormStep/tests.spec.tsx
vaszig: Even if I do not render this here, it's rendered by the `SingleStepForm` component. When I removed it I was getting an error.

src/components/FormStep/tests.spec.tsx
vaszig: The same as above.

src/components/FormStep/tests.spec.tsx
sergei-maertens: Ah clear!

sergei-maertens (approved):

PR #968, opened 2026-04-13
Fix default hidden component inside editgrid with new logic evaluation causing crash

### LINKED ISSUE #6183: Problems with editgrids and new (frontend) logic evaluation
When there is a backend logic rule that affects the visibility of an editgrid (or its parent) two problems occur:

- Editgrid items get cleared immediately upon saving
- If the editgrid contains a component that is hidden by default, the form crashes completely

Note that it doesn't matter whether the rule was triggered or not.

viktorvanwijk:
Closes open-formulieren/open-forms#6183

~TODO: requires renderer to be updated to new version!~


viktorvanwijk:
The updated `package-lock.json` is the result of running `npm i --legacy-peer-deps` locally. Running just `npm i` results in a dependency conflict with `vega` being on version 6.2.0, while it is defined as "^5.25.0" in @utrecht/component-library-react@13.1.0.

New errors occur in CI, though. It appears to be originating from the fact that the pinned versions for the individual Utrecht packages in the renderer, differ from the versions that the overall utrecht component react library require. Running `npm ci --legacy-peer-deps` locally seems to fix the issue. Is the solution adding this flag to `.npmrc`? Note that this flag was removed last week though with the removal of formio-js (427daf9203d7f4f17d62e0840987ddefcebf8d25)

EDIT: I can also get it to work by installing the following packages manually, which are a subset of the packages added to the renderer recently ([c2a13391](https://github.com/open-formulieren/formio-renderer/pull/310/commits/e6c4d8f87cdb2d9fe4b33cd2fe84851790c58fec))
"@utrecht/fieldset-react": "^1.0.12",
"@utrecht/form-field-description-react": "^1.0.11",
"@utrecht/form-field-react": "^1.0.12",
"@utrecht/form-label-react": "^1.0.11",
"@utrecht/link-react": "^1.0.11",
"@utrecht/radio-button-react": "^1.0.12",
"@utrecht/textbox-react": "^1.0.15",

sergei-maertens:
Regression test looks fine, will take a look into the packaging issues asap.

sergei-maertens:
@viktorvanwijk I bumped the utrecht component library to the latest minor/patch release and the `--legacy-peer-deps` approach works again, I don't know for how long :stuck_out_tongue: but that should unblock this PR if chromatic is happy.

sergei-maertens:
If I understand it correctly, this should be backported to at least 3.5.x. Does it affect 3.4.x too?

sergei-maertens (approved):

viktorvanwijk:
> If I understand it correctly, this should be backported to at least 3.5.x. Does it affect 3.4.x too?

Ah yeah, forgot to add the label... Should be just 3.5.x, as these crashes occurred with logic rules only.

PR #970, opened 2026-04-14
Upgrade build tooling

### LINKED ISSUE #926: Build tooling upgrade
- [ ] Storybook (to latest, 10?)
- [ ] NodeJS (to 24 LTS possible?)
- [ ] Typescript to 5.9?

sergei-maertens:
Partly closes #926


.nvmrc
```diff
@@ -1 +1 @@
-20.19
+24.14
```
annashamray, line 1: DO you need to update "target" in tsconfig?

src/components/EmailVerification/EmailVerificationModal.tsx
```diff
@@ -31,11 +31,8 @@ const EmailVerificationModal: React.FC<EmailVerificationModalProps> = ({
   emailAddress,
 }) => {
   const [isVerified, setIsVerified] = useState(false);
-
-  useEffect(() => {
-    // once the modal closes, reset the internal state
-    if (!isOpen && isVerified) setIsVerified(false);
-  }, [isOpen, isVerified, setIsVerified]);
+  // once the modal closes, reset the internal state
+  if (!isOpen && isVerified) setIsVerified(false);
```
annashamray, line 35: Could you please explain why you removed `useEffect` here?

annashamray (changes requested):

.nvmrc
sergei-maertens: No, that target determines how far back the bundler/TS need to transpile. With 2021 we say: compile for browsers that implement the ES specification from 2021. We're pretty conservative here because people from poorer backgrounds may be using older (mobile) devices and don't always have the latest browser versions.

src/components/EmailVerification/EmailVerificationModal.tsx
sergei-maertens: The linter pointed out it is not necessary :-) See https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

annashamray: TIL!

.nvmrc
annashamray: Thanks for the explanation!

annashamray (approved):

PR #971, opened 2026-04-20
Switch (unit) tests to Vitest browser mode

### LINKED ISSUE #926: Build tooling upgrade

sergei-maertens:
Closes #926

Replaces testing-library/jsdom with vitest browser mode in a real browser and updates all assertions in the tests.

Some mocks have been obsoleted thanks to this.


.github/workflows/ci.yml
```diff
@@ -80,7 +80,10 @@ jobs:
           cache: npm
 
       - name: Install dependencies
-        run: npm ci
+        run: |
+          npm ci
+          npm run build:design-tokens
```
vaszig, line 85: Did we add this because now we have a real browser?

src/components/appointments/CreateAppointment/CreateAppointment.spec.tsx
```diff
@@ -68,72 +65,71 @@ afterEach(() => {
 });
 
 describe('Create appointment session expiration', () => {
-  it('resets the session storage/local state', {timeout: 15_000}, async () => {
-    const user = userEvent.setup({delay: null});
-    mswServer.use(
+  test('resets the session storage/local state', {timeout: 15_000}, async () => {
```
vaszig, line 68: What's the diff between `it` and `test`?Why do we update this in all the tests?

src/components/appointments/cancel/CancelAppointment.integration.spec.tsx
```diff
@@ -42,14 +43,12 @@ const Wrapper = () => {
...
+  test('renders the correct page for a cancel route', async () => {
+    const screen = await render(<Wrapper />);
 
-    const textbox = await screen.findByRole<HTMLInputElement>('textbox', {
-      name: 'Your email address',
-    });
-    expect(textbox).toBeVisible();
-    expect(textbox.name).toBe('email');
-    expect(screen.getByRole('button')).toBeVisible();
+    const textbox = screen.getByRole('textbox', {name: 'Your email address'});
```
vaszig, line 49: Do we now use synchronous because we await the render?

vaszig (approved):

.github/workflows/ci.yml
sergei-maertens: Yes! The icon was not actually visible without including the CSS, and caused tests to fail :)

src/components/appointments/CreateAppointment/CreateAppointment.spec.tsx
sergei-maertens: they're aliases, but now it's made consistent.

`test` is more obvious to people new to frontend testing and the preferred approach for that reason.

src/components/appointments/cancel/CancelAppointment.integration.spec.tsx
sergei-maertens: this is the difference between queries that return an element vs queries that return a locator!

`textbox` is not an HTML element yet, it only resolves through `await expect.element(...).toSomething()`

PR #973, opened 2026-04-22
Bump @open-formulieren/formio-renderer to 1.5.2

viktorvanwijk:


viktorvanwijk:
This update is necessary to backport fixes in the renderer to the 3.4.x stable branch. Looking at the bright red CI, though, I think it is wise to create a 1.1.x stable branch in the renderer instead, and update to that version on the 3.4.x branch.

PR #974, opened 2026-04-22
Switch to v8 coverage

### LINKED ISSUE #6099: Test edge cases with new logic evaluation
**Step (not) applicable action**

The fix for #5996 included executing rules with a "step (not) applicable" action on the steps on and after the resolved step (which is based on the input variables of the rule). This is still flawed unfortunately. Consider the following case:

* Form with three steps
* Logic rule that triggers when checkbox on step 2 is checked, and marks step 3 as not applicable -> will be executed from step 2 onward
* User submits step 2 with the checkbox checked and reaches the overview, because step 3 will not be applicable
* Navigates back to step 1 to change something -> logic rule is not executed on step 1 -> step 3 will be applicable again

Ideally we would implement #6035, but it's not straightforward. Another solution might be to execute the rule on ALL steps. It can be executed in the frontend, so performance shouldn't suffer too much, but it's not ideal

**`isCompleted` flag**
There was some doubt whether this flag would be updated properly upon step submission, but it seems to be going right because the submission is also refreshed (with the new renderer) upon step completion. For the old renderer, the submission is updated by performing another check logic call after step completion.

sergei-maertens:
More build tooling improvements - making the test setup more consistent with other repositories.

Closes open-formulieren/open-forms#6099, includes a bonus fix (that should be backported)!


src/components/FormStep/FormStepNewRenderer.tsx
```diff
@@ -151,6 +151,7 @@ const FormStepNewRenderer: React.FC = () => {
     if (requireBackendEvaluation) {
       scheduleLogicCheck();
     } else {
+      console.log(valuesRef.current);
```
viktorvanwijk, line 154: :)

sergei-maertens: FFS eslint... you had one job :sweat_smile:

PR #975, opened 2026-04-22
Drop UMD bundle support

### LINKED ISSUE #6164: :boom: 4.0 - overview of deprecated functionalities removal
Scavenged from grepping on `DeprecationWarning` in the codebases, milestone 4.0 issues and post-its.

TODO: create sub-issues for the tasks listed here

* [x] Remove formio.js from the SDK & support for old renderer / associated feature flag https://github.com/open-formulieren/open-forms-sdk/pull/963
* ~~Remove (through forced new renderer) support for `textfield` address autofill/derivation -> sort out a migration path with address NL~~ (See #6239)
* [ ] Remove legacy logic evaluation & associated feature flag 
    - SDK: https://github.com/open-formulieren/open-forms-sdk/pull/981
    - backend: #6261 
* [x] Remove JCC SOAP appointment plugin (@vaszig)
* [x] Remove Elastic-APM support (#6257)
* [x] Simplify registration workflow (ZGW APIs -> change default to use OF-generated public reference) (https://github.com/open-formulieren/open-forms/issues/6291)
* [ ] Check what we can do with prefill data validation to mark as authentic/prevent tampering without relying on the form field being marked as read-only
* [x] Delete `authentication_success` signal - it's unused afaik (#6238)
* [x] Switch the default/legacy SDK bundles being offered - there are now compatibility URLs
    - SDK part: https://github.com/open-formulieren/open-forms-sdk/pull/975
    - Backend part: #6215
* [x] Remove formio registry `build_serializer_field` fallback (#6154)
* [ ] StUF-ZDS: prefer branch number from auth data rather than form variable (@vaszig)
* [x] `CosignV2Data` -> make `cosign_date` required (#6238)
* [x] Remove formio translations endpoints in the backend (if not used by our own backend JS!) https://github.com/open-formulieren/open-forms/pull/6237
* [x] Remove fallbacks in backtotop-link CSS (done in f2e2121a162234d8453cbfbc4a136cf86aec7f52)

Explicitly *NOT* yet scheduled for removal because it's still used too often:

* cosign v1
* Objects API registration options v1 (template based)

sergei-maertens:
Partly closes open-formulieren/open-forms#6164

This removes the fat UMD bundle from our build pipeline and artifacts. It has been deprecated a while ago. We are now ESM-only, and offer a "source" build and bundle variant for inclusion in the backend.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -1,7 +1,6 @@
 // eslint-disable-next-line @typescript-eslint/triple-slash-reference
 /// <reference types="vitest/config" />
 import {codecovVitePlugin} from '@codecov/vite-plugin';
-import replace from '@rollup/plugin-replace';
 import {sentryVitePlugin} from '@sentry/vite-plugin';
 import {storybookTest} from '@storybook/addon-vitest/vitest-plugin';
 import react from '@vitejs/plugin-react';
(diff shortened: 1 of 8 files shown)


vite.config.ts
```diff
@@ -75,28 +53,21 @@ const getOutput = (buildTarget: typeof process.env.BUILD_TARGET): OutputOptions
...
-       */
       return {
         dir: `dist/bundles/`,
-        format: 'umd',
-        exports: 'named',
-        name: 'OpenForms',
-        generatedCode: 'es2015',
-        entryFileNames: 'open-forms-sdk.js',
+        format: 'esm',
+        preserveModules: false,
```
viktorvanwijk, line 62: If I understand correctly, for the docker image we want to optimize the bundle so we don't use `preserveModules` and `inlineDynamicImports`. But for the npm package we want to keep the original source, so we use  `preserveModules: true`?

viktorvanwijk (commented):
Some questions :)

I found one more reference to UMD in `README.npm.md` https://github.com/open-formulieren/open-forms-sdk/blob/e6a1b4f861b10720e36efea2d1a4b46eb399cf30/README.npm.md?plain=1#L24

About the bundle size: shouldn't it decrease with this change? I see codecov still lists the UMD build:

Image

sergei-maertens:
> About the bundle size: shouldn't it decrease with this change? I see codecov still lists the UMD build:

:shrug: Codecov bundle analysis seems to be weird and I don't fully understand how this works. I suspect it's reusing the old stats from the main branch if a report is missing? I had the same expectation.

I'll address the README.npm.md reference! Weird that I missed that when grepping :thinking:

vite.config.ts
sergei-maertens: Yes - I basically move the configuration blocks around, it shows a bit weird in the diff. UMD used to be the default, now the esm-bundle block is moved to that default.

But to answer the actual question:

* the bundle artifact does merge the various source files into a (single) bundle so that the client can download all the code in a single HTTP request. We have some dynamic imports here and there that Vite uses as bundle splitting points, so those are the individual chunks you see and that code is lazy loaded when it's actually needed (like the leaflet map, file upload component etc.)
* for the NPM package, we just do minimal transpilation so that another consumer (mostly the NL DS storybooks) can import our SDK code from our package and use plain modules, without having to load all of our code and then they benefit from tree shaking - all our API interaction code/components are not relevant for them. So that's why we preserve the module structure indeed.

There may be a time in the future where we build the SDK in our backend project JS setup and don't need the docker image trick anymore, but that's going to take a while to explore and set up.

viktorvanwijk (approved):

PR #976, opened 2026-04-24
Upgrade to latest renderer 1.1.x patch release

sergei-maertens:
Necessary for the SDK patch release to be bundled in the backend release.


sergei-maertens:
The failing UI tests in Chromatic probably have to do with the double emitted events in the modal. They don't happen in a real browser it seems, and in the main branch there have been test workarounds/patches made. So, I think we can ignore this in chromatic for 3.4.x - it will be end-of-life in a couple of months too.

PR #988, opened 2026-06-04
[OF#3535] Fix wrong 'completed' submission step

### LINKED ISSUE #3535: Pausing a form allows user to click next step, even if progress is blocked

2.2.0 - latest

A user can pause a form, resume and just click the next step in the progress-indicator to skip the step they are supposed to resume. 

Start form
Pause form
Continue form
Click next available step

_No response_

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

vaszig:
Closes open-formulieren/open-forms#3535 partly

### DIFF
--- a/src/data/submission-steps.ts
+++ b/src/data/submission-steps.ts
@@ -96,7 +96,14 @@ export const checkStepLogic = async (
   return result.data!;
 };
 
-type SubmissionStepCreateOrUpdateBody = Pick<SubmissionStep, 'data'>;
+type SubmissionStepCreateOrUpdateBody = Pick<SubmissionStep, 'data'> & {
+  /**
+   * Flag to indicate whether the step is going to be saved via a form being suspended or
+   * a regular submit of the step.
+   */
+
+  fromSuspension: boolean;
+};
 
 export const saveStepData = async (
   /**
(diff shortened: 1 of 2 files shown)


src/data/submission-steps.ts
```diff
@@ -56,6 +56,11 @@ export interface SubmissionStep {
    * but it can be mutated as part of the frontend logic evaluation.
    */
   canSubmit: boolean;
+  /**
+   * Flag to indicate whether the step is going to be saved via a form being suspended or
+   * a regular submit of the step.
+   */
+  fromSuspension: boolean;
```
sergei-maertens: we don't need this at this level, right? It should be a write-only property that's included in our PUT call, but I wouldn't expect to see this in our `GET` calls, and this type models what we get in GET response data.

src/data/submission-steps.ts
```diff
@@ -96,7 +101,7 @@ export const checkStepLogic = async (
   return result.data!;
 };
 
-type SubmissionStepCreateOrUpdateBody = Pick<SubmissionStep, 'data'>;
+type SubmissionStepCreateOrUpdateBody = Pick<SubmissionStep, 'data' | 'fromSuspension'>;
```
sergei-maertens:
```suggestion
type SubmissionStepCreateOrUpdateBody = Pick<SubmissionStep, 'data'> & {
  /**
   * Flag to indicate whether the step is going to be saved via a form being suspended or
   * a regular submit of the step.
   */
  fromSuspension: boolean;
};
```
probably the better approach :)

src/data/submission-steps.ts
```diff
@@ -110,12 +115,17 @@ export const saveStepData = async (
    * Nesting can occur here, if a key like `foo.bar` is set, it creates a parent object
    * for the key `foo` with a child property `bar`.
    */
-  data: SubmissionStep['data']
+  data: SubmissionStep['data'],
+  fromSuspension: SubmissionStep['fromSuspension'] = false
 ): Promise<void> => {
   // if data is not valid, this throws a `ValidationError` with the `asFormikProps`
   // method, which contains the errors in the suitable format for the formio-renderer.
+  console.log('dsklfjds;jflkdsfjflksjflkj');
```
sergei-maertens: Ah, you're fluent in Welsh too?

sergei-maertens (changes requested):

src/data/submission-steps.ts
vaszig: This has been added to the serializer and I don't do any custom operations so it's in the `GET` calls too with the default to `False` value, that's why I added it.

vaszig: And I guess we have to choose between  this here and the one in your comment below.

sergei-maertens: I'd update the backend serializer to be `write_only=True` and then it won't be returned for GET requests. There's no reason to expose that information IMO :)

sergei-maertens (approved):

PR #989, opened 2026-06-04
fix font size for Profile component

### LINKED ISSUE #6279: Font size of profile plugin output differs from general font size in overview page

3.5.0

### Customer reference

dimpact

The output from profile plugin has a bigger font size than the overview page.
Reported by D-impact, confirmed on our latest version.

1. Go to Formulieren on Maykin test, choose 'profieltest' - form
2. Fill the form in and continue to the overview page
3. Compare the font size of E-mailadres/Telefoonnummer and Text1/Text2

Image

Same size font on the overview page

### Which renderer?

new renderer

### Screen resolution

None

### Device

Desktop

### OS

Linux

### Browser

Firefox

**Tasks**

- [x] Fix renderer (`main` branch)
- [x] Update SDK (main + stable/3.5.x to new renderer patch release)
- [x] Implement a 3.4.x CSS only workaround (renderer approach/backport is not possible there)

annashamray:
fixes: https://github.com/open-formulieren/open-forms/issues/6279

### DIFF
--- a/src/components/Summary/FormStepSummary.stories.ts
+++ b/src/components/Summary/FormStepSummary.stories.ts
@@ -556,3 +556,115 @@ export const WithCustomerProfileAsSecondElement: Story = {
     ],
   },
 };
+
+export const WithMultipleValues: Story = {
+  args: {
+    data: [
+      {
+        name: 'textfield',
+        value: ['first', 'second'],
+        component: {
+          id: 'textfield',
+          type: 'textfield',
+          key: 'textfield',
+          label: 'Textfield',
+          multiple: true,
+        },
+      },
+      {
+        name: 'BSN',
+        value: ['XXXXXXXXX', 'XXXXXXXXX'],
+        component: {
+          id: 'bsn',
+          type: 'bsn',
+          key: 'bsn',
+          label: 'BSN',
+          multiple: true,
+        },
+      },
+      {
+        name: 'Date',
+        value: ['2026-01-01', '2026-02-02'],
+        component: {
+          id: 'date',
+          type: 'date',
+          key: 'date',
+          label: 'Date',
+          multiple: true,
+        },
+      },
+      {
+        name: 'Email',
+        value: ['first@example.org', 'second@example.org'],
+        component: {
+          id: 'email',
+          type: 'email',
+          key: 'email',
+          label: 'Email',
+          multiple: true,
+        },
+      },
+      {
+        name: 'Phone number',
+        value: ['0612345678', '+3187654321'],
+        component: {
+          id: 'phoneNumber',
+          type: 'phoneNumber',
+          key: 'phoneNumber',
+          label: 'Phone number',
+          multiple: true,
+        },
+      },
+      {
+        name: 'Select',
+        value: ['Option 1', 'Option 2'],
+        component: {
+          id: 'select',
+          type: 'select',
+          key: 'select',
+          label: 'Select',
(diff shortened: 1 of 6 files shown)


annashamray:
According to the NLDS [documentation](https://nl-design-system.github.io/utrecht/storybook/?path=/docs/css_css-unordered-list--docs) the properties including `list-style-type` can be configured via design tokens. Am I correct that I don't need to add any extra classes  for it because it's supported by NL DS community?

src/components/Summary/FormStepSummary.stories.ts
```diff
@@ -450,6 +450,16 @@ export const WithPartnersAsSecondElement: Story = {
...
+      {
+        name: 'First textfield',
+        value: 'first value',
+        component: {
+          id: 'component5',
+          type: 'textfield',
+          key: 'textfield1',
+          label: 'First textfield',
+        },
+      },
```
sergei-maertens: Please read the story name again and check where you added this extra component 😄

sergei-maertens (changes requested):
I'd also like to see a story for comparison with a textfield with `multiple: true` for example. I think there are still places in `ComponentValueDisplay.tsx` that use the old list component, that would have the same inconsistency.

sergei-maertens:
> According to the NLDS [documentation](https://nl-design-system.github.io/utrecht/storybook/?path=/docs/css_css-unordered-list--docs) the properties including `list-style-type` can be configured via design tokens. Am I correct that I don't need to add any extra classes for it because it's supported by NL DS community?

We don't yet have definitions for those design tokens in https://github.com/open-formulieren/design-tokens and now the level 1 effectively changes from a dash to a disc :thinking: It would be nice if we can keep the marker the same and only address the font-size issue.

src/components/Summary/FormStepSummary.stories.ts
annashamray: Haha, moved it to the separate story :)

annashamray:
@sergei-maertens I've added a story with several components with `"multiple=True"`, but it looks like all of them have `;` as a separator. 
I've created a PR in the design-tokens repo to change marker type, but I wan't able to make it work locally. I've also created an [issue](https://github.com/nl-design-system/utrecht/issues/3168) in NL DS repo about it

src/scss/nl-design-system-community.scss
```diff
@@ -2,3 +2,4 @@
 @use '@utrecht/img-css';
 @use '@utrecht/heading-1-css';
 @use '@utrecht/heading-3-css';
+@use '@utrecht/unordered-list-css';
```
sergei-maertens: can be removed again after renderer upgrade

sergei-maertens (changes requested):

sergei-maertens:
Commits to 3.4.x:

* 6d0c47855a637d55fce9be80792f70718ac824d8
* 39ddd28315a2821519d999fa50698a11a4ce8329

Backport to 3.5.x can be done once this is wrapped up and merged.

sergei-maertens:
Okay, parking this up for now because we have some other bugfixes that should be included in the next renderer patch version :)

sergei-maertens (approved):

sergei-maertens:
Backported to 3.5.x in f07436dfb940aaf8569adf7b45780e426038bfb9. Backporting to 3.4 is not feasible due to the dependency updates.

PR #991, opened 2026-06-10
Harden CI workflow

### LINKED ISSUE #6126: Harden CI workflows
Following the Trivy supply chain attack, we want to take actions to improve our supply chain attack resiliency.

The focus is on:

- preventing malicious actors from pushing modified Open Forms container images (by stealing credentials via supply chain attacks)
- limiting our attack surface and being resilient to phishing
- limiting the amount of secrets used in/expose to Github Actions workflows

June 30th at the latest, we'll flip the organisation/repository settings on which actions are allowed.

**Tasks**

- [x] Audit all our workflows in (active) repositories for actions used (grep on `uses: ` in the YAML):
    * any action from the `maykinmedia` github organisation is okay and will be allow-listed
    * official github actions usage is okay (`actions` and `github` organisations on github)
    * inventarize which third party actions we use and whether we want to put them on the allowlist or implement our own alternative
- [ ] Pin actions used to commit hashes rather than tags
    * Include a comment that says to which tag this points
    * Document/educate the development team that changes to these hashes must be thoroughly audited/reviewed. Company wide, we'll explore if we can automate this.
    * Note that this does not protect transitive dependencies (!)
- [ ] Github does not inject secrets available in the repository to workflows that don't make use of them. E.g. the code quality workflow should not have the `CODECOV_TOKEN` secret injected at all.
    * Verify and confirm this behaviour
    * Check if we can limit secret exposure in the `ci.yml` workflow by moving some steps into a sub-workflow. E.g. if the Trivy scanner can run in its own workflow with access to the image artifact but without access to the `DOCKER_TOKEN` secret, that would be ideal.
- [ ] Explore if we can limit release tag pushes to only GPG-signed tags - which requires our development team to sign them to create a release. Otherwise, explore tag protection rules and share with the rest of the company.
- [ ] Explore docker image build attestion so that published images can be traced back to a github tag/release/workflow run
- [ ] Explore github immutable releases: https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases

viktorvanwijk:
Partly closes open-formulieren/open-forms#6126

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,10 +23,12 @@ jobs:
       version: ${{ steps.vars.outputs.version }}
       git_hash: ${{ steps.vars.outputs.git_hash }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          persist-credentials: false
       - name: Extract version info
         id: vars
-        uses: maykinmedia/open-api-workflows/actions/extract-version@v6
+        uses: maykinmedia/open-api-workflows/actions/extract-version@79102b911003d75203ca2fed7df01ad79d9b6bba # v6.4.0
 
   build:
     name: Create 'production' build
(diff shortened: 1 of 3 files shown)


viktorvanwijk:
TODO: bump the following packages to latest version after new release with pinned commit hashes
- design-tokens
- formio-renderer
- leaflet-tools
- monaco-json-editor
- types

sergei-maertens (approved):

PR #992, opened 2026-06-16
Fix excessive address-autocomplete/latlng-search calls

### LINKED ISSUE #6354: "latlng-search" and "address-autocomplete" are firing on every interaction

3.4.9 & 3.5.3

### Customer reference

DB: 162

This is seems to cause "throttling" for submitters, where forms have multiple location in steps, or a lot of questions after providing a address. 

Image

3.4.9: Seems to generate twice as much requests compared 3.5.3
3.5.3: only 1 request where you would normally see a "check_logic" call.

sergei-maertens:
(Partly) closes open-formulieren/open-forms#6354 

---

~~Ideally this PR also includes https://github.com/open-formulieren/formio-renderer/pull/341~~ Included with the new renderer patch release via #989

---

The original implementation used getFieldMeta with the assumption that the identity was stable and would prevent unnecessary useEffect calls when the postcode/house number values don't change, however that's not the case. It depends on formik.values, formik.touched and formik.errors which change constantly when filling out the rest of the form.

The reworked implementation avoids this by using Formik's getIn directly on the values that we care about, and avoids using values (or other meta) in the useEffect dependency array. Additionally, we now bail early if there are no updates to be done at all because there are no candidates to be updated, avoiding some more network IO.

Finally, the debounce of 200ms was quite short and has been upped to 300ms, which should hopefully strike a better balance between snappy UI and giving users enough time to type their house number.

### DIFF
--- a/src/data/geo.ts
+++ b/src/data/geo.ts
@@ -7,6 +7,7 @@ import type {
 import AbstractProvider from 'leaflet-geosearch/src/providers/provider.js';
 
 import {get} from '@/api';
+import {getCached, setCached} from '@/cache';
 import {logError} from '@/components/Errors';
 
 /**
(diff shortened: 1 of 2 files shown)


src/data/geo.ts
```diff
@@ -18,18 +19,25 @@ export interface AutoCompleteResult {
...
+const GEO_CACHE_TTL = 15 * 60 * 1000; // 15 minutes
+
 export const autoCompleteAddress = async (
   baseUrl: string,
   postcode: string,
   houseNumber: string
 ): Promise<AutoCompleteResult | null> => {
+  const cacheKey = `${postcode.replace(' ', '')}|${houseNumber}`;
+  const cacheResult = getCached<AutoCompleteResult>(cacheKey, GEO_CACHE_TTL);
+  if (cacheResult !== null) return cacheResult;
```
vaszig, line 31: Will this be null or undefined?It depends on the api response right?

vaszig (approved):

src/data/geo.ts
sergei-maertens: cache returns `null` if it's not present in the cache :) the API result will never be null - we return `null` if we get an API error, but in that case we don't store the result in the cache either, so `null` works correctly as a marker for "no cache hit"

sergei-maertens:
Backports:

* stable/3.5.x: 9975b180b9fbea632a9f5edf5e80145a3d7f28b9
* stable/3.4.x: e9eb7dd34f9a431cb25073cb7832a11d932099a3

PR #994, opened 2026-06-17
[OF#6277] Add `UNDEFINED_VALUE` for variable action with `null` values

### LINKED ISSUE #6277: Update frontend and backend JSON logic implementation
This is to properly check if a value is present in the context, instead of defaulting to `None`/`null`. In the backend, it means putting a custom object as default in the `not_found` argument of `get_var`. Need to implement something similar in the frontend, possibly need a custom override of the "var" operation.

To not break the `!` and `!!` operators, probably need to make sure the custom value is nullish, like the `undefined` value in JS. This also makes sure `{"if": [{"var": "foo"}, ...]}` will still work

vaszig:
Closes open-formulieren/open-forms#6277 partly


src/components/FormStep/logic.spec.ts
```diff
@@ -231,6 +231,101 @@ test('clearOnHide behaviour is applied', () => {
...
+        {
+          component: 'textField',
+          action: {
+            type: 'property',
+            property: {value: 'hidden', type: 'bool'},
+            state: true,
+          },
+        },
+        {
+          component: '',
```
viktorvanwijk: I would leave `component` out for variable actions

src/components/FormStep/logic.spec.ts
```diff
@@ -231,6 +231,101 @@ test('clearOnHide behaviour is applied', () => {
...
+    defaultConfiguration: {components},
+  };
+  let updatedComponents: AnyComponentSchema[] = [];
+  let dataUpdates: JSONObject | null = {};
+
+  evaluateBackendRules({
+    submission,
+    step,
+    rules,
+    inputData: {checkbox: true, textField1: ''},
```
viktorvanwijk:
```suggestion
    inputData: {checkbox: true, textField1: '', textField: ''},
```
To match the backend test everything is visible initially :)

src/logic/json-logic/extensions/context.ts
```diff
@@ -7,19 +7,76 @@
 import {defaultMethods} from 'json-logic-engine';
 
 import type {JsonLogicEngineMethod} from './types';
+import {isInPath} from './utils';
+
+// Object for treating null values (missing) in a more specific way.
+// Symbol would not work in our situation because it would evaluate to [Object Object]
+export const UNDEFINED_VALUE = Object.freeze({
+  __jsonLogicUndefinedValue: true,
+});
```
viktorvanwijk: Is there a reason why we cannot use the JS-native `undefined` value? I think we only have to define such a custom object in Python, because that concept is not present there

src/logic/actions/variable.ts
```diff
@@ -35,6 +36,13 @@ export const applyVariableAction = (
   // isEqual is necessary for deep equality checks with Arrays and Objects
   if (isEqual(currentValue, targetValue)) return;
 
-  logicState.data = setIn(updatedData, componentKey, targetValue);
-  logicState.initialValues = setIn(logicState.initialValues, componentKey, targetValue);
+  if (targetValue === UNDEFINED_VALUE) {
+    // when the variable is not present we do not want to have null but the specific
+    // UNDEFINED_VALUE which means we have to populate the variable with its initial
+    // value
+    logicState.data = setIn(updatedData, componentKey, logicState.initialValues[componentKey]);
```
viktorvanwijk: Why is this necessary? It seem like this does something different from the backend, where we do nothing (i.e. return a `{}` as the data mutation) if we encounter an undefined value.

It should be enough to just gate the setting of values behind a `if (targetValue !== UNDEFINED_VALUE)`, no?

src/logic/json-logic/extensions/context.ts
```diff
@@ -7,19 +7,76 @@
...
 
 const originalVar = defaultMethods['var'].method;
+const originalCat = defaultMethods['cat'].method;
+const originalSubstr = defaultMethods['substr'] as (args: unknown[]) => unknown;
```
viktorvanwijk: Why is the casting necessary here? Also, we should know what the expected arguments and return value are, right?

src/logic/json-logic/extensions/context.ts
```diff
@@ -7,19 +7,76 @@
...
+ * {someVar: null}    { var: "someVar" }	              null
+ * {someVar: null}	  { var: ["someVar", "default"] }   "default"
+ * {}	                {var: "someVar"}                  UNDEFINED_VALUE
+ * {}                 { var: ["someVar", "default"] }   "default"
+ */
 export const customVar: JsonLogicEngineMethod = (key, context, above, engine) => {
-  let fallback: unknown;
-  if (Array.isArray(key) && typeof key[0] === 'string' && key.length > 1) {
-    fallback = key[1];
+  if (Array.isArray(key) && typeof key[0] === 'string') {
```
viktorvanwijk: It seems like it is possible to pass our custom fallback value directly to the `originalVar` function, like:

```javascript
resolved = originalVar([varName, fallback], context, above, engine);
```

However, it seems to return `null` anyway even if we explicitly pass the JS-native `undefined` to it, which is not ideal. I would suggest to just do a full custom implementation of the var operation then, and skip calling `originaVar` entirely. Basically converting the json-logic-py implementation into JS

viktorvanwijk (commented):
General question: do the other operators affected by https://github.com/maykinmedia/json-logic-py/blob/2bd86798d74820c035fe48ff56fc70b9551aa411/json_logic/__init__.py#L410 still work as expected now?

src/logic/json-logic/extensions/context.ts
vaszig: Just wanted to make it explicit and mimic the `json-logic-py` library but ok, I changed it according to your suggestion.

src/logic/json-logic/extensions/context.ts
vaszig: Because of the type of the original `substr` function (`(args: [any, any, any]) => any`). I couldn't find another way of declaring the types tbh.

vaszig:
> General question: do the other operators affected by https://github.com/maykinmedia/json-logic-py/blob/2bd86798d74820c035fe48ff56fc70b9551aa411/json_logic/__init__.py#L410 still work as expected now?

Are you referring to the `substr` for example and `{}` as data?Because this is fixed indeed.

viktorvanwijk:
> Are you referring to the `substr` for example and `{}` as data?Because this is fixed indeed.

Yeah the `substr`, but also the `date` operator for example

vaszig:
> > Are you referring to the `substr` for example and `{}` as data?Because this is fixed indeed.
> 
> Yeah the `substr`, but also the `date` operator for example

Have tested these operators with missing data (variables) and seem to work as expected.

src/logic/json-logic/extensions/context.ts
```diff
@@ -8,18 +8,72 @@ import {defaultMethods} from 'json-logic-engine';
...
+  if (Array.isArray(key)) {
+    varName = key[0];
+
+    if (key.length > 1) {
+      hasDefault = true;
+      notFound = key[1];
+    }
+  } else {
+    varName = key;
+  }
```
viktorvanwijk: Like this maybe?
```suggestion
  let varName: JSONValue | undefined;
  let defaultValue: JSONValue | undefined = undefined;
  if (Array.isArray(key)) {
    [varName, defaultValue] = key as [JSONValue, JSONValue] | [JSONValue] | [];
  } else {
    varName = key;
  }
```
The casting is not nice, but as least it's clear what we expect `key` to be

src/logic/json-logic/extensions/context.ts
```diff
@@ -8,18 +8,72 @@ import {defaultMethods} from 'json-logic-engine';
...
+
+  if (!varName) {
+    return context;
+  }
+
+  let data = context;
+
+  for (const segment of String(varName).split('.')) {
+    if (data == null) {
+      return hasDefault ? notFound : undefined;
```
viktorvanwijk: Then you can just return `defaultValue` here I think, no?

src/logic/json-logic/extensions/context.ts
```diff
@@ -8,18 +8,72 @@ import {defaultMethods} from 'json-logic-engine';
...
-    const varName = key[0];
-    // lookup without default and use original, if it comes back as `null` it was either
-    // present with `null` or was absent in the context - in both cases we fall back to
-    // the default value
-    const resolved = originalVar([varName], context, above, engine);
-    if (resolved === null) return fallback;
+const originalCat = defaultMethods['cat'].method;
+const originalSubstr = defaultMethods['substr'] as (args: unknown[]) => unknown;
+
+// json logic concerning the var operator is tricky, so we impelement this from scratch
```
viktorvanwijk:
```suggestion
// json logic concerning the var operator is tricky, so we implement this from scratch
```

src/logic/json-logic/extensions/context.ts
viktorvanwijk: OK, yeah I see the original type hints are a mess. Could you move these to the relevant functions, though? I don't see a reason why they need to be defined as global constants

src/logic/json-logic/extensions/context.ts
```diff
@@ -8,18 +8,72 @@ import {defaultMethods} from 'json-logic-engine';
...
+};
+
+export const customSubstr: JsonLogicEngineMethod = key => {
+  if (!Array.isArray(key)) {
+    return originalSubstr(key);
+  }
+
+  const normalized = key.map(value => (value === undefined ? null : value));
+  if (normalized[0] === null) {
+    return null;
```
viktorvanwijk, line 71: Not sure if we have tests for this, but in theory they can all be variable expressions, which means they can all be `null`/`undefined` (this is also what the backend does [here](https://github.com/maykinmedia/json-logic-py/blob/2bd86798d74820c035fe48ff56fc70b9551aa411/json_logic/__init__.py#L413))

src/logic/json-logic/var.spec.ts
```diff
@@ -0,0 +1,58 @@
+import type {JSONObject} from '@open-formulieren/types';
+import {expect, test} from 'vitest';
+
+import evaluate from './index';
+
+test('evaluate with existing variable', () => {
```
viktorvanwijk, line 6: Could you add some more tests with `var` being `[]` and `[foo]`?

src/logic/json-logic/extensions/context.ts
vaszig: Yes, I have for `null` and I am adding for `undefined` too.

src/logic/json-logic/var.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+});
+
+test('evaluate with empty array', () => {
+  const expression: JSONObject = {var: 'a'};
+  const data: JSONObject = {a: []};
+
+  const result = evaluate(expression, data);
+
+  expect(result).toStrictEqual([]);
+});
```
viktorvanwijk, line 31: I actually meant the following: `{var: ['foo']}` and `{var: []}`, not the data being these values :grin:

src/logic/json-logic/var.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+test('evaluate array of values', () => {
+  const expression: JSONObject = {var: 'a.1'};
+  const data: JSONObject = {a: [3, 5]};
+
+  const result = evaluate(expression, data);
+
+  expect(result).toBe(5);
+});
```
viktorvanwijk, line 94: Some tests where the result should be `undefined` are also a good idea for this nested-data access I think. For example:
```javascript
expression = {var: 'a.1'};
data = {a: 'a string instead of a list'};
```

src/logic/json-logic/extensions/context.ts
```diff
@@ -4,22 +4,72 @@
...
+  }
+
+  if (!varName) {
+    return context;
+  }
+
+  let data = context;
+
+  for (const segment of String(varName).split('.')) {
+    if (data === null || (typeof data !== 'object' && !Array.isArray(data))) {
```
vaszig: I think your test suggestion revealed another missing case I had.

viktorvanwijk: My apologies, the test I suggested isn't actually supposed to be `undefined`.

In the backend:
```python
>>> jsonLogic({"var": "a.1"}, {"a": "this is a string"})
>>> 'h'
>>> jsonLogic({"var": "a.b"}, {"a": "this is a string"})
>>>
```

Hopefully you didn't spend too much time "fixing" it :grimacing:

vaszig: No problem at all, I will fix it!

viktorvanwijk (approved):

PR #995, opened 2026-06-19
[OF#6350] Add support for hiding form title

### LINKED ISSUE #6350: Add support for hiding the form title
We want to give the ability to hide the title of the form, as there might be two h1 elements on the page.

vaszig:
Closes open-formulieren/open-forms#6350

### DIFF
--- a/.storybook/decorators.tsx
+++ b/.storybook/decorators.tsx
@@ -4,13 +4,13 @@ import type {AnyComponentSchema} from '@open-formulieren/types';
 import type {Decorator} from '@storybook/react-vite';
 import {Document} from '@utrecht/component-library-react';
 import {Formik} from 'formik';
+import {NuqsTestingAdapter} from 'nuqs/adapters/testing';
 import React from 'react';
 
 import {ConfigContext, FormContext} from '@/Context';
 import type {ConfigContextType} from '@/Context';
 import {BASE_URL, buildForm} from '@/api-mocks';
 import Card from '@/components/Card';
-import { NuqsTestingAdapter } from 'nuqs/adapters/testing';
 import {LiteralsProvider} from '@/components/Literal';
 import {SubmissionStatusContext} from '@/components/PostCompletionViews';
 import {AnalyticsToolsConfigContext} from '@/components/analytics/AnalyticsToolConfigProvider';
@@ -78,6 +78,7 @@ export const withConfig: Decorator = (Story, {parameters}) => {
   // General configuration
   const DEFAULTS: ConfigContextType = {
     baseUrl: BASE_URL,
+    showFormTitle: true,
     basePath: '',
     clientBaseUrl: '',
     baseTitle: '',
--- a/src/components/CoSign/CosignStart.tsx
+++ b/src/components/CoSign/CosignStart.tsx
@@ -1,7 +1,7 @@
 import {FormattedMessage} from 'react-intl';
 
 import Body from '@/components/Body';
-import Card from '@/components/Card';
+import FormContainer from '@/components/FormContainer';
 import {LiteralsProvider} from '@/components/Literal';
 import LoginOptions from '@/components/LoginOptions';
 import MaintenanceMode from '@/components/MaintenanceMode';
@@ -25,7 +25,7 @@ const CosignStart: React.FC = () => {
 
   const userIsFormDesigner = IsFormDesigner.getValue();
   if (!userIsFormDesigner && form.maintenanceMode) {
-    return <MaintenanceMode title={form.name} />;
+    return <MaintenanceMode />;
   }
 
   if (outagePluginId) {
@@ -34,7 +34,7 @@ const CosignStart: React.FC = () => {
     );
     if (!loginOption) throw new Error('Unknown login plugin identifier');
     return (
-      <Card
+      <FormContainer
         title={
           <FormattedMessage
             description="Form start outage title"
@@ -44,13 +44,13 @@ const CosignStart: React.FC = () => {
         }
       >
         <AuthenticationOutage loginOption={loginOption} />
-      </Card>
+      </FormContainer>
     );
   }
 
   return (
     <LiteralsProvider literals={form.literals}>
-      <Card title={form.name}>
+      <FormContainer>
         {userIsFormDesigner && form.maintenanceMode && <MaintenanceMode asToast />}
 
         {authError && <AuthenticationError parameter={authError[0]} errorCode={authError[1]} />}
@@ -71,7 +71,7 @@ const CosignStart: React.FC = () => {
           onFormStart={() => {}}
           isolateCosignOptions={false}
         />
-      </Card>
+      </FormContainer>
     </LiteralsProvider>
   );
 };
--- a/src/components/FormContainer.tsx
+++ b/src/components/FormContainer.tsx
@@ -1,0 +1,21 @@
+import {useContext} from 'react';
+
+import {ConfigContext} from '@/Context';
+import useFormContext from '@/hooks/useFormContext';
+
+import type {CardProps} from './Card';
+import Card from './Card';
+
+/**
+ * Container component for wrapping the Card component.
+ *
+ * This is a wrapper that helps render (or not) the title of a form.
+ */
+const FormContainer = (props: CardProps & Omit<React.ComponentPropsWithoutRef<'div'>, 'title'>) => {
+  const {showFormTitle} = useContext(ConfigContext);
+  const form = useFormContext();
+  const cardTitle = showFormTitle ? props.title || form.name : undefined;
+  return <Card {...props} title={cardTitle} />;
+};
+
+export default FormContainer;
--- a/src/components/FormStart/index.tsx
+++ b/src/components/FormStart/index.tsx
@@ -10,6 +10,7 @@ import Body from '@/components/Body';
 import Card from '@/components/Card';
 import FormMaximumSubmissionsError from '@/components/Errors/FormMaximumSubmissionsError';
 import ExistingSubmissionOptions from '@/components/ExistingSubmissionOptions';
+import FormContainer from '@/components/FormContainer';
 import {LiteralsProvider} from '@/components/Literal';
 import LoginOptions, {type OnFormStartOptions} from '@/components/LoginOptions';
 import MaintenanceMode from '@/components/MaintenanceMode';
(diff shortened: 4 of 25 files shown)


src/components/Card.tsx
```diff
@@ -60,14 +62,18 @@ const Card: React.FC<CardProps & Omit<React.ComponentPropsWithoutRef<'div'>, 'ti
...
   if (mobileHeaderHidden) modifiers.push('mobile-header-hidden');
 
   const className = getBEMClassName('card', modifiers);
   return (
     <div className={className} {...restProps}>
       {/* Emit header/title only if there is one */}
-      {title && <CardTitle title={title} headingLevel={1} headingType={titleHeadingType} />}
+      {title && showFormTitle && (
+        <CardTitle title={title} headingLevel={1} headingType={titleHeadingType} />
+      )}
```
sergei-maertens: this is not the right approach - the `Card` component is isolated/decoupled from the form. You should be able to use this component in other places with titles other than the form title.

Instead, the places where the form title is passed to the card should be updated to conditionally render the form title, e.g. in `src/components/FormStart/index.tsx`.

Because this code is repeating in a number of places, I'd abstract this in a new component:

```tsx
const FormContainer = (props) => {
  const {showFormTitle} = useContext(ConfigContext);
  const form = useFormContext();
  return <Card {...props} title={showFormTitle ? form.name : undefined} />
};
```

and then you can update the existing places where `<Card> ... </Card>` is rendered with `<FormContainer> ... </FormContainer>`

sergei-maertens (changes requested):

src/components/Card.tsx
sergei-maertens: See also for example `src/components/MaintenanceMode.tsx` that would break with your patch

.storybook/decorators.tsx
```diff
@@ -130,9 +131,9 @@ export const withPageWrapper: Decorator = Story => (
  * Card props can be customized through story `parameters.card`.
  */
 export const withCard: Decorator = (Story, {parameters: {card: cardProps = {}}}) => (
-  <Card {...cardProps}>
+  <FormContainer {...cardProps}>
     <Story />
-  </Card>
+  </FormContainer>
 );
```
sergei-maertens, line 137: this should still use the `Card` component, no?

The decorator is named quite explicitly `withCard` - the form title settings/context shouldn't affect these stories as they're only stories.

src/components/appointments/cancel/CancelAppointment.stories.tsx
sergei-maertens, line 1: I don't understand why these stories are updated that use the card component as intended - standalone. They don't represent the actual form container component for the scaffolding.

src/components/appointments/cancel/CancelAppointmentSuccess.tsx
sergei-maertens, line 1: This too shouldn't be updated - the component didn't display the form title in the first place.

I don't find these appointment forms part of the scope for hiding the title with embedding. IMO it's limited to regular and single step forms.

src/components/CoSign/CosignDone.tsx
sergei-maertens: I think not rendering the title on these pages is far more damaging than potentially having two page titles.

If the form name/title is not displayed in the first place, there isn't anything to omit with `showFormTitle: false`, so I'd say: keep the existing behaviour.

sergei-maertens (changes requested):

src/components/appointments/cancel/CancelAppointment.stories.tsx
vaszig: I thought so too but I then saw https://open-formulieren.github.io/open-forms-sdk/?path=/story/private-api-appointments-cancellation-cancel--missing-time-parameter which does show the form title (`Error`).

sergei-maertens: I wouldn't say that's the "form title", it's just a placeholder for the story

src/components/appointments/cancel/CancelAppointmentSuccess.tsx
vaszig: Here too, https://open-formulieren.github.io/open-forms-sdk/?path=/story/private-api-appointments-cancellation-success--success. There is a title, but I guess you mean there is no form title specifically?So that's why I shouldn't change the Card here?

sergei-maertens: yep!

src/components/CoSign/CosignStart.tsx
```diff
@@ -34,7 +34,7 @@ const CosignStart: React.FC = () => {
     );
     if (!loginOption) throw new Error('Unknown login plugin identifier');
     return (
-      <Card
+      <FormContainer
```
sergei-maertens, line 37: I think this needs to remain a normal `Card`, it's not the form name being used.

src/components/FormStart/index.tsx
```diff
@@ -108,9 +109,9 @@ const FormStart: React.FC = () => {
   // do not re-render the login options while we're redirecting
   if (doStart && !hasAuthErrors) {
     return (
-      <Card>
+      <FormContainer>
```
sergei-maertens: I think this too should be a plain `Card` because no form name/title is passed.

src/components/PostCompletionViews/PostCompletionView.tsx
```diff
@@ -33,7 +33,7 @@ const PostCompletionView: React.FC<PostCompletionViewProps> = ({
   );
 
   return (
-    <Card title={header}>
+    <FormContainer title={header}>
```
sergei-maertens: I can't tell if this is using the form name as title or something else 🤔 did you double check it?

src/components/PostCompletionViews/StatusUrlPoller.tsx
```diff
@@ -82,7 +82,7 @@ const StatusUrlPoller: React.FC<StatusUrlPollerProps> = ({
 
   if (loading) {
     return (
-      <Card
+      <FormContainer
```
sergei-maertens: I think this too should be a plain `Card` because no form name/title is passed.

src/components/Sessions/SessionExpired.tsx
```diff
@@ -28,7 +28,7 @@ const SessionExpired: React.FC = () => {
 
   return (
     <>
-      <Card
+      <FormContainer
```
sergei-maertens: I think this too should be a plain `Card` because no form name/title is passed.

src/components/Summary/GenericSummary.tsx
```diff
@@ -76,7 +77,7 @@ const GenericSummary: React.FC<GenericSummaryProps> = ({
   }
 
   return (
-    <Card title={title}>
+    <FormContainer title={title}>
```
sergei-maertens: I'm not sure what we pass here as `title`? If it's something else than the form name, I'd stick with a plain `Card` instead.

src/components/Card.stories.tsx
```diff
@@ -22,4 +25,10 @@ export const Default: Story = {
     children: <Body>By Blanck Mass</Body>,
     mobileHeaderHidden: false,
   },
+  decorators: [withConfig],
+  play: async ({canvasElement}) => {
+    const canvas = within(canvasElement);
+
+    expect(canvas.queryByText('The Rig')).toBeInTheDocument();
+  },
```
sergei-maertens: I don't think this test adds much value - we already have the Chromatic snapshots that did their work :)

src/components/FormContainer.tsx
```diff
@@ -0,0 +1,20 @@
...
+ * Container component for wrapping the Card component.
+ *
+ * This is a wrapper that helps render (or not) the title of a form.
+ */
+export const FormContainer = (
+  props: CardProps & Omit<React.ComponentPropsWithoutRef<'div'>, 'title'>
+) => {
+  const {showFormTitle} = useContext(ConfigContext);
+  const form = useFormContext();
+  return <Card {...props} title={props.title || (showFormTitle ? form.name : undefined)} />;
```
sergei-maertens: isn't this a bit of a weird construct? I'd expect if the config says "do not show the title", that then even the explicitly passed in title would be hidden. So it should probably be something more akin:

```tsx
const cardTitle = showFormTitle ? (props.title || form.name) : undefined;
```

src/components/MaintenanceMode.tsx
```diff
@@ -38,7 +39,7 @@ export interface MaintenanceModeProps {
 const MaintenanceMode: React.FC<MaintenanceModeProps> = ({title, asToast = false}) => {
   const alert = <MaintenanceModeAlert />;
   if (asToast) return alert;
-  return <Card title={title}>{alert}</Card>;
+  return <FormContainer title={title}>{alert}</FormContainer>;
 };
```
sergei-maertens, line 42: do we still have call sites that pass the `title`? If not, we can remove the prop entirely.

sergei-maertens (changes requested):
This PR seems to be tougher than expected 😬

src/components/MaintenanceMode.tsx
vaszig: Yes, we do have one place (https://github.com/open-formulieren/open-forms-sdk/blob/main/src/components/Errors/ErrorDisplay.tsx#L40) where we pass it, but it's custom message, so I would keep the `Card` here.

src/components/Summary/GenericSummary.tsx
vaszig: Yes, we pass a custom title and not the form name so I am keeping the `Card`.

src/components/PostCompletionViews/PostCompletionView.tsx
vaszig: It uses the `header` (which is passed along all the places that it's being used) so I guess I should leave the `Card` here.

src/components/CoSign/CosignStart.tsx
vaszig: But this is passing the form title that's why I left it (`formName: form.name`).

vaszig:
> This PR seems to be tougher than expected 😬

Yes, because I approached that in a different way, but indeed I agree with your suggestions.

src/components/CoSign/CosignStart.tsx
sergei-maertens: yes, agree!

sergei-maertens (approved):

PR #996, opened 2026-06-23
Fix missing window title update

### LINKED ISSUE #6359: Title is missing "Step title" and not updating on navigation

3.5.3

### Customer reference

Leiden 73

It seems the title is always missing the "step" title. And always displays as
`form title | Organisation name.`

Expected behaviour is:
`Step title | Form title | Organisation name.`

Also when navigating back from for example "Controleer en bevestig", this title "sticks"

Image

This is all new behaviour

_No response_

_No response_

### Which renderer?

None

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

sergei-maertens:
Closes open-formulieren/open-forms#6359


vaszig (commented):
I would add the hook to the `SingleFormStepNewRenderer` too. I think it makes sense, do you agree?

sergei-maertens:
> I would add the hook to the `SingleFormStepNewRenderer` too. I think it makes sense, do you agree?

Not really - there's only a single step in the form (so form name and step are basically the same thing) and there are no analytics to track in terms of navigation between steps. That's why I left it out.

vaszig (approved):

sergei-maertens:
Backports:

* stable/3.5.x: 98a64f4110acad972d2a2a53b087d1c92cac03d6
* stable/3.4.x: 0fbb014d19dbbf19f99bf425636ae7125dc6d393

PR #997, opened 2026-06-26
Ensure SDK matches backend clearOnHide behaviour

### LINKED ISSUE #6408: Ensure SDK matches new clearOnHide behavior from the backend
After #6275 and #6276, need to make sure the backend logic in the frontend now behaves the same as the backend.

viktorvanwijk:
Closes open-formulieren/open-forms#6408

### DIFF
--- a/src/logic/actions/property.ts
+++ b/src/logic/actions/property.ts
@@ -1,10 +1,7 @@
 import {getRegistryEntry} from '@open-formulieren/formio-renderer';
 import {processVisibility} from '@open-formulieren/formio-renderer/visibility.js';
-import type {JSONObject, JSONValue} from '@open-formulieren/types';
-import {getIn, setIn} from 'formik';
 import {set} from 'lodash';
 
-import {getComponentEmptyValue} from '@/components/FormStep/logic';
 import type {LogicAction, PropertyAction} from '@/data/logic';
 
 import type {LogicEvaluationState} from './types';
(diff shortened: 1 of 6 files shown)


src/logic/actions/property.ts
```diff
@@ -66,38 +63,17 @@ export const applyPropertyAction = (
...
-        // state of the parent(s) via hasHiddenParent(targetComponent, logicState),
-        // but because of the `clearValueCallback` to match the backend behaviour, this
-        // is currently not relevan/correct. See
-        // https://github.com/open-formulieren/open-forms/issues/6121.
+        // state of the parent(s) via hasHiddenParent(targetComponent, logicState).
+        // This is relevant if there are multiple logic rules (or a combination of logic rules and
+        // simple conditionals) targeting a layout component and a child of this layout component in
+        // an opposite way. For example, the first rule marks a fieldset as hidden, and the second
+        // marks a child textfield as visible. In the current situation, the textfield would be
+        // added to the data, even though the fieldset is hidden.
```
viktorvanwijk: We could look into this, but it requires some changes in the backend too, ~because the actions are not applied directly to the configuration.~ because it's also not checked there. Probably best to do a separate PR

src/logic/actions/property.ts
```diff
@@ -66,38 +63,17 @@ export const applyPropertyAction = (
...
         parentHidden: false,
         initialValues: logicState.initialValues,
         getRegistryEntry,
         componentsMap,
-        // Ensure we restore the default value OR empty value when clearing the
-        // value so that we match the backend behaviour. The formio-renderer will take
-        // care of properly removing the key from the submission data.
-        // See https://github.com/open-formulieren/open-forms/issues/6121
-        clearValueCallback: (values: JSONObject, key: string): JSONObject => {
```
vaszig, line 83: Will this be still in the renderer?I see we have it as an optional property there.

src/logic/actions/property.ts
vaszig: Will this be part of the same ticket?I see that you mention that this PR closes it, so just asking.

vaszig (commented):
Only a couple of questions for my understanding.

src/logic/actions/property.ts
viktorvanwijk: Yeah I would say separate ticket as well

src/logic/actions/property.ts
viktorvanwijk: For now, yes, but we will remove it right before the 4.0 release (see https://github.com/open-formulieren/formio-renderer/pull/346#issuecomment-4807863002)

vaszig (approved):
It looks good to me, maybe Sergei can also have a look since he was part of the backend reviewing too.

src/logic/actions/property.ts
viktorvanwijk: Yeeaahh, I immediately ran into this when testing with a big DH form. Going to fix it here as the PR is not that big (yet). Created a separate one for the backend, though

sergei-maertens: > Yeeaahh, I immediately ran into this when testing with a big DH form. Going to fix it here as the PR is not that big (yet). Created a separate one for the backend, though

Does that scenario even make sense? What happens in the formio playground if you set up a similar for with component conditionals?

viktorvanwijk: I had difficulty coming up with a scenario at first, but I think DH has provided us with one:

Consider a checkbox that shows a fieldset (hidden by default) and this fieldset contains a checkbox which hides a textfield (shown by default) also in that fieldset. The default state of that step is to show only the checkbox, but our current implementation will set a value for the textfield when processing the second logic rule, because we do not look at the fieldset's visibility. This creates an infinite loop of check-logic calls, because the textfield ends up in the data diff, but then removed again by the renderer.

sergei-maertens: That text field should never have ended up in the data, right? So the backend shouldn't add that data. Arguably the backend should take the visibility state into account, but you could also argue that the trigger of the textfield logic rule needs to deal with that

viktorvanwijk: > That text field should never have ended up in the data, right? So the backend shouldn't add that data.

Indeed, that's what happens with the fix in open-formulieren/open-forms#6421

> but you could also argue that the trigger of the textfield logic rule needs to deal with that

True, but is this something we want to enforce for the form designers?

sergei-maertens: Check, I hadn't seen that PR yet!

> True, but is this something we want to enforce for the form designers?

If we can handle it in our code without bending over backwards, that's of course the best approach. With your patches linked above, it looks like we don't need to consider this any further :)

viktorvanwijk:
Note: CI is still failing because I need a new version of the renderer :)

sergei-maertens (approved):

PR #1005, opened 2026-07-17
[OF#6458] Fix invalid dates and datetime in json logic input data

### LINKED ISSUE #6458: New Logic: when using a date component in logic trigger, the form crashes.

3.5.x

### Customer reference

Raakvlak 22

When a logic trigger uses a date field from a step, the step crashes on load. 

Image

If the same rule is used on a user variable (date) the logic works as expected.

_No response_

_No response_

### Which renderer?

None

### Screen resolution

None

### Device

None

### OS

None

### Browser

_No response_

vaszig:
Fixes open-formulieren/open-forms#6458

### DIFF
--- a/src/logic/json-logic/extensions/comparators.ts
+++ b/src/logic/json-logic/extensions/comparators.ts
@@ -22,6 +22,11 @@ const comparatorOverrideFactory = (original: JsonLogicEngineMethod): JsonLogicEn
       }
       return originalArg;
     });
+
+    if (args.some(arg => arg === null || arg === undefined)) {
+      return null;
+    }
+
     return original(modifiedArgs, context, above, engine);
   };
 };
--- a/src/logic/json-logic/extensions/date.ts
+++ b/src/logic/json-logic/extensions/date.ts
@@ -25,6 +25,10 @@ export const jsonLogicDate: JsonLogicEngineMethod = ([dateString]) => {
   // anything that isn't a string.
   if (typeof dateString !== 'string') return dateString;
 
+  if (dateString === '') {
+    return null;
+  }
+
   let _dateString: string = dateString;
   // to match backend behaviour, timezone-aware datetimes are converted by looking only
   // at the (naive) date part, without converting to UTC/local timezone. So, we strip
--- a/src/logic/json-logic/date.spec.ts
+++ b/src/logic/json-logic/date.spec.ts
@@ -19,6 +19,24 @@ test('accepts ISO-8601 datetime string but truncates to date', () => {
   expect(result).toBe('2026-03-12');
 });
 
+test('greater-than comparison with empty string', () => {
+  const expression: JSONObject = {
+    '>': [{date: ''}, {date: '2026-03-01'}],
+  };
+
+  const result = evaluate(expression, {});
+
+  expect(result).toBeNull();
+});
+
+test('greater-than without date operator', () => {
+  const expression: JSONObject = {
+    '>': [{date: '2000-10-10'}, '2000-01-01'],
+  };
+
+  expect(() => evaluate(expression, {})).toThrow();
+});
+
 test('greater-than comparison with ISO-8601 date strings', () => {
   const expression: JSONObject = {
     '>': [{date: '2026-03-01'}, {date: '2026-03-01'}],
@@ -39,6 +57,24 @@ test('greater-than-equals comparison with ISO-8601 date strings', () => {
   expect(result).toBe(true);
 });
 
+test('greater-than-equals comparison with empty string', () => {
+  const expression: JSONObject = {
+    '>=': [{date: ''}, {date: '2026-03-01'}],
+  };
+
+  const result = evaluate(expression, {});
+
+  expect(result).toBeNull();
+});
+
+test('greater-than-equals without date operator', () => {
+  const expression: JSONObject = {
+    '>=': [{date: '2000-10-10'}, '2000-01-01'],
+  };
+
+  expect(() => evaluate(expression, {})).toThrow();
+});
+
 test('less-than comparison with ISO-8601 date strings', () => {
   const expression: JSONObject = {
     '<': [{date: '2026-03-01'}, {date: '2026-03-01'}],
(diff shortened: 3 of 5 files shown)


vaszig:
What I understood from taking a look at the backend code for `today` operator is that it's not (correctly) evaluated.

https://github.com/open-formulieren/open-forms/blob/main/src/openforms/utils/json_logic/partial_evaluation.py#L126
https://github.com/open-formulieren/open-forms/blob/main/src/openforms/utils/tests/test_json_logic.py#L384

src/logic/json-logic/extensions/comparators.ts
```diff
@@ -14,13 +22,34 @@ import type {JsonLogicEngineMethod} from './types';
...
+      if (typeof arg === 'string') {
+        if (arg === '') return null;
+
+        if (hasDate) {
+          const timestamp = parseDateString(arg);
+          if (timestamp !== null) {
+            return timestamp;
+          }
+        }
+      }
```
viktorvanwijk: Did you add this because there was still a crash in the frontend, even with the early exit of `null` values?

I think this is exactly the problem I was pointing out with the `today` operator, which should be fixed in the backend during partial evaluation.

Our `json-logic-py` library actually raises an error for this situation, and I would expect the SDK to do the same:
```python
>>> jsonLogic({">": [{"date": "2000-10-10"}, "2000-01-01"]})
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "/home/viktor/open-forms/env/lib/python3.12/site-packages/json_logic/__init__.py", line 415, in jsonLogic
    return operations[operator](*values)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/viktor/open-forms/env/lib/python3.12/site-packages/json_logic/__init__.py", line 281, in <lambda>
    ">": lambda a, b: less(b, a),
                      ^^^^^^^^^^
  File "/home/viktor/open-forms/env/lib/python3.12/site-packages/json_logic/__init__.py", line 70, in less
    return a < b and (not args or less(b, *args))
           ^^^^^
TypeError: '<' not supported between instances of 'str' and 'datetime.date'
```

vaszig: I guess you mean the extra `hasDate` right?This was added because the logic rule (foe example disable next step if the date is yesterday) was not properly executed. There is no crash btw even without this.

viktorvanwijk: > I guess you mean the extra hasDate right?

Yes, and the conversion to date objects of strings specifically. I'm not sure this should be necessary.

> There is no crash btw even without this.

Hmm, I did experience one. Might be a difference in test setup or commented out code :thinking:

vaszig:
I have revisited this PR and added only the necessary after this change in the backend too. Viktor, let me know if you agree and if I understood what you meant concerning the evaluation of the `today` operator.

https://github.com/open-formulieren/open-forms/pull/6480

| Our json-logic-py library actually raises an error for this situation, and I would expect the SDK to do the same

Now the same is happening for the SDK, it raises an error.

src/logic/json-logic/extensions/date.ts
```diff
@@ -25,6 +25,10 @@ export const jsonLogicDate: JsonLogicEngineMethod = ([dateString]) => {
   // anything that isn't a string.
   if (typeof dateString !== 'string') return dateString;
 
+  if (dateString === '' || dateString === null) {
```
viktorvanwijk: This is indeed what I had in mind!

Minor thing perhaps, but this check on `null` will already be caught by the line above, no?

src/logic/json-logic/extensions/comparators.ts
```diff
@@ -22,6 +22,11 @@ const comparatorOverrideFactory = (original: JsonLogicEngineMethod): JsonLogicEn
       }
       return originalArg;
     });
+
+    if (args.some(arg => arg === null)) {
```
viktorvanwijk: Hmm, with the 'undefined' PR being merged, we might have to look for undefined values here as well, like we do in the [backend](https://github.com/maykinmedia/json-logic-py/blob/2bd86798d74820c035fe48ff56fc70b9551aa411/json_logic/__init__.py#L321) 🤔 

Probably a separate PR is better for this

viktorvanwijk (commented):
I think the `datetime` operator suffers from the same issue. Probably good to fix that as well.

Could you add the "why" of this change to the commit message 😇

src/logic/json-logic/extensions/comparators.ts
```diff
@@ -22,6 +22,11 @@ const comparatorOverrideFactory = (original: JsonLogicEngineMethod): JsonLogicEn
       }
       return originalArg;
     });
+
+    if (args.some(arg => arg === null || arg === undefined)) {
```
vaszig, line 26: Not sure if this is enough (because you mentioned creating another PR-so maybe I am missing something), but I think you mean when we receive an undefined value from the `var` operator?

viktorvanwijk: Yeah something like this should be it. I mentioned a separate PR because it will affect all operators defined [here](https://github.com/maykinmedia/json-logic-py/blob/2bd86798d74820c035fe48ff56fc70b9551aa411/json_logic/__init__.py#L321), not only the comparator operators.

vaszig: Agree yes, I will create a separate ticket for this. So, we can focus now on this PR and your upcoming review right?

viktorvanwijk (approved):
If you can add some explanation to the first commit message as well, I think it's OK like this :)

Especially for these kind of bug fixes, it's important to mention why these early exits are implemented (to match the backend behaviour).

vaszig:
Ticket https://github.com/orgs/open-formulieren/projects/2/views/27?pane=issue&itemId=217850992&issue=open-formulieren%7Copen-forms%7C6487 was created for the remaining tasks (`undefined` handling for all the operators).

vaszig:
Backports

stable/3.5.x: 2d6b35d4b9821729da56cc6d7fb41afe6cbb0ae5

PR #1011, opened 2026-07-27
[OF#6487] Added explicit undefined check to all operators

### LINKED ISSUE #6487: Make sure `undefined` is handled in the SDK json logic
After updating the backend `maykin-json-logic-py` fork we did the same for the SDK json logic. We also want to make sure that the `undefined` value is handles in all the operators like we do in the backend:

https://github.com/maykinmedia/json-logic-py/blob/2bd86798d74820c035fe48ff56fc70b9551aa411/json_logic/__init__.py#L321

vaszig:
Closes open-formulieren/open-forms#6487

### DIFF
--- a/src/logic/json-logic/overrides.spec.ts
+++ b/src/logic/json-logic/overrides.spec.ts
@@ -1,5 +1,5 @@
 import type {JSONObject} from '@open-formulieren/types';
-import {expect, test} from 'vitest';
+import {describe, expect, test} from 'vitest';
 
 import evaluate from './index';
 
@@ -18,16 +18,20 @@ test.each([null, undefined])(
   }
 );
 
-test.each([
-  [{'+': [null, 5]}],
-  [{'-': [null, 5]}],
-  [{'*': [null, 5]}],
-  [{'/': [null, 5]}],
-  [{'%': [null, 5]}],
-  [{min: [null, 5]}],
-  [{max: [null, 5]}],
-])('math with null-ish values short-circuits, expression: %o', (expression: JSONObject) => {
-  const result = evaluate(expression, {});
+describe.each(['>', '>=', '<', '<=', '+', '-', '*', '/', '%', 'min', 'max'])(
+  '%s operator',
+  operator => {
+    test.each([
+      ['undefined', undefined],
+      ['null', null],
+    ])('returns null with %s', (_, value: null | undefined) => {
+      // @ts-expect-error undefined is not part of the JSONObject type, but we're testing
+      // the behaviour here
+      const expression: JSONObject = {
+        [operator]: [value, 8],
+      };
 
-  expect(result).toBe(null);
-});
+      expect(evaluate(expression, {})).toBeNull();
+    });
+  }
+);


src/logic/json-logic/extensions/math.ts
```diff
@@ -19,7 +19,7 @@ export const customAddition: JsonLogicEngineMethod = args => {
   const original = defaultMethods['+'];
 
   // if any null-ish value is in the args, shortcut to match backend behaviour
-  if (args.some(arg => arg == null)) return null;
+  if (args.some(arg => arg === null || arg === undefined)) return null;
```
sergei-maertens: how necessary is this? because `arg == null` already covers `arg` being `undefined`:

```ts
null === null // true
undefined === undefined // true
null == undefined // true
null === undefined // false
```

src/logic/json-logic/math.spec.ts
```diff
@@ -0,0 +1,19 @@
+import type {JSONObject} from '@open-formulieren/types';
+import {describe, expect, test} from 'vitest';
+
+import evaluate from './index';
+
+describe.each(['+', '-', '*', '/', '%', 'min', 'max'])('%s operator', operator => {
```
sergei-maertens: so, were these tests failing before your patch?

sergei-maertens (changes requested):

src/logic/json-logic/math.spec.ts
vaszig: No, `arg == null` was already covering the cases, I added them because of the explicit checks above.

src/logic/json-logic/extensions/math.ts
vaszig: It's not really necessary since what you have mentioned is indeed the case. The only reason that I did that is because of this comment (or maybe I misunderstood something), that's why I named that "explicit check".

https://github.com/open-formulieren/open-forms-sdk/pull/1005#discussion_r3637346919

sergei-maertens: right - perhaps @viktorvanwijk wasn't aware of the `==` behaviour for `undefined` and `null`?

I think as long as we have tests that cover those backend operators with `undefined` as an argument, that's the most important part. The implementation itself doesn't need to be updated as long as those tests pass, so it seems like the scope of this ticket is purely adding more tests.

src/logic/json-logic/math.spec.ts
sergei-maertens: okay - good to have the explicit tests indeed, tough I think they belong in `overrides.spec.ts` which already exists (there's some short-circuit tests there for `null` already).

Additionally, I want some explicit tests cases for the other uncovered operators: `>`, `>=`, `<`, `<=`, ~~`years` (is this even used anywhere?)~~

src/logic/json-logic/extensions/math.ts
viktorvanwijk: > right - perhaps @viktorvanwijk wasn't aware of the == behaviour for undefined and null?

I did not realize there where only two equality signs, my bad

vaszig: Np, I thought that we wanted to make this really explicit for the readers too!reverted the changes, left only the additional test cases which I am going to put in the correct file.

sergei-maertens (approved):

PR #1012, opened 2026-07-30
Add callout page

### LINKED ISSUE #6317: Create help callout page
Image

# [Refinement](https://github.com/open-formulieren/open-forms/issues/5787#issuecomment-3870621080):

**Callout**

* Click the help button on the callout page -> what happens? a11y? -> aria-disabled
* Callout is only shown when help text on form is configured + display option selected (before/after start page) + text in the global config (disable config option on form if the preconditions are not met)

Aspects:

* global configuration -> (tinymce images: not setup right now, so we go for a simple `models.ImageField` instead): 1D
* only display configuration options for callout display if the global config is done for its content: 2 days
* displaying in the SDK -> (add new page/route(s) conditional before/after start page, preserve query parameters, new component + CSS + design tokens & render image): 5-10days
* help button -> (interactive vs non interactive, navbar separation): 3 days
* add configuration to form detail endpoint -> (global configuration + form-specific configuration): 1 day

# [Acceptatiecriteria](https://github.com/open-formulieren/open-forms/issues/5787#issuecomment-4178223829):

**Story:** Als gebruiker wil ik eenvoudig contact opnemen met de gemeente Den Haag om mij te helpen bij mijn aanvraag, tijdens het invullen van het formulier. 

### Onderdelen in scope
- Hulpknop: het telefoon icoon dat aangeklikt kan worden
- Callout: de introductie van de hulpknop
- De dialog: het schermpje dat getoond wordt door op de hulpknop te klikken

**Acceptatiecriteria**
- Deze feature (hulpknop, callout en dialog) kan optioneel ‘aan’ of ‘uit’ gezet worden per formulier
- De hulpknop en callout staan vast qua tekst, afbeelding en design in alle formulieren
- De dialog kan per formulier qua tekst aangepast worden (wysiwyg editor)
afbeelding tonen is nice-to-have
- De callout kan getoond worden:
optie 1: na inloggen (met DigiD)
optie 2: voor inloggen (met DigiD)
- De hulpknop staat bovenaan de pagina en is niet sticky

# Tasks

- [x] Add fields (content + image) to display on this page to the `GlobalConfiguration` (the image is optional but like the other fields is fixed in position)
- [x] Add a new form configuration option `help_callout_display` with the options `before_startpage`, `after_startpage`, `none`
  - Show a disabled field for this option whenever the content field in the `GlobalConfiguration` is **not** filled in
  - Check what happens to the single-step form configuration when adding this option -> should be disabled
- [x] Expose the new `GlobalConfiguration` fields and form specific fields through the form detail endpoint for the front end
- [x] Include a new route in the SDK to show the page's content (see point above for the when and how)
  -   Based on the `help_callout_display` configuration option, an additional route is added in the SDK, matching the screenshot. If it's before the startpage, it follows after the (optional) introduction page. if after the startpage, it comes directly after the start page 
- [x] Implement additional design token wishes
- [x] Add some documentation

viktorvanwijk:
Partly closes open-formulieren/open-forms#6317


public/images/helpdesk.jpg
sergei-maertens, line 1: Do we need any copyright/attribution for this (stock?) image?

src/data/forms.ts
```diff
@@ -115,4 +115,9 @@ export interface Form {
   submissionStatementsConfiguration: SubmissionStatementConfiguration[];
   submissionReportDownloadLinkTitle: string;
   communicationPreferencesPortalUrl: string;
+  helpCalloutPage: {
+    display: 'before_start_page' | 'after_start_page' | 'never';
+    content: string;
+    image: string | null;
```
sergei-maertens, line 121: Hrm, I need to check what the API actually returns when there's no image, but it wouldn't surprise me that it's actually `null` 🤔 

My backend documentation may be wrong for the help dialog image.

src/components/HelpButton.tsx
sergei-maertens, line 1: This is what's to be replaced with my button implementation, right?

src/components/HelpCalloutPage/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+                  defaultMessage: 'Help callout page image',
+                })}
+                className="openforms-help-callout-page-dialog__image"
+              />
+            )}
+
+            <Body
+              modifiers={['wysiwyg']}
+              component="div"
+              dangerouslySetInnerHTML={{__html: helpCalloutPage.content}}
```
sergei-maertens: you'll want to chunk that through `DOMPurify` before rendering it, and we should also use the utrecht `HTMLContent` component for the styling :) see my PR for the dialog content for an example

src/components/HelpCalloutPage/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+        appearance="primary-action-button"
+        className="openforms-start-link"
+        onClick={() => setHelpCalloutPageVisited(true)}
+      >
+        <FormattedMessage
+          description="Button text for link to continue from introduction page to start page"
+          defaultMessage="Continue"
+        />
+        <FAIcon icon="arrow-right-long" />
+      </Link>
```
sergei-maertens: Should be put that in its own component so that the messages en class names are locked in the component rather than duplicating them?

Similar to how we have the `PreviousLink` component?

src/components/HelpCalloutPage/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+const HelpCalloutPage: React.FC = () => {
+  const {steps, helpCalloutPage} = useFormContext();
+  const {submission, setHelpCalloutPageVisited} = useSubmissionContext();
+  const {preserveQueryParams} = useQueryParams();
+  const intl = useIntl();
+  const buttonRef = useRef<HTMLButtonElement | null>(null);
+
+  const [triangleOffset, setTriangleOffset] = useState<number>(0);
+
+  useLayoutEffect(() => {
```
sergei-maertens: this probably took quite some time... BUT - can't we make our own lives easier by making this a design token, and leave it to the theme builders to set up the correct offset?

public/images/helpdesk.jpg
viktorvanwijk: Hmm, not sure actually

src/data/forms.ts
viktorvanwijk: Yeah it's `null`, I checked it
image

src/components/HelpButton.tsx
viktorvanwijk: Yes

src/components/HelpCalloutPage/index.tsx
viktorvanwijk: Isn't the sanitation in the backend sufficient here?

src/components/HelpCalloutPage/index.tsx
viktorvanwijk: Discussed in office: the help button is fixed width, so yes that should be possible.

Was a fun little exercise anyway though :)

src/components/HelpCalloutPage/index.tsx
sergei-maertens: Trust no one! :wink: If a bug gets introduced in the backend that can lead to XSS - defense in depth with multiple layers protects against that.

(you could also make an argument that we don't even need the backend sanitizing because we use CSP anyway)

public/images/helpdesk.jpg
viktorvanwijk: What about this AI-generated one?

file_00000000e3c881f482fc678ddbb88648

src/data/forms.ts
sergei-maertens: I updated my generated docs in my backend PR accordingly :)

src/components/FormStart/tests.spec.tsx
```diff
@@ -99,6 +101,7 @@ test('Start form without having logged in', async () => {
   await expect.poll(() => startSubmissionRequest).not.toBeUndefined();
   const requestBody = await startSubmissionRequest!.json();
   expect(requestBody.anonymous).toBe(true);
+  await vi.waitFor(() => expect(screen.getByText('Step page')).toBeVisible());
```
sergei-maertens:
```suggestion
  await expect.element(screen.getByText('Step page')).toBeVisible();
```
this should be equivalent - it encapsulates the underlying `expect.poll(...)` implementation

src/components/FormStart/tests.spec.tsx
```diff
@@ -380,3 +383,60 @@ test('Form start page shows invisible login buttons when auth_visible=all', asyn
...
+
+test('With help callout page content shown after the start page', async () => {
+  mswWorker.use(mockSubmissionPost());
+  const form = buildForm({
+    helpCalloutPage: {display: 'after_start_page', content: '<p>Some content</p>', image: null},
+  });
+
+  const screen = await render(<Wrap form={form} />);
+  await userEvent.click(screen.getByRole('button', {name: 'Begin'}));
+  await vi.waitFor(() => expect(screen.getByText('Help callout page')).toBeVisible());
```
sergei-maertens: See above

src/components/FormStart/tests.spec.tsx
```diff
@@ -380,3 +383,60 @@ test('Form start page shows invisible login buttons when auth_visible=all', asyn
...
 });
+
+test('With help callout page content shown after the start page', async () => {
+  mswWorker.use(mockSubmissionPost());
+  const form = buildForm({
+    helpCalloutPage: {display: 'after_start_page', content: '<p>Some content</p>', image: null},
+  });
+
+  const screen = await render(<Wrap form={form} />);
+  await userEvent.click(screen.getByRole('button', {name: 'Begin'}));
```
sergei-maertens:
```suggestion
  await screen.getByRole('button', {name: 'Begin'}).click();
```
vitest browser mode exposes some nice top-level API :) you should only seldomly need the `userEvent` import

src/components/HelpCalloutPage/HelpCalloutPage.mdx
```diff
@@ -0,0 +1,44 @@
...
+A complete example is shown below:
+
+```json
+{
+  "of": {
+    "help-callout-page": {
+      "dialog": {
+        "triangle": {
+          "size": {"value": "35px"},
+          "margin-right": {"value": "35px"}
```
sergei-maertens: should this be `margin-inline-end` rather than `margin-right`? I don't know what the orientation of this thing would become with right-to-left text, but I suspect the flexbox content with the icon would swap it around in its current form 🤔

src/components/HelpCalloutPage/HelpCalloutPage.mdx
```diff
@@ -0,0 +1,44 @@
...
+        },
+        "margin-top": {"value": "1em"},
+        "margin-bottom": {"value": "2em"},
```
sergei-maertens:
```suggestion
        "margin-block-start": {"value": "1em"},
        "margin-block-end": {"value": "2em"},
```

src/components/HelpCalloutPage/HelpCalloutPage.mdx
```diff
@@ -0,0 +1,44 @@
...
+        "padding": {"value": "3em"},
+        "column-gap": {"value": "3em"}
+      },
+      "mobile": {
+        "dialog": {
+          "padding": {"value": "2em"},
+          "row-gap": {"value": "2em"}
+        }
+      },
```
sergei-maertens: We don't do viewport-specific design tokens anymore, instead it's up to theme designers to include breakpoint-specific CSS overrides (see https://github.com/open-formulieren/open-forms/issues/5547)

src/components/HelpCalloutPage/HelpCalloutPage.stories.ts
```diff
@@ -0,0 +1,64 @@
...
+  title: 'Views / HelpCalloutPage',
+  component: HelpCalloutPage,
+  decorators: [withForm, withNuqs, withRouter],
+  parameters: {
+    formContext: {
+      form: buildForm({
+        helpCalloutPage: {
+          display: 'before_start_page',
+          content: DEFAULT_CONTENT,
+          image: '/images/helpdesk.png',
```
sergei-maertens: I think this will break when deployed to github pages because there storybook is running on a subpath.


```suggestion
          image: './images/helpdesk.png',
```
(I hope this resolves against the public folder)

src/components/HelpCalloutPage/HelpCalloutPage.stories.ts
```diff
@@ -0,0 +1,64 @@
...
+      form: buildForm({
+        helpCalloutPage: {
+          display: 'before_start_page',
+          content: DEFAULT_CONTENT,
+          image: null,
+        },
+      }),
+    },
+  },
+} satisfies Meta<typeof HelpCalloutPage>;
```
sergei-maertens: rebase gone wrong?

src/components/HelpCalloutPage/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+const HelpCalloutPage: React.FC = () => {
+  const {steps, helpCalloutPage} = useFormContext();
+  const {submission, setHelpCalloutPageVisited} = useSubmissionContext();
+  const {preserveQueryParams} = useQueryParams();
+  const intl = useIntl();
+
+  if (!helpCalloutPage.content || helpCalloutPage.display === 'never')
+    // In case a user tries to navigate to this page manually when it shouldn't be visible, just
+    // redirect to the start page directly.
```
sergei-maertens, line 24: nice - another edge case is that a form was configured with the callout page, and then some removed the content of that in the global configuration (probably worth adding as a story?)

src/components/HelpCalloutPage/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+    // No need to preserve query parameters, because the submission was already created.
+    nextPageUrl = `/stap/${steps[0].slug}`;
+  }
+
+  return (
+    <FormContainer>
+      <div>
+        <FormStepTopNav>
+          <FormHelpButton disabled />
+        </FormStepTopNav>
```
sergei-maertens, line 52: The designs in Figma show a previous link as well - I think that only makes sense in the display `after_start_page` though

src/components/HelpCalloutPage/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+        <div className="openforms-help-callout-page-dialog">
+          <div className="openforms-help-callout-page-dialog__triangle"></div>
+          <div className="openforms-help-callout-page-dialog__container">
+            {helpCalloutPage.image && (
+              <Image
+                src={helpCalloutPage.image}
+                alt={intl.formatMessage({
+                  description: 'Help callout page image alt text',
+                  defaultMessage: 'Help callout page image',
+                })}
```
sergei-maertens, line 63: for the dialog I opted for `aria-hidden="true"` as the image is purely illustrative and not critical for non-sighted users IMO

src/components/HelpCalloutPage/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+                className="openforms-help-callout-page-dialog__image"
+              />
+            )}
+
+            <FormHelpContent content={helpCalloutPage.content} />
+          </div>
+        </div>
+      </div>
+
+      <NextLink url={nextPageUrl} onClick={() => setHelpCalloutPageVisited(true)} />
```
sergei-maertens: should this be an `onClick`, or maybe an `useEffect` when the page/component renders? 🤔 

If they navigate in the menu rather than clicking this button/link, the state wouldn't be updated, would it?

src/scss/components/_help-callout-page.scss
```diff
@@ -0,0 +1,47 @@
+@use 'microscope-sass/lib/bem';
+
+$triangle-size: var(--of-help-callout-page-dialog-triangle-size, 35px);
+$dialog-background-color: var(--of-help-callout-page-dialog-color, var(--of-color-secondary));
```
sergei-maertens: The `of-dialog-secondary` design token(s) are deprecated - can we replace it with the blue color that's used as fallback background color for the button? I think it's supposed to be the same color.

src/scss/components/_help-callout-page.scss
```diff
@@ -0,0 +1,47 @@
...
+
+.openforms-help-callout-page-dialog {
+  margin-top: var(--of-help-callout-page-dialog-margin-top, 1em);
+  margin-bottom: var(--of-help-callout-page-dialog-margin-botton, 2em);
```
sergei-maertens: I think it's better to use the logical units rather than top/bottom/left/right

src/scss/components/_help-callout-page.scss
```diff
@@ -0,0 +1,47 @@
...
+
+  @include bem.element('triangle') {
+    width: 0;
+    height: 0;
+    border-left: $triangle-size solid transparent;
+    border-right: 0;
+    border-top: 0;
```
sergei-maertens: here too some logical properties might make more sense to be ready for right-to-left page layouts?

Though this is beyond my experience - I really don't know if the icon/link should move in the page or not 🤷

src/scss/components/_help-callout-page.scss
```diff
@@ -0,0 +1,47 @@
...
+    border-bottom: $triangle-size solid $dialog-background-color;
+    position: relative;
+    margin-left: auto;
+    margin-right: var(--of-help-callout-page-dialog-triangle-margin-right, 35px);
```
sergei-maertens: not sure if it should be margin or a `transform(translateX(...))` rule - either way I think it's good to make the design token name agnostic and use something like `--of-help-callout-page-dialog-triangle-inline-offset` instead?

src/scss/components/_help-callout-page.scss
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  @include bem.element('container') {
+    background: $dialog-background-color;
+    padding: var(--of-help-callout-page-dialog-padding, 3em);
```
sergei-maertens: I think it's usually better to break out the padding into the individual components and have a design token for each, so `padding-block-end`, `padding-block-start`, `padding-inline-end` and `padding-inline-start` (NL Design System uses alphabetic sorting as well)

src/scss/components/_help-callout-page.scss
```diff
@@ -0,0 +1,47 @@
...
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+    column-gap: var(--of-help-callout-page-dialog-column-gap, 3em);
+    justify-content: flex-start;
+    align-content: space-around;
```
sergei-maertens:
```suggestion
    align-content: space-between;
```
if we already have the padding, I think on left/right it can stick to the edge, no?

src/scss/components/_help-callout-page.scss
```diff
@@ -0,0 +1,47 @@
...
+
+    @include mobile-only {
+      padding: var(--of-help-callout-page-mobile-dialog-padding, 2em);
+      flex-direction: column;
+      row-gap: var(--of-help-callout-page-mobile-dialog-row-gap, 2em);
+    }
```
sergei-maertens: this belongs in `src/scss/_responsive.scss` instead, and should override the respective design tokens with our theme-specific values. The MDX can document this suggested CSS override like I did with the help dialog :)

src/scss/components/_help-callout-page.scss
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  @include bem.element('image') {
+    --of-image-max-inline-size: var(--of-help-callout-page-image-max-inline-size);
+
+    @include mobile-only {
+      max-width: 100%;
+      flex-shrink: 1;
+    }
```
sergei-maertens: Same remark here about documentation + our responsive overrides

sergei-maertens (changes requested):

src/components/FormStart/tests.spec.tsx
viktorvanwijk: I had that at first, but it's really flaky without it - it doesn't seem to properly navigate each time

src/components/HelpCalloutPage/HelpCalloutPage.stories.ts
viktorvanwijk: Copy from the default :grin:

src/components/HelpCalloutPage/index.tsx
viktorvanwijk: Yeah I was doubting about that. Didn't add anything because I also didn't see the previous link on the start page (when there is an introduction page). We could add it after the start page indeed

src/components/HelpCalloutPage/HelpCalloutPage.stories.ts
viktorvanwijk: Where do you expect it to break exactly? It seems to work in chromatic and the corresponding storybook build

src/components/HelpCalloutPage/index.tsx
viktorvanwijk: I think I have that as a unit test in `HelpCalloutPage.spec.tsx`: 'No content configured', or is that not the situation you meant?

src/components/HelpCalloutPage/index.tsx
viktorvanwijk: Yeah I did think about that, but went with this one for the "step completed" behaviour. Also, I was under the impression that navigating like this is only possible for admins, but I see now that is only true for actual form steps. Will change it :)

viktorvanwijk: Ugh, my solution didn't work anyway, because the check mark would be removed again after starting the submission. Also the using `useEffect` seems to have the same behavior :/

src/components/HelpCalloutPage/HelpCalloutPage.stories.ts
viktorvanwijk: Discussed in office: [this storybook](https://open-formulieren.github.io/open-forms-sdk/?path=/docs/introduction--docs)

src/components/HelpCalloutPage/index.tsx
viktorvanwijk: Discussed in office: it's tricky to get this working properly, and given that adding it to the navigation bar was not in the original specification of the ticket, will leave it out of scope for now

Reference to the original commits: adc23458cbe31266738cb5e358012f52503efa15 and dfce63ea1089cbb4a777be32e7b2f0b00a4970ca

src/components/FormStart/tests.spec.tsx
viktorvanwijk: Ah I see now that you suggested to use `.element`, that does seem to work indeed :)

src/components/HelpCalloutPage/index.tsx
sergei-maertens: yeah that's sufficient!

sergei-maertens (approved):
Let's go :)

PR #1013, opened 2026-07-30
Add form step help dialog + trigger button

sergei-maertens:
Closes open-formulieren/open-forms#6318

**TODO**

- [x] Define our own design token values & set neutral defaults
- [x] Document available design tokens

## Screenshots

image

image

PR #1014, opened 2026-07-31
Contact and help shared functionality

sergei-maertens:
Extracted the shared components for #1012 and #1013 into a separate PR

### DIFF
--- a/src/scss/components/_form-help-button.scss
+++ b/src/scss/components/_form-help-button.scss
@@ -1,0 +1,53 @@
+@use 'microscope-sass/lib/bem';
+
+// @import instead of @use because breakpoints are defined globally
+@import 'microscope-sass/lib/responsive';
+
+.openforms-form-help-button {
+  --utrecht-button-background-color: var(--of-form-help-button-background-color, #d3e3ec);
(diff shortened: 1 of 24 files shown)


src/scss/components/_form-help-button.scss
```diff
@@ -0,0 +1,53 @@
...
+  --utrecht-button-padding-inline-start: var(--of-form-help-button-padding-inline-start, 8px);
+
+  margin-inline-start: auto;
+
+  // Render the disabled state closely to the normal state, but allow themes to
+  // customize this if necessary
+  &.utrecht-button.utrecht-button--disabled {
+    --utrecht-button-disabled-background-color: var(
+      --of-form-help-button-disabled-background-color,
+      var(--of-form-help-button-disabled-background-color)
```
viktorvanwijk: Is the double definition necessary in case that token doesn't exist, or is it a typo? (Or something else?)

viktorvanwijk (approved):

src/scss/components/_form-help-button.scss
sergei-maertens: Oops, the second should drop the disabled modifier! Falling back to the normal styles

